PR #30, opened 2020-06-18
Update README.md to show recursive clone command

Blquinn:
Found that the gir files are a module, so you need --recursive to fetch them.


badcel:
Thank you!

Blquinn:
No problem. This project looks awesome by the way, can't believe how well it works for a few months of work!

PR #37, opened 2020-07-26
Initial Windows Support

firox263:
Adds support for running gir.core on Windows.

**Preview**
image

**Proposed Changes**
- Adds a `ConvertLibName()` method to the Generator, which maps shared library names to DLL names when on Windows.
- Adds `WINDOWS.md` with instructions for building.
- Adds a new sample called **GtkMinimal** which only uses the default GTK stack (e.g. no webkit).

For mapping DLL names, I've opted for the MSYS2 convention as that's the recommended way to get the GTK binaries (e.g. `libglib-2.so.0` becomes `libglib-2-0.dll`). I know [vcpkg](https://github.com/Microsoft/vcpkg) uses a different naming convention (`glib-2.dll` I believe?), so that's something to consider down the road. I'm also not sure how robust this method is, but it is working for the entire GTK stack.

With WINDOWS.md, perhaps this should be on the wiki rather than in the main repo?

Finally, I created a minimal example which only uses the core gtk stack as webkit2gtk apparently doesn't build on windows (see https://github.com/msys2/MINGW-packages/issues/3484).

I haven't had the opportunity to test the example on linux yet.

Any and all feedback is much appreciated.


firox263:
Not sure what happened there.

Changed to use implicit typing to be consistent with the rest of the codebase.

badcel:
Thank you very much.

PR #45, opened 2020-08-01
Fix namespaces

### LINKED ISSUE #27: Namespaces and class names
The current class names of the high level classes are mostly prefixed with a "G" to avoid conflicts with the generated class for the low level api.

The namespace is "[Library].Core". The low level classes live in the namespace "[Library]". 

Should the high level namespace be part of the low level namespace? I would like to separate the layers completely, allowing simpler class names and clean namespaces.

Should there be one global namespace for the high level api or should every library be on its on namespace? How would theses namespaces be named?

Feel free to post your opinion.

badcel:
Fix #27

### DIFF
--- a/Libs/Gtk4/Core/Classes/Application.cs
+++ b/Libs/Gtk4/Core/Classes/Application.cs
@@ -1,9 +1,7 @@
-using Gio.Sys;
-
 namespace Gtk
 {
     public class Application : Gio.Application
     {
-        public Application(string applicationId) : base(Sys.Application.@new(applicationId, ApplicationFlags.flags_none)) {}
+        public Application(string applicationId) : base(Sys.Application.@new(applicationId, Gio.Sys.ApplicationFlags.flags_none)) {}
     }
 }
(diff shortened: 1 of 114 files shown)


Libs/Gtk4/Core/Classes/Application.cs
```diff
@@ -1,9 +1,9 @@
-using Gio;
+using Gio.Sys;
```
firox263, line 1: I'm not too comfortable with importing another library's `Sys` namespace like this. Ideally only the corresponding higher-level wrapper should ever access it (i.e. We'd wrap `ApplicationFlags` in an enum to avoid re-exporting `Gio.Sys`).

Maybe for now we could do `@new(..., Gio.Sys.ApplicationFlags.flags_none)` and then later refactor this out?

badcel: You are right. I think there should be a clear distinction between the implementation and the Sys namespace. I removed all usings which where Like "using *.Sys".

PR #48, opened 2020-08-09
Type System Integration for GObject

### LINKED ISSUE #46: [RFC-1] Type System Integration with GObject
# Integrating C# and GObject
## RFC
**Please see the RFC here: https://github.com/gircore/rfcs/blob/master/rfc-001-type-integration.md**
It explains the proposal and some potential solutions.

## Summary
Type Integration means that when the user creates a class that derives from GObject.Object, the library will automatically create a corresponding GType behind the scenes. Effectively, it allows us to write "native" GObject code using C#.

This enables us to implement the following features, among many others:
 * **GObject Properties** - Subclasses can expose properties to the gobject type system, allowing for them to be viewable/changeable in GtkInspector, etc.
 * **Overriding Virtual Functions** - Subclasses can override virtual functions (See GtkWidgetClass). For example, creating a custom container widget in Gtk with full control over sizing and layout.
 * **Composite Template Support** - Supports binding a subclass to a composite template defined in an XML file. Particularly useful for complex custom widgets.
 * **Bidirectional Bindings** - The possibility of using C# defined types from C and/or other language bindings. With .NET 5 we can use source generators to output native-callable functions.

Like with the namespace refactor (#27, PR #45), this will be a huge change that affects almost the entire codebase. For this reason, it is probably best to focus on completing this before we write more bindings.

firox263:
Transferred my working branch to the main repo. Currently only `Libs/GObject` compiles, although @badcel PR fixes many of the other bindings (will need to be rebased after my last commit).

Let's use this draft PR to track progress on the implementation and leave #46 for discussions about the high-level interaction.

Resolves #46

### DIFF
--- a/Libs/GObject/Core/Classes/Closure.cs
+++ b/Libs/GObject/Core/Classes/Closure.cs
@@ -8,6 +8,8 @@ namespace GObject
     public partial class Closure
     {
         private IntPtr handle;
+        public IntPtr Handle => handle;
+        
         private readonly Action? callback;
         private readonly ActionRefValues? callbackRefValues;
 
@@ -33,7 +35,5 @@ namespace GObject
 
             callbackRefValues?.Invoke(ref param_values);
         }
-
-        public static implicit operator IntPtr (Closure closure) => closure.handle;
     }
 }
--- a/Libs/GObject/Core/Classes/Closure.cs
+++ b/Libs/GObject/Core/Classes/Closure.cs
@@ -23,7 +23,7 @@ namespace GObject
 
         private Closure(Object obj)
         {
-            handle = Sys.Closure.new_object((uint)Marshal.SizeOf(typeof(Sys.Closure)), obj);
+            handle = Sys.Closure.new_object((uint)Marshal.SizeOf(typeof(Sys.Closure)), obj.Handle);
             Sys.Closure.set_marshal(handle, MarshalCallback);
         }
 
--- a/Libs/GObject/Core/Classes/Object.ClassImpl.cs
+++ b/Libs/GObject/Core/Classes/Object.ClassImpl.cs
@@ -1,0 +1,94 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace GObject
+{
+    public partial class Object
+    {
+        // Given a type, it will walk up the object's inheritance tree
+        // to find the first non-user-defined class (i.e. implemented by
+        // a C library).
+        private static Type GetBoundaryType(Type type)
+        {
+            while (IsSubclass(type))
+                type = type.BaseType;
+            
+            return type;
+        }
+
+        private static Sys.TypeQuery QueryType(ulong gtype)
+        {
+            // Create query struct
+            Sys.TypeQuery query = default;
+
+            // Convert to Pointer
+            IntPtr ptr = Marshal.AllocHGlobal(Marshal.SizeOf(query));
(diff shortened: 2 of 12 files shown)


badcel:
I'm going to create the code to get the _get-type_ functions working. Should the code go into this branch or in develop and be merged into this?

firox263:
> I'm going to create the code to get the _get-type_ functions working. Should the code go into this branch or in develop and be merged into this?

Probably straight into this branch, since it's not very useful for `develop` anyway

badcel:
I added support for the get-type function. I hope it works as expected.

firox263:
In the process of selectively porting firox263#1 to this PR.

Currently the last obstacle before we can get the samples sort-of running again is how we handle constructors. I don't want to make `Handle` settable or public as that opens the door to all sorts of problems, but we also can't use `g_object_new_with_properties()` either as that doesn't set things like application for `ApplicationWindow`.

Maybe we could pass a factory delegate down the inheritance chain in addition to IntPtr. E.g.
```cs
public Label(string text) : this(() => Sys.Label.@new(text)) {}
internal Label(Object.Factory? ctor) : base(ctor)
```

So then we'd have:
```cs
public Object(ConstructorDelegate? ctor = null)
{
    var type = GetType();

    // Ensure type exists. We must call this before using
    // any typedict functions.
    EnsureType(type);

    // Create new GObject
    if (ctor != null)
        handle = ctor.Invoke();
    else
    {
        // Fallback to g_object_new
        var zero = new IntPtr(0);
        handle = Sys.Object.new_with_properties(
            TypeDictionary.FromType(GetType()),
            0, ref zero, new Sys.Value[0]
        );
    }
    
    Initialize(out closures);
}
```

Our wrapper classes would probably have an extra constructor chain (technically they don't necessarily need to implement this), but none of this is passed on to the end user.

```cs
[Wrapper]
public class InitiallyUnowned : Object
{
    protected InitiallyUnowned() {}
    protected InitiallyUnowned(IntPtr handle) : base(handle) {}
    protected InitiallyUnowned(Factory? ctor) : base(ctor) {}

    ...
}
```

Downside is more verbosity, upside is no lazy init. I'll experiment with it a bit and report back.

badcel:
> I don't want to make Handle settable or public 

Yes, I don't want it to be setable either. I would prefer a solution where it is not public. But if there is only one good solution with a public (non setable) `Handle` property it would be okay for me.

> we also can't use `g_object_new_with_properties()` either as that doesn't set things like application for ApplicationWindow.

We could probably pass the appliction to `g_object_new_with_properties()` ourself. If this helps in the process of subclass creation we could evaluate more in this direction.

> Maybe we could pass a factory delegate down the inheritance chain 

This would be a solution, but the subclass would need to provide the delegate right? I would like to try to avoid a delegate for subclasses, as it is quite unusual.

> I might try and rethink the way we handle wrappers entirely.

I'm going to rethink it, too.

At first I create a new pull request for the get-type implementation in the generator. Perhaps we will have different branches which experiment object creation so it could be usefull in the develop branch.

badcel:
I just started to play a little bit around and get to know the topic. I'm going to create a new branch for this and create a separate draft pull request.

I would suggest that we both share our work in those drafts and try to find a possible solution which sattisfies our needs. I don't think that we will have like 2 competing drafts, but more like 2 prototypes on which we can build some final code.

PR #51, opened 2020-08-16
Add basic support for get-type methods

badcel:


firox263:
Everything looks fine to me 👍

I'm not sure whether this fits `develop` though, as the type dictionary as a concept only exists in `gobject-rewrite`, and I'd rather we didn't diverge from the typedict code (unless you want to change the typedict paradigm, then feel more than free to go ahead and merge).

Thoughts?

badcel:
I don't see how this is connected to the type dictionary? For me it is just a more complete implementation of the gir file in the `Sys` namespace. E.g. if someone else builds on top of the Sys namespace, the method would just be available.

firox263:
Fair enough. I'll go ahead and merge 👍

Issue #33, opened 2020-07-02
Vala to CS example


gavr123456789:
Were there any experiments with creating an object in Vala and importing it to C#?  
Or just forecasts of how well it will work.  
  
(Also interested in Meson)

badcel:
No I did not test vala. The generator expects a gir file to generate the wrapper. The vala compiler can output a gir file so I think it should work. There could be small issues with the types, if there are some bits which are missing in the type resolver.

I want to make this project like an entrypoint for C# developers to use the GObject-Stack. Do you have a special vala library in mind which is missing in this repository?  Otherwise I think it would be great to have an example for a custom vala library, so that everybody can write a wrapper for their own library, allowing to stay in the same C# GObject type hierarchy like the rest of the projects hosted here.

If there is a need there could even be a special dotnet tool to allow the generation of bindings from arbitrary gir files. 
If the bindings generated from a vala gir file is failing for you I would consider this a bug in this project.

gavr123456789:
> If there is a need there could even be a special dotnet tool to allow the generation of bindings from arbitrary gir files.
  
Do you already have any plans for this? That would be so cool.  
  
> Do you have a special vala library in mind which is missing in this repository?   
  
No, but I just know a few GTK libraries(for example [1](https://notabug.org/grindhold/libgtkflow), [2](https://github.com/lcallarec/live-chart) [3](https://gitlab.com/doublehourglass/dnd_order_list_box)) written in Vala. You can also use Vala to write application bottlenecks(since it generates C) and there still no better language to write something for GTK.

badcel:
> Do you already have any plans for this? That would be so cool.

No, it should not be very hard. There are a lot of projects which provide their own tools. But  with the current implementation it would only generate the lower layer. I would like any contribution in this direction.

> No, but I just know a few GTK libraries(for example 1, 2 3)

I didn't know these, they are very interesting!

Next I wanted to check if I could get s.th. to work with gtk4.

badcel:
There should be some kind of `dotnet tool` support to allow 3rd party developers to create new GObject C# bindings based on top of our stack. I changed the title to make this a little bit clearer.

badcel:
Another library is https://github.com/elementary/granite from elementary os

badcel:
#1078 is a little more concrete so let's close this.

PR #52, opened 2020-08-16
Custom GObjects

badcel:
Some remarks for this implementation:

- If a System.Type has no Method `Sys.Type GetGType()` it is assumed that this type is a wrapper
- The Binding class is an example implementation. The [GBinding](https://developer.gnome.org/gobject/stable/GBinding.html) has no `_new` function. Meaning it can only be created with an existing pointer or via the construct only properties.
- I don't use a type dictionary. I'm always asking the C typesystem for information. If this does not impact the performance I prefer it, because there is less code and no duplicated type information (C vs C#)

I suppose that in most cases it is okay, to just call the base constructor without any parameters. Because after the base constructor finished the object is constructed and we can set the properties of our instance afterwards in the C# constructor.

If there are some properties which are construct only (like in GBinding) we could call the `Object(params Prop[] properties)` constructor without the need to implement this kind of constructor in a subclass.

------------------------------
Edit:

**Todo**

- Register/FromExisting split
- Remove old properties code
- Rename property descriptors (remove 'Ex')
- Move Sys.Type to GObject.Type
- Generator 
  - properties 
  - events 
  - events via property indexer
- Samples
- Test other libraries (Gtk, libhandy, etc)
- rename ConstructProp to ConstructParam as this is an actual construct in GObject used for a near-identical purpose.
- No custom code in wrapper libraries. Properly wrap `Sys.Type` + `Sys.Value` in `Core`


badcel:
Missing;
- ClassInit: Call static constructor if called from C world
- InstanceInit: Call constructor if called from C world

firox263:
I am liking this approach quite a lot more. The type registration is fundamentally the same but you've managed to bypass the typedict, which I always considered a necessessary evil.

I agree that we'd be better off not having a type dictionary if we can avoid it, since we're basically just reimplementing GType.

I might have a go at handling object lifecycles (wrapping, lifetimes, toggle refs, etc) on top of this branch to see if we can stick with this going forward.

badcel:
I would suggest you create a new branch for your lifecycle management code on top of this. In this way we can discuss the pros and cons afterwards if i have s.th. working.

I think the code in this branch could be improved. I like your style of coding / documenting, it is very easy to read and follow along. I think whatever code we decide to take over, we should work over it together.

badcel:
I updated the code to generate the core classes partially. The code is compiling but not working because I had to remove the connections to the properties. I think this could be the start of our future gir.core base.

Generator/Templates/Core/class.sbntxt
```diff
@@ -8,11 +8,11 @@ namespace {{ namespace }}
...
     public partial class {{ name }} : {{ parent }}
     {
-        internal new static GObject.Type GetGType() => new GObject.Type(Sys.{{ name }}.get_type());
+        protected internal new static GObject.Type GetGType() => new GObject.Type(Sys.{{ name }}.get_type());
         
         #region Constructors
-        protected {{ name }}(IntPtr ptr) : base(ptr) {}
-        protected {{ name }}(ConstructProp[] properties) : base(properties) {}
+        protected internal {{ name }}(IntPtr ptr) : base(ptr) {}
+        protected internal {{ name }}(ConstructProp[] properties) : base(properties) {}
```
na2axl, line 15: There is no `params` keyword for construct properties, is this intended ?

```suggestion
        protected internal {{ name }}(params ConstructProp[] properties) : base(properties) {}
```

Generator/Templates/Core/class.sbntxt
```diff
@@ -0,0 +1,18 @@
...
+using GObject;
+
+namespace {{ namespace }}
+{
+    // AUTOGENERATED FILE - DO NOT MODIFY
+    
+    [Wrapper("{{ type_name }}")]
+    public partial class {{ name }} : {{ parent }}
+    {
+        protected internal new static GObject.Type GetGType() => new GObject.Type(Sys.{{ name }}.get_type());
```
na2axl, line 11: The return type here (GObject.Type) mismatch with the one in the type system (GObject.Sys.Type)

Libs/GObject/Core/Classes/Object.TypeDictionary.cs
```diff
@@ -0,0 +1,153 @@
...
+                }
+                
+                // We are a wrapper, so register types recursively
+                Console.WriteLine("Registering Recursively");
+                System.Type baseType = type;
+                while (!Contains(baseType))
+                {
+                    Console.WriteLine(baseType.Name);
+                    var methodInfo = GetGTypeMethodInfo(baseType)!;
+                    var typeid = (ulong) methodInfo.Invoke(null, null);
```
na2axl, line 119: Got an exception here:

```log
System.InvalidCastException: Unable to cast object of type 'GObject.Type' to type 'System.UInt64'.
```

Fixed with:

```suggestion
                    var typeid = (ulong) (Type) methodInfo.Invoke(null, null);
```

After added an explicit cast to `ulong` in the `GObject.Type` struct.

Libs/GObject/Core/Classes/Object.TypeDictionary.cs
```diff
@@ -0,0 +1,153 @@
...
+                    gtype = new Sys.Type(typeid);
```
na2axl, line 120: Why cast to `ulong` and reconstruct the GType ? If we are sure that a `GetGType()` method always return a `GObject.Type`, we can simply do:

```cs
gtype = (Type) methodInfo.Invoke(null, null);
```

Generator/Templates/Core/class.sbntxt
firox263: This one is intentional. We want to either move `Sys.Type` -> `GObject.Type` or return a `ulong` directly I believe. We'll need to do this before we're ready to merge

Libs/GObject/Core/Classes/Object.TypeDictionary.cs
firox263: I remember changing this at one point, not sure why it's still like this. You're right 👍

Libs/GObject/Core/Classes/Object.TypeDictionary.cs
firox263: Yep, you're right, it should be like that. There was another step at one point which needed the typeid, I must've forgotten to refactor that out.

Generator/Templates/Core/class.sbntxt
firox263: I don't think so.

@badcel?

badcel: You are right, this was not intendet. Please add the params. Yesterday I was focused on getting it in a compileable state. I did not do any review. Thanks for having a critical lool @na2axl :+1:

Generator/Templates/Core/class.sbntxt
badcel: Yes this is correct. I would say we do not expose either `ulong` or `Sys.Type` as a Type just `GObject.Type`. Otherwise it would probably be confusing.

badcel:
@na2axl Thank you for your review :superhero: . Today I'm not going to have much time to code. It would be nice if you could take over your proposals into your PR.

Libs/GObject/Core/Classes/Object.TypeDictionary.cs
na2axl: Resolved in 0148472

Libs/GObject/Core/Classes/Object.TypeDictionary.cs
na2axl: Resolved in 0148472

badcel:
**Todo**
- Register/FromExisting split
- Remove old properties code
- Rename property descriptors (remove 'Ex')
- Move Sys.Type to GObject.Type
- Generator (properties / events / events via property indexer)
- Samples
- Test other libraries (Gtk, libhandy, etc)
-  rename ConstructProp to ConstructParam as this is an actual construct in GObject used for a near-identical purpose.

See #57 

**Register/FromExisting split**
Rescue comment from @firox263 (https://github.com/gircore/gir.core/pull/58#issuecomment-682498696)
> This would apply to both properties and signals. Since for our wrappers, we are binding to a property/signal that has already been declared by GObject/Gtk/etc, I would prefer if there was a distinction made in the intent of the function. Even if they were fundamentally the same, having 'BindExisting' (for example) and 'Register' makes it very clear what the purpose of the descriptor is. The end-user would only ever use Register as existing properties are already defined by the bindings.

>I don't think we need to have individual classes (like Property and ExistingProperty implementing IProperty). This seems like overkill to me. However it would greatly simplify property registration, so it could be an option. Perhaps all we need is a simple boolean indicating whether it is user-defined or unmanaged-defined.

na2axl:
I will continue my work on properties. So I think I will take the 3 first tasks on the To-do. Can I work directly on this branch or I have to open another PR?

badcel:
I think we should stick with PRs as it is easier to see the diferences and give feedback.

badcel:
@firox263 
>  However it would greatly simplify property registration, so it could be an option. Perhaps all we need is a simple boolean indicating whether it is user-defined or unmanaged-defined.

Why would it greatly simplify? You would need to check for each property if it is binding an existing property or not, right? If someone calls the wrong `Register` method it would even lead to completely unexpected results. I think this would be bad API design, except if the user of a subclass does not see the `BindExisting` method somehow.

My proposal would be to check the `IsSubclass` method before any class initialization happened. If `IsSubclass` is true we need to install all properties of this type (with `BindingFlags.DeclaredOnly`). Probably we would need to recurseively install the properties of all subclasses in the hierarchy. In this way the user can't make a mistake and we decide per type if we install properties, not per property.

The good thing of your proposal is that it does not need recursion what is very good.

-------------------------

New thought: There are other GTK libraries out there (https://github.com/gircore/gir.core/issues/33#issuecomment-657268057).

If someone wants to write a C# wrapper for those classes, it could be possible that the dev is wrapping and subclassing the type at the same time just to avoid the work to create an additional wrapper type. In this case my suggestion to decide per type if properties are to be installed will fail.

Beforehand I assumed that a user of our libraries will not need the `BindExisting` method, which is why i disliked it to be available. But for this usecase it is necessary. I believe that the libraries should not force a developer into some design decisions (here: create a wrapper class, otherwise it will not work). So I changed my mind while I wrote this text and agree to the suggestion of @firox263. :+1:

na2axl:
I have a question BTW, if someone want to create bindings for an existing C library, he will also need to create native functions with `DLLImport`s ? And also create a `TypeDescriptor` for his C# type ?

If these both answers are yes, so the `IsSubclass` check will also work for user code.

The difference will appear when the user will instead create a **port** not a **binding**. The port case will be mostly a complete C# rewrite of the C library, in this case the user re-create every widgets/objects in C#, so even the properties are assumed to be created from C#. The binding case is already what we are doing now, and the user will have to create a `TypeDescriptor` to say to Gir.Core: Hey, this type is coming from the C world!

I think if we follow this path the `IsSubclass` check will be enough.

badcel:
@na2axl I'm not sure, perhaps you are right. I'm just going to write done a scenario as this is quite complex:

 - There is a small GTK Library containing `MyFancyButton` written in C
 - Our C# dev wants to extend `MyFancyButton`
 - He wants to add a bool property like `PlaySoundIfClicked`
 - He creates a class `MyFancyButton2` in C# and uses the type descriptor of `MyFancyButton`
 - He wraps some properties of the `MyFancyButton` C class in `MyFancyButton2`
 - He adds the `PlaySoundIfClicked` property

Now the `PlaySoundIfClicked` is property is in the same C# class like the C wrapped propperties of `MyFancyButton`. The `PlaySoundIfClicked` property must be registered in the type system, the properties of the C class not, as he is using the type descriptor of `MyFancyButton`. 

For this to work the user would need to provide some `DLLImports` at least for the `get_type` method.

----------------------

Another thought: We could provide tooling to generate the `Sys` namespace. If this generator generates a wrapper for `MyFancyButton`, too, the dev would not use the type descriptor of `MyFancyButton` but just inherit from `MyFancyButton` and we could decide property registration per class again.

But this would force our users to take our tooling. The question would be if we want so support like manually written bindings with suboptimal code structuring (adding features to a wrapped class).

-----------------------------

I'm not totally sure about this. This is how I understood our typesystem is going to work. What do you guys think?

na2axl:
@badcel 

Yeah your example is a nice catch of use case. I think we should limit the imagination of end users, by writing somewhere, maybe in docs or in the wiki, that if the user want to add functionalities on a C class bound to C#, he have to create the C# wrapper first, use the `TypeDescriptor` on that wrapper, then create the C# subclass next, and do any fancy things he want.

Also is it possible to install properties on an object after it's initialized ? I mean, if the user bind a native GNOME library to C# using Gir.Core, the `class_init` will happen in the C world right ? In this case can Gir.Core be able to install a new property ?

firox263:
I think by being too flexible with how we let binding-authors write binding code we risk over-complicating our code. Some of our guarantees rely on classes being generated in a specific way and defining specific attributes/functions/etc. For the purposes of keeping the end-user facing API as simple as possible, I think it's reasonable to expect that binding-authors use our tooling, like @na2axl suggests:

> I think we should limit the imagination of end users, by writing somewhere, maybe in docs or in the wiki, that if the user want to add functionalities on a C class bound to C#, he have to create the C# wrapper first, use the TypeDescriptor on that wrapper, then create the C# subclass next, and do any fancy things he want on that subclass only.

---

> **na2axl:** Also is it possible to install properties on an object after it's initialized ? I mean, if the user bind a native GNOME library to C# using Gir.Core, the class_init will happen in the C world right ? In this case can Gir.Core be able to install a new property ?

I believe this is correct. As far as I know, you can't define new properties on an already registered type. For this reason, binding-authors will probably need to subclass to add new functionality anyway.

> **badcel:** We could provide tooling to generate the Sys namespace.

I agree. Referencing `gtk-rs` again, they have a two tier system which creates the `mylibrary-sys` crate (same as our `MyLibrary.Wrapper`), and a partially defined high level `mylibrary` crate (same as our `MyLibrary.Core`). I would vote for auto-generating the `Sys` and the `<Class>.Generated.cs` files, and letting the binding author then define the rest of the partial classes. That way we know external bindings will respect our guarantees.

> **badcel:** But this would force our users to take our tooling. The question would be if we want so support like manually written bindings with suboptimal code structuring (adding features to a wrapped class).

I would say flexibility for the end user is more important than flexibility for people creating bindings, so I don't think we should support suboptimal bindings or adding features to wrapped classes.

---

Quickly on having a `BindExisting/Register` separation: I mainly want there to be some kind of distinction between registering a new property and binding an existing one. If we change how the function works purely based on the type of class (wrapper vs subclass), it feels a bit arbitrary and doesn't indicate the true purpose of the function.

It could be as simple as each function just wrapping a call to the constructor, but with a runtime check for whether it is a subclass or not (e.g. If you called `Register()` from a wrapper, or vice-versa, it would throw a `NotSupportedException` so it visibly fails, rather than silently does nothing).

na2axl:
I'm ok with the `BindExisting/Register` separation. Until now I've not minded that end users will have the possibility to create wrappers, so the necessity of the separation was not clear for me. But it will definitely be more meaningful for end users. BTW, `BindExisting` is a bit too blaahh name of this function... Maybe something soft like `Resolve` ? ^^

firox263:
> BTW, BindExisting is a bit too blaahh name of this function... 

Yup. It's getting annoying even typing it 😅

Maybe `Bind()` could be an option? I also quite like `Resolve()`.

badcel:
Okay I think you guys are right:
 - We should provide good tooling to make it easy for 3rd party library authors to create their own bindings (based on our GObject / GTK packages). I agree that new classes should just be subclassed from wrapped C classes to keep the type hierarchy clean.
- I like the disctinction between Binding / Registering to clarify the developers intent.  I have one more suggestion: What about `Register()` and `Wrap()`? For me resolve sounds like it is searching for the property.

na2axl:
`Wrap` is ok to me 👍🏾

badcel:
I added the todo list to the top of this PR

badcel:
I started the work on generating the properties automatically on the branch [GenerateProperties](https://github.com/gircore/gir.core/tree/GenerateProperties). I think this will take some time as I figure out the things to get the generator to work. If there are changes how we want to create the properties it should be quite simple to adopt.

firox263:
Comments carrying over from #70:
 - We currently use `ConstructParameter` for the class name. While this is more "correct" in C# terms, maybe we should use `ConstructParam` to map with GObject's own terminology, for consistency with our generated classes (plus if we somehow generate parts of GObject, we'll probably need to do this anyway?). Thoughts?
 - Since we use `Signal<SignalArgs>` a lot, I wonder if we can create a concrete base class which assumes `SignalArgs`, and then a generic class on top for differening SignalArgs types. This is probably unnecessary, but something to think about.

We're getting really close to finishing this up, so I think with the generator improvements, we can call this done.

I've managed to get some basic struct generation working, but I'm looking into how we could improve the generator to interpret Gir data in terms of "Ordinary Structs", "Disguised Structs", "Opaque Structs" (Similar to disguised but where `introspectable=0`; not entirely sure of the difference), and "Class Structs" (for vfuncs, etc).

I'm reasonably confident we can generate GObject virtual functions and the corresponding glue code automatically.

na2axl:
Hi @firox263 thank you for the review:

- For the `ConstructParameter` name I agree to use `ConstructParam` instead if it match better the GObject naming (also, it's short 😄)
- For the `Signal<SignalArgs>` problem, I've used this in the QuickStart app just because I have not wanted to create `SignalArgs` implementations, but if we plan to generate all these implementations for each generated signal we will have a `Signal<SignalArgs>` somewhere again ?

I will rename `ConstuctParameter` to `ConstructParam` directly in this PR.

firox263:
I think the plan was to use SignalArgs (or  BasicSignalArgs, etc) for signals with an object and user data pointer, which is the default. 

We'd then only generate a TSignalArgs if the signal takes additional parameters (like TreeView, with a TreeModel).

na2axl:
Ok I see, this will simplify a lot of things because I've feared that we will generate a lot of `SignalArgs` like:

```cs
// Partial class generated only to store SignalArgs implementations of a GtkButton
partial class Button
{
    // `click` signal
    public class ClickSignalArgs : SignalArgs {}

    // `activate` signal
    public class ActivateSignalArgs: SignalArgs {}

    // etc...
}
```

So yes I agree for cases like this we can create a concrete `Signal` implementation which is just a proxy to `Signal<SignalArgs>`.

Issue #57, opened 2020-08-21
[RFC002] Static GProperty Descriptors


na2axl:
# Static GProperty Descriptors

## RFC

Please see the RFC here: https://github.com/gircore/rfcs/blob/master/rfc-002-static-properties.md
It explains the proposal and some potential solutions.

## Summary

To allow Gir.Core be a part of the current C# UI ecosystem, the implementation of GProperties can be done using Static Properties, to match what it's done now using DependencyProperty in other libraries. For that we have the following goals:

* Declare GProperties of current Wrappers using Static Properties;
* Allow library users to declare GProperties of theirs Subclass using Static Properties;
* An high-level API will provide access to the GProperty value, to retrieve and edit it through the corresponding Property Descriptor.
* Property Descriptors must know the type (the C# type) of the GProperty;
* Property Descriptors must know the type (the C# type) of the GObject in which it will be registered;
* Property Descriptors must know the name of the GProperty;
* Property Descriptors must know if the GProperty is write-only, read-only, or read-write;
* Property Descriptors must allow an easy registration and initialization of the GProperty in the GObject;
* Property Descriptors must also describe child GProperties;
* All of these points must be implemented in an easy to use, DependencyProperty-like manner.

These goals can lead us to these achievements:

* Type System Integration with GObject, as discussed in #46, and described by [[RFC001]](https://github.com/gircore/rfcs/blob/master/rfc-001-type-integration.md).
* Solid MVVM binding model, as discussed in #34, and described by [[RFC003]](https://github.com/gircore/rfcs/blob/master/rfc-003-mvvm-binding.md).

badcel:
This looks good to me, we should see how the signals and properties integrate into: https://github.com/gircore/gir.core/pull/52

na2axl:
Good, I will rebase you recent changes on #52 into #58 and try to continue on this way then :-)

firox263:
One thing I would like to try and keep are strongly typed signal args. It makes the developer experience much nicer as it integrates with intellisense, etc

For example:
```cs
class DestroySignalArgs : SignalArgs
{
    string someRelevantArg;
    int someOtherRelevantArg;
} 
```

badcel:
:+1: 

I think you could just try to use the new properties / signals in `Window` and `Button`, this is enough for the small quickstart demo.

With this we could check the overall design and decide if this is what we want.

na2axl:
[...]

PR #58, opened 2020-08-21
Static GProperty Descriptor

na2axl:
This PR is related to the discussion on #57 about the implementation of property descriptors.

The PR is based on #52

For now, I've implemented it in an non-intrusive way, so existing code is not removed. I've also modified the GtkMinimal app to test the current implementation and everything seems to work as expected.

@gircore/coreteam 

---

**EDIT**: I'm using the GtkMinimal app as draft for testing purposes, but everything will be cleaned when the PR will be ready to merge.


na2axl:
Finally managed to implement basic support for properties and signals descriptors. The end user API seems very cool to me (you can see more in the Gtk QuickStart app):

```cs
var window = new Window("Title")
{
    DefaultWidth = 400,
    DefaultHeight = 300,
    Child = new Button("Close")
    {
        // Experimental for now, but very cool ^^
        [Button.ClickedSignal] = OnCloseButtonClick,
    }
};
```

na2axl:
For now child properties are not yet implemented, I'm facing strange API design and I fear that at the end there will no MVVM binding with child properties.

This is because, as said in the [official documentation](https://developer.gnome.org/gtk3/stable/GtkContainer.html) (see section **Child Properties**), child properties **are object properties that are not specific to either the container or the contained widget, but rather to their relation**. This means that there is *nowhere* we can add these kind of properties, neither in the container class nor in the contained class, like we are already doing with object properties.

Example:
```cs
// We can't declare a child property like this...
public static readonly ChildProperty<bool> ExpandChildProperty = ChildProperty<bool>.Register<Box>(
    "expand",
    nameof(Expand),
    get: (o) => o.Expand,
    set: (o, v) => o.Expand = v
);
// ...because there is nowhere to add this property
public bool Expand
{
   get => GetChildProperty(ExpandChildProperty); // Even this is invalid, we need to know the child widget
   set => SetChildProperty(ExpandChildProperty, value); // Invalid too...
}
```

So to fix that, I plan to do something like this:

```cs
// Declare child properties in the container class, since they remains to it
public static readonly ChildProperty<bool> ExpandChildProperty = ChildProperty<bool>.Register<Box>("expand"); // Only need to know the native property name

// Now to get/set child properties, just call anywhere 
box.GetChildProperty(ExpandChildProperty, child); // Get the value of the given property in `child`
box.SetChildProperty(ExpandChildProperty, child, value); // Set the given value to the given property in `child`
```

But this will unfortunately lead us to no MVVM on child properties 😞

firox263:
@na2axl The API looks brilliant!

I like that it's still regular Gtk and it doesn't stop the user from coding in traditional Gtk style, but at the same time they have a very expressive almost-declarative way of creating UIs.

As for child properties, that's the same way I would do it. I think we can probably deal with not having MVVM for child properties, since they are entirely layout related and won't be apart of Gtk4 anyway (which is only a few months away).

badcel:
@na2axl This looks absolutely great! I have to take a deeper look into it :+1: 

- I wand to think through your property descriptors, to get the complete picture and a feeling for the code.
- I never fully understood GTKs child properties. Now is probably the right time to look into it.

na2axl:
Happy to ear this @badcel, I will try to finish with child properties today to make the PR mergeable.

na2axl:
Seems that the samples are too much incompatibles with the code right now, only the QuickStart sample is working.

The code is ready to merge 😃

firox263:
No need to worry about the samples. We're planning on splitting GtkApp into multiple smaller samples, and GtkMinimal won't work anyway until I reimplement composite templates.

I will try review your PR over the next few days 👍

firox263:
I've only had a cursory glance so far, just a quick comment:

**We will probably need to split properties into two APIs for consumption by:**
 1. Wrappers (e.g. binding to an _existing_ property defined in C)
 2. End-users (creating a *new* gproperty from C#).

This wouldn't be too difficult on top of what we've already got. It could be as simple as having `Register()` and `FromExisting()`, where perhaps `Register()` could add the property to a static dictionary (or similar), which we then use in `class_init` to register with GLib.

If we went this route, I imagine we could simply find/replace `Register` to `FromExisting()/BindExisting()/etc` and then add `Register` back in later alongside composite templates and other `class_init` related functions.

I will have a more in-depth look a bit later. Looking good 👍

na2axl:
@firox263 

> This wouldn't be too difficult on top of what we've already got. It could be as simple as having Register() and FromExisting(), where perhaps Register() could add the property to a static dictionary (or similar), which we then use in class_init to register with GLib.

In the current design the `Register` method is doing nothing more than create a `Property<T>` instance, its real goal was to match a `DependencyProperty`-like syntax. I though that in the `class_init` you just need to _know_ the list of GProperties of the class, and according to the type of that class (Wrapper or Subclass), you will install these properties or not in the GType, isn't that ? If it's not the case so there are some steps I have missed.

> On another note, while I am a fan of the implicit methods, the Register function does seem a bit too verbose. Looking at Avalonia as an example, perhaps we could make properties read/write by default. Not sure if this is possible?

Happy to see you've checked that ^^. My design is highly inspired from Avalonia's [direct properties](http://avaloniaui.net/docs/authoring-controls/defining-properties#direct-avaloniaproperties) since they match exactly our case. The syntax is verbose, for sure, but I see his benefits for when we will implement MVVM bindings. The `get` and `set` methods from the `Property<T>` will be used on data binding time, otherwise we will do reflections and this can be overkill...

badcel:
@na2axl Could you merge with the current `CustomGObject` branch?

badcel:
> I though that in the class_init you just need to know the list of GProperties of the class, and according to the type of that class (Wrapper or Subclass), you will install these properties or not in the GType, isn't that ?

I'm not sure about this. We probably need to test it. I think what @firox263 means is if our GObject is just a wrapper around e.g. a `Gtk.Button` and the Button is created all the propperties are already installed, because the type is defined and we don't need to register anything (in case the object is a pure wrapper).

If we register a subclass of a subclass at first we will register the parent type and afterwards our own type. So each class does only need to care for their own properties and not about their parents properties? If it would be like this, it would be very nice, because it would be very simple.

To differentiate both cases (don't register anything / register only the properties of the class) we could just take our `WrapperAttribute`. Than we woudn't need a differnt type of properties. If we would use a different type of Property the benefit would be that we wouldn't need to search for the attribute. But from my point of view this is just a duplication of information:
 - Wrapper attribute: This is a warpper class
 - Special property class: This is a wrapped property of a wrapper class

As the `GetType` method is static and the `WrapperAttribute` is part of the class definition we could create some kind of static `TypeInformation` like our property descriptors, which don't describe the property from C# point of view, but the type. We could bundle the `GetType` method together with the name of the wrapped C GOBject class and get easy access to the desired information with one call.

badcel:
I updated my PR to your branch so it is not using _InternalsVisibleTo_ and I added a type descriptor class to describe a class.

na2axl:
@badcel I've checked that, I'm OK with the `TypeDescriptor` API, it's better than the `Wrapper` and allow us to cache the GType, so no unnecessary calls ^^. But I have seen that the `TypeDescriptor` uses `Sys.Type` instead of `GObject.Type`, is it intended ? Since @firox263 have said it will be the `GObject.Type` instead which will be exposed to the user.

Also when you will be done will the PR just say me to merge.

badcel:
Yes it is intentional. As the type dictionary is based on `Sys.Type`. `GObject.Type` is just another wrapper around `Sys.Type` so I did not use it.  It show's the `Sys.Type` a little more than the old `GetGType` method did (as it is now `protected`).

If we don't want do leave a trace of the `Sys` namespace I would convert it to `GObject.Type` (including the type dictionary).

@firox263 what do you think?

firox263:
> @firox263 what do you think?

My preference is that a library's `Sys` namespace should only be visible to that specific library and shouldn't be exposed/"leaked" outside of the assembly. If there's something in the library that other libraries also need, it's probably a sign that API should be public anyway.

badcel:
I made some more changes to my PR. I made the type descriptor private as it should never be used from another class. The type is now only searched in the current class and ignoring the class hierarchy as this could otherwise lead to wrong results.

I think we can merge it now.

Afterwards I would like to merge @na2axl code into my `Custom GObjects` branch so we can create new pull requests inside the gir.core repository and don't need to fork the repo.

firox263:
As for merging this, @na2axl you've done a brilliant job making this not conflict with #52.

I don't really have the time at the moment to properly review this line by line (and won't until end of next week), but I'm satisfied with how this all works on a high level, **with the exception of the Register/FromExisting functions which I'd like to sort out first**.

We're going to need to do a lot of cleanup on the codebase once merged into #52:
- Remove old properties code
- Rename property descriptors (remove 'Ex')
- Move `Sys.Type` to `GObject.Type`
- Generator
- Samples
- Test other libraries (Gtk, libhandy, etc)

Perhaps in this case, if you're happy with this @badcel we can merge this into #52 once we've resolved the wrapper vs subclass registration and continue working on it incrementally from there.

---

A quick word on the whole FromExisting/Register thing:

This would apply to both properties and signals. Since for our wrappers, we are binding to a property/signal that has already been declared by GObject/Gtk/etc, I would prefer if there was a distinction made in the intent of the function. Even if they were fundamentally the same, having 'BindExisting' (for example) and 'Register' makes it very clear what the purpose of the descriptor is. The end-user would only ever use `Register` as existing properties are already defined by the bindings.

I don't think we need to have individual classes (like `Property` and `ExistingProperty` implementing `IProperty`). This seems like overkill to me. However it would greatly simplify property registration, so it could be an option. Perhaps all we need is a simple boolean indicating whether it is user-defined or unmanaged-defined.

---

On a completely unrelated note, we should rename `ConstructProp` to `ConstructParam` as this is an actual construct in GObject used for a near-identical purpose.

firox263 (approved):

na2axl:
@badcel I've just merged your changes and into this PR

PR #62, opened 2020-08-23
Add contribution guidelines

### LINKED ISSUE #59: Code style
We should define a code style we all agree on to avoid later discussion and ensure a consistent code base for gir.core This could be an RFC.

badcel:
Fixes #59


badcel:
@firox263 If you agree, too, please merge.

PR #64, opened 2020-08-24
Adds GObject reference to DBus sample

knuxbbs:


badcel:
@knuxbbs : Thank you for your pull request. You are right the project references are incorrect. Actually there only needs to be a reference towards `Gio` for the DBus sample to work. If you fix this I'm happy to merge.

PR #65, opened 2020-08-29
Fix for non-public constructors

firox263:
Fix for #52, allowing us to call protected and private `IntPtr` constructors (we cannot retrieve the Content Area of a Gtk.Dialog as Container's constructors are protected).

Also changes `o.Handle` to `handle`, as `o` is undefined in this call. I've switched `Activator.CreateInstance` to `GetConstructor` as it is more explicit - we could add error handling if the constructor is not found and notify the user?

### DIFF
--- a/Libs/GObject/Core/Classes/Object.cs
+++ b/Libs/GObject/Core/Classes/Object.cs
@@ -226,12 +226,17 @@ namespace GObject
                 throw new InvalidCastException();
 
             // Create using 'IntPtr' constructor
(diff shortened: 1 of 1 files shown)


badcel (approved):

firox263:
A thought:

We should probably check if the GType of generic type 'T' is equal to the gtype from the pointer. That way, we can avoid the expensive lookup in the type dictionary for cases where we know the type we want (e.g. `Dialog.GetContentArea`).

PR #67, opened 2020-08-30
Remove implicit casts

### LINKED ISSUE #53: Remove implicit casts from Object
Remove implicit casts from Object: This increases readability.

```C#
public static implicit operator IntPtr (Object? val);
public static implicit operator Object? (IntPtr val);
```

badcel:
Fixes #53

### DIFF
--- a/Libs/GObject/Core/Classes/Closure.IDisposable.cs
+++ b/Libs/GObject/Core/Classes/Closure.IDisposable.cs
@@ -2,7 +2,7 @@ using System;
 
 namespace GObject
 {
-    public partial class Closure
+    internal partial class Closure
     {
         protected bool disposedValue = false;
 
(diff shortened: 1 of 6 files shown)


na2axl (approved):
Everything is ok for me, the issue I've mentioned can be resolved in another PR.

badcel:
> Maybe it will be better to have a talk about the project structure after the main goals in #52 are achieved?

Sure. To get familar with your thoughts you could just open an issue and describe them. Afterards we can find a date to talk this through.

firox263:
> I've seen that the way we are separating projects now (Name.Core and Name.Wrapper) give us some limitations

I think our current arrangement is not intentional as such, but rather because we just built on top of what was already there (initially we had pure ABI structs in `Name.Wrapper`, hence the strange split)

In my opinion, we should keep the wrapper entirely auto-generated and move Sys.Value, etc to the Core namespace (we're already doing this with `Sys.Type` -> `GObject.Type`). This way, we get around any limitations while still keeping the nice separation between unmanaged functions and our 'safe' API.

firox263 (approved):

PR #69, opened 2020-08-30
Create Value struct in Core project

badcel:


badcel:
This is obsolete as we will merge wrapper / core layers.

PR #70, opened 2020-08-31
Cleaning properties and signals

na2axl:
PR to continue the work on #57.

## Checklist

- [x] Remove old properties implementaion
- [x] Clean dummy properties with `Ex` suffix
- [x] Rename `ConstrucProp` to `ConstructParameter`
- [x] `Property<T>.Register()` and `Property<T>.Wrap()` methods
- [x] `Signal<T>` implementation where `T` is a `SignalArgs`
- [x] `Signal<T>.Register()` and `Signal<T>.Wrap()` methods

Issue #68, opened 2020-08-30
Project structure (Core / Wrapper)


badcel:
> I've seen that the way we are separating projects now (Name.Core and Name.Wrapper) give us some limitations (by @na2axl)

I think our current arrangement is not intentional as such, but rather because we just built on top of what was already there (initially we had pure ABI structs in `Name.Wrapper`, hence the strange split)

In my opinion, we should keep the wrapper entirely auto-generated and move Sys.Value, etc to the Core namespace (we're already doing this with `Sys.Type` -> `GObject.Type`). This way, we get around any limitations while still keeping the nice separation between unmanaged functions and our 'safe' API.

_Originally posted by @firox263 in https://github.com/gircore/gir.core/pull/67#issuecomment-683391417_

badcel:
You are right. In the beginning of this project I did not implement the (full) generation of structs as they were not very important to get things working. Primarily I'm missing the creation of struct fields, which i added manually. But afterwards I added convenience functions which do not belong into the wrapper.

We should **not**  have any custom code in the `Sys` namespace but just allow access to the C world. Custom code should only be in the `Core` namespace, meaning we should just create their `Core` counterparts.

This does clean up our project internals and alligns to the rest of our work. I think in this way we can solve our ambiguities, too.

I'm going to add an additional TODO point into #54.

badcel:
As we have decided to switch to a 3layer design in #71 

We need to decide how the 2nd layer should look and which features are to be implemented.

In the 2nd layer we want to wrap GTK closely with as few differences as possible. To get this working we have to fix naming issues.

My suggestion would be:
 - Use prefix `On` for events (e.g. `OnAdd`)
 - Use suffix `Prop` for properties (e.g. `Label.LabelProp`). 
 - Keep the class names as they are defined in the gir files.

As we follow a 3 layer design I fully agree to @firox263 that in this case the 2 lower layers must be autogenerated so we have no problems following up in the 3rd layer if s.th. changes. This means methods would be autogenerated:
 1. The method signature almost always involves the handle of the object as the first parameter. We can probably hide this in the autogenerated code. 
 2. Some methods contain an error pointer, meaning that an exception occured. We can probably catch this in the generated code, too and throw an exception instead.

But all other parameters will stay as is as we can't make any assumptuns regarding them. (At least I don't know any other rules we could use to interpret some  parameters.)

Regarding properties and events I would suggest that we just use what we already have via @na2axl design and support `INotifyPropertyChanged`.

[...]

Issue #63, opened 2020-08-23
Toolkit layer / Higher level C# integration


badcel:
This is a collection of ideas which could lead to better integration of this library into thje C# eco system. Those ideas are low on the priority list, as long as we are working on the basics of gir.core.

- Gio.SimpleAction: It should be able to create an action which is supporting the `ICommand`interface

badcel:
The scope of GirCore has changed in a way that it's not trying to be as idiomatic as possible but builds a bridge between GObject and dotnet.

If there is need for special implementations they can be part of a separate project.

PR #71, opened 2020-08-31
Generate properties and events

badcel:
Push the generator project to support generating properties and events.

## Checklist

- [ ] Generate properties
- [ ] Generate events
- [ ] Support events via property indexer


badcel:
@firox263, @na2axl : In the current code there are some problems resulting from girfile + C#:

 1. The Gtk `Label.Label` property: It is not allowed to be named exactly like the class
 2. There are properties which are named like functions in the type hierarchy: `AppChooserWidget.ShowAll` (property) vs. `Widget.ShowAll()` (method). The property hides the method and both member do fundamentally different things (show applications in  a single list vs. show all widgets). We would need the new keyword but would still hide the method.
 3. There are several classes which define properties like a parent class. There is no clue in the GIR that this is the case. `IconView.Margin` vs. `Widget.Margin`. We must insert the new keyword to hide the parent method.

I can probably figure out point 3. I need to look for hidden properties and insert the new keyword.

Regarding 1 + 2 this is a little bit more of a problem, because this destroys the original api.

Do you have any suggestions?

na2axl:
Hi @badcel,

I've experienced something similar in the Quickstart app when implementing events, for example the `Add` of the `Container` class conflicts with the `Add` method of the widget class...

Having a quick look on GtkSharp, I see they have added the `Prop` prefix to property names who conflicts with other members, we can do the same, but yes this will not be C#-beautiful.

na2axl:
@badcel I think this PR should be merged in #70 instead of #52 since you will need my work on cleaning old code, etc...

badcel:
@na2axl : I think we should kepp the PRs separated. In this way we can work independently. Your pull request is for sure going first into #52. Afterwards or in between I'm going to pull your PR into mine to be able to merge with #52 in the end.

@na2axl @firox263 Here comes my wall of text: :smile: 

Regarding the naming problems: Those are pretty major for me, as they decide the general direction in which this project should be heading. I think we should do our bindings in the right way with as few sacrifices as possible in general.

My original intent for this project was to create an API as nice and as convenient for C# developers as possible. This would allow name changes, restructuring of classes and so on. The end result would be an API which is partially related to the official GTK API but it would be an almost 100% solution for a C# dev who searches a UI Framework and likes our model. One benefit of this thought was that even if GTK would change underneath we could keep the API stable for the end user.

Now we changed the general route of the project to be a wrapper over GTK, which is fine for me as we can always put an additional layer on top of it if needed. But if we wrap GTK it would be important for me to avoid workarounds. If we have to rename some properties due to language restrictions this would be not consistent. Either all properties are suffixed "prop" or all properties are lower case or perhaps s.th else, like all classes start with a "G". I just want to create a consistent "product" which is as reliable and as consistent for the end user as the original GTK is.

The next question in the devlopment of this would be: _Do we generate the methods, too?_ I'm not very sure about this as there are methods which we probably don't want in our API as they don't fit C# (e.g. callbacks instead of async/await). But if we don't take them over how to decide which we provide? Are they hand written? Do we have some config file? Actually we must take over everything in this case and be a wrapper at all costs. No matter how good / bad the API is for a C# dev, because we wrap: That's our job. The benefit would be we would be 100% autogenerated without sacrifice, but without(*) nice features, too. **But** In this way we could again be a solid foundation for another library on top. We do one thing and we do it well.

But if we start to somehow be a little bit of both worlds (100% auto generated vs 100% the perfect lib for a C# dev) the API is not very atractive to C# devs. The API will always be in some "in between" state and we will probably have configuration files to maintain, make the code complex because of workarounds, etc.. 

PyGTK does not sacrifices on my points mentioned above:
 1. [Label has a property `label`](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Label.html#Gtk.Label.props.label)
 2. [AppChooserWidget has a property `show-all`](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/AppChooserWidget.html#Gtk.AppChooserWidget.props.show_all) but allso a method [`Widget.show_all()`](https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Widget.html#Gtk.Widget.show_all)

The problem is that we don't know which problems will arise due to C# language restrictions in the future. If we want to follow the current direction and be a wrapper over GTK we should create some general naming concept to avoid conflicts and be consistent. 

If you look at GTKSharp it sacrificed on those things:
 - On the `AppChooserWidget` you can't call `Widget.ShowAll` even if it is a widget what is not very good
 - There are pull requests regarding corner cases which were not met by the generator script. Either types or leaks or whatever.

I hope that if we follow very strict rules to be reliable even in those corner cases and create a great foundation of libraries.

----------------------------
(*) there could be some nice features like the property / event system by @na2axl, but the target would not be to give a C# dev a "pleasant" api but a wrapper around GTK.

na2axl:
@badcel 

> Now we changed the general route of the project to be a wrapper over GTK, which is fine for me as we can always put an additional layer on top of it if needed. But if we wrap GTK it would be important for me to avoid workarounds. If we have to rename some properties due to language restrictions this would be not consistent. Either all properties are suffixed "prop" or all properties are lower case or perhaps s.th else, like all classes start with a "G". I just want to create a consistent "product" which is as reliable and as consistent for the end user as the original GTK is.

I understand this need. But I think we have the possibility to not be limited on doing a simple wrapper matching exactly the original GTK API. Even if we are doing this we will hit other limitations (the `Add` event problem in the `Container` class for example). GTK have unfortunately not  the same architecture of a C# project, properties can be but are not really mappable exactly to C# (GTK doesn't have the property inheritance limitation for example). GtkSharp have done heavy concessions for that fact (renamed properties, removed properties, etc...). So having the goal to _wrap_ exactly the GTK API (and other GNOME libraries) will head us to hit the same problems with the cancelation of those awesome functionalities we have in mind. For that, back to the previous paragraph.

> My original intent for this project was to create an API as nice and as convenient for C# developers as possible. This would allow name changes, restructuring of classes and so on. The end result would be an API which is partially related to the official GTK API but it would be an almost 100% solution for a C# dev who searches a UI Framework and likes our model. One benefit of this thought was that even if GTK would change underneath we could keep the API stable for the end user.

This IS the goal we MUST achieve from my point of view. C# is C#, C is C, workflows are drastically different. We don't have to forget that if peoples choose our framework, it will be for the need to build **cross platforms apps using C#**, not essentially **GTK apps using C#**. In that way, we can provide a custom API partially related to GTK, with all these nice features we can provide, with custom classes/properties/events and so on, but which at the end produce a GTK application. Many other frameworks are doing the same, the first ones who comes in my mind are `Xamarin.Forms` which the `MonoMac` edition is GTK based (on the old and not maintained [Gtk#](https://github.com/mono/gtk-sharp) repo), and the new upcoming [.NET MAUI](https://github.com/dotnet/maui) which will arrive with .NET 6, even this one is based on the old Gtk# repo, but a discussion [here](https://github.com/dotnet/maui/issues/71) suggest to use GtkSharp instead.

Those two projects are great examples of what it's possible by implementing a custom API backed by GTK. Following this goal will certainly oblige us to do more work, but will allow us to implement every fancy functionalities that C# devs want. Now talking about methods.

> The next question in the devlopment of this would be: Do we generate the methods, too? I'm not very sure about this as there are methods which we probably don't want in our API as they don't fit C# (e.g. callbacks instead of async/await). But if we don't take them over how to decide which we provide? Are they hand written? Do we have some config file? Actually we must take over everything in this case and be a wrapper at all costs. No matter how good / bad the API is for a C# dev, because we wrap: That's our job. The benefit would be we would be 100% autogenerated without sacrifice, but without(*) nice features, too. But In this way we could again be a solid foundation for another library on top. We do one thing and we do it well.

This will finally depend on what goal we pursue. If we want to make this project a wrapper, methods can be generated and some fixes can be added later manually. If we make this project a GTK based UI framework, methods can be but don't really need to be generated, because raw methods will be directly called from the high level API. Now a sentence is interesting in this paragraph.

> But In this way we could again be a solid foundation for another library on top.

This is a _one kill two birds_. Yes, we can provide a raw, hard and strong wrapper over the GTK API, and then in another library use this wrapper to provide the UI framework we want, with all nice features we can add. This way, those who want a wrapper, just choose the wrapper lib, and those who want to develop GTK-based cross platforms apps using an high-level C# workflow just choose the UI framework. But again, this can be a lot of work.

**My POV**: I think with the current state of Gir.Core we can bring it to a good high-level UI framework with a custom API without too much headaches, the only drawback (but for me it's not one ^^)  will be to doesn't follow exactly the official GTK API.

badcel:
@na2axl : Thank you for your feedback, it is very welcome as it shows in which direction you want to get with this project and it is matching my intention to provide a GTK UI framework which is as good as possible from a C# point of view.

The question is just which way to go:
 1. I think we could overcome all the naming issues. E.g we could just suffix all properties with `Prop` and all events with `Event`. In this way we could keep auto generating (including methods) and be consistent as it is not important that this is not idomatic C# because we will build on top of it. This would allow other frameworks to be build on top of this "GTK middle layer".
 2. We skip the "middle layer" (what my original intent was) and build directly on the raw API (the wrapper projects) and create the best C# GTK experience we can imagine. For this we would need to check if this could be partially auto generated, as we probably have to take over the inheritance chain in general (but would not need to provide all available methods / properties in it).

For me both directions are valid.

The first provides an additional layer others can use what I like. The work for a high level framework would be postponed. I even don't know if we need one, e.g. if other frameworks like MAUI / Uno would adopt this middle layer we can just save the work. They probably won't have as much features as we could have, because all those frameworks are multi platform and must somehow find the smallest denominator of all toolkits they use.
On the other hand GTK tries to be cross platfom and is getting better and better. So the question would be why do we want to accept to be used in a trimmed version from a huge cross platform toolkit, which itself uses several different toolkits, if our favourite toolkit is already cross platform? We only miss out on iOS and android, but at least for me `libhandy` is the right aproach to do mobile applications.

The second aproach saves the work on the middle layer and we could directly start with the great API but we miss the chance to build a higher level foundation. Nevertheless other projects could still be build on top of the wrapper-projects (in the same way we would) as those projects are probably the layer which enforces as few restrictions as possible on a newly build project.

firox263:
Not to be outdone, here's my wall of text to end all walls of text.

I'm going to try and address everything in this so fair warning, it's very long 😅

---

For me, it is very important to both establish what the purpose of the library is, and to understand that whatever option we pursue will never satisfy both 100% C# native library and 100% accurate bindings. Gtk itself isn't one monolithic library, but rather a collection of high and low-level building blocks that the user chooses to use at their discretion. For example, I could make a text editor using just Gtk, but if I wanted to create a custom editor widget, I may well (and likely _will_) need to use `Cairo`, `Pango`, and `Gdk` directly.

**Interop**
Personally, I think of GObject (in the context of language bindings at least) in a similar manner to COM, in that it provides an interface between differing languages and type systems, and while not elegant, is a necessary evil in bridging C# and OOP-style C.

Important to note is that gir.core (as I currently understand it) isn't just a GUI toolkit, but rather a family of GObject-based libraries, including Gtk, but also GStreamer, DBus, etc (Eventual GStreamer support is particularly important to me). Consequently, we cannot base our decisions for the 'Core'-layer API solely on the assumption that we are creating GUI.

Additionally, we don't want to diverge too far from the library's GObject exposed interface as we need not only great C-C# interop, but also C#/C# interop (using GObject classes). If we follow GObject reasonably closely, it means we get C#/C# "for free" between different GObject-based libraries.

I am thinking of the following use-case:
 - We define bindings for Gtk
 - Author A defines bindings for 'GObject-based Lib A'
 - Author B defines bindings for 'GObject-based Lib B'
 - End User makes a C# app/library on top of Gtk, Lib A, and Lib B.

If we follow a set of rules for the actual binding code, we get A-B interop "for free", and there are no surprises for the end user. As this does not stop us from _also_ exposing high level abstractions, the end user still gets a nice C#-style API, but can use the `Core` API directly should they wish to.

**Stability**
> **badcel:** One benefit of this thought was that even if GTK would change underneath we could keep the API stable for the end user.

Personally, I'm not a fan of this idea, in the sense that we'd be steadily accumulating legacy code in order to keep a stable API/ABI. Of course we'd want the library to provide a great stability guarantee (otherwise nobody would use it), but I think for this, we should match the stability guarantees of wrapped libraries. So if upstream deprecates something, we'd deprecate it too, and we'd never diverge too from upstream to the point that the libraries become unmaintainable.

If we then built higher-level abstractions on top of our 'Core'-level code, we are free to change the implementation while providing a stable (but perhaps more limited) API for end-users in traditional C# style. If the user needed deeper access for more advanced features, they are able to use the less traditional 'Core'-level API directly to build their own higher level abstractions, with the same stability guarantees you would get if using Gtk directly.

**C# Ecosystem**
> **na2axl:** This IS the goal we MUST achieve from my point of view. C# is C#, C is C, workflows are drastically different. We don't have to forget that if peoples choose our framework, it will be for the need to build cross platforms apps using C#, not essentially GTK apps using C#. In that way, we can provide a custom API partially related to GTK, with all these nice features we can provide, with custom classes/properties/events and so on, but which at the end produce a GTK application.

Fully agree. You make an excellent point here about the "end product" that the developer will use. We are fundamentally trying to let C# developers create GUI powered by Gtk, with the expectation that the API will put C# first and foremost. One point I would like to raise however is that we _should not limit_ users from accessing the raw Gtk API should they need to. My biggest frustrations with C# GUI in the past is they are primarily widget-catalogs, and implementing your own widgets is more difficult than it should be.

> **na2axl:** My POV: I think with the current state of Gir.Core we can bring it to a good high-level UI framework with a custom API without too much headaches, **the only drawback (but for me it's not one ^^) will be to doesn't follow exactly the official GTK API.**

For me this is a drawback in a few ways. I think we can safely diverge from Gtk's API when it is necessary (e.g. property naming, where we couldn't ever match if we tried) or convenient (e.g. using `System.DateTime` over `GDateTime` because that's simply common sense). Ultimately, I would prefer remaining consistent with the upstream API, as the further we diverge, the harder mapping between becomes.

> **na2axl:** Many other frameworks are doing the same, the first ones who comes in my mind are Xamarin.Forms which the MonoMac edition is GTK based (on the old and not maintained Gtk# repo), and the new upcoming .NET MAUI which will arrive with .NET 6

On an aside, this is an interesting point for me. Are we aiming to compete with `Xamarin.Forms`, etc? Or should we instead provide a solid foundation for both existing toolkits (for linux GUI support) and user code, and then build a higher-level 'Gtk C# Toolkit' on top to implement many commonly used patterns?

**A Higher Level Library?**
> **na2axl:** This is a one kill two birds. Yes, we can provide a raw, hard and strong wrapper over the GTK API, and then in another library use this wrapper to provide the UI framework we want, with all nice features we can add. This way, those who want a wrapper, just choose the wrapper lib, and those who want to develop GTK-based cross platforms apps using an high-level C# workflow just choose the UI framework. But again, this can be a lot of work.

Ironically enough, I think killing one bird with two stones might be our best option. I believe the biggest challenge we will face in this project is _maintaining_ up to date bindings for the many libraries, and the closer we map, the more achievable this will be.

> **badcel:** I hope that if we follow very strict rules to be reliable even in those corner cases and create a great foundation of libraries.

> **badcel:** But In this way we could again be a solid foundation for another library on top. **We do one thing and we do it well.**

Personally, this is my preferred option. We create bindings that are closely mapped, technically sound, and relatively easy to maintain (e.g. no 2000 line XML file 😅), exposing the functionality we need for higher level abstractions. Then, we use this 'solid foundation' to create a 'C#-native' toolkit which acts comparably to other frameworks.

**Conclusion**
My proposal would be as follows:
 - **GirCore Wrapper Layer:** Autogenerated function calls (`Gtk/Wrapper`). Never used outside the Assembly.
 - **GirCore Core Layer:** Reasonably high-level GObject bindings (e.g. `Gtk/Core`) that are perfectly usable by themselves, provide all the tools to integrate deeply with C#, but perhaps do not implement them directly. We design this in a way where the user can implement the features they need. Combination of Autogenerated definitions and manually written code (with `Class.cs` and `Class.Generated.cs`). Other toolkits like Xamarin.Forms/MAUI/etc can use this as a backend.
 - **Toolkit/Core Library:** A high level toolkit built on top of gir.core. This has a stable API, high level C# constructs, the deep integration with C# you would expect from a modern toolkit, etc. It also gives us the flexibility to abstract over changes to Gtk API and perhaps even ease the transition between major Gtk versions. Maybe this could take the form of a helper library, rather than a complete toolkit, where we implement many common patterns (e.g. It could sit 'along side' Gtk instead of on top of it). It would provide an opportunity to implement Gtk-based Reactive Programming as @na2axl proposed, while allowing for traditional procedural/OOP-style code as well.

My main reasoning is that if we map closely, we can considerably reduce time spent fixing things and playing 'catch up' with the libraries we wrap. Many of the big Mono/Gtk# programs implemented their own "Widget Libraries" on top of regular Gtk# to be better integrated with the language. Our high level toolkit could take this position, and provide implementations of commonly used C# patterns which the user can use at their discretion.

I hope I've presented my point clearly enough. Can't say I intended to write this much haha.

Would very much like to hear your thoughts.

@badcel Just saw your new comment, I would (personally) prefer option one. But I too want to "create the best C# GTK experience we can imagine", so I think it's important we find a balance between the two approaches. For signals, we could use `On<SignalName>` (e.g. 'click' becomes `OnClick`). But I agree, consistency is most important here.

na2axl:
@firox263 your comment was not long for nothing hahaha, you have convinced me and I agree with your proposal. But this can drastically change the path we are following now, because if I understand what you said:

- The **Wrapper** layer is just the collection of raw bindings to the different GNOME libs we have (Gtk, GObject, DBus, etc...). This is already done with the generator of @badcel so no problem with this.
- The **Core** layer is the high-level C# project which wrap the **Wrapper** layer. My problem starts here... If we provide this library to end-users or other frameworks like MAUI as a wrapper over several GNOME libs, do we have to add other functionalities like MVVM binding, INPC, etc ? (I think we must at least add INPC), signals/properties indexers and other stuffs ? Even the idea of @badcel in #63 to for example make a `SimpleAction` implements `ICommand`, do we have to implement this in this layer ? Because we have...
- ...The **Toolkit** layer which is our UI framework, properly said. If the **Core** layer already has MVVM and other fancy stuff, what will be the main goal of this layer ? In this case I fear it will be just a bunch of classes renamed (eg `Label` to `Text`, `VBox` to `Column`) with custom properties.

Maybe we need to review the current path we are following because with current developments and ideas the **Core** layer is pretty similar to the **Toolkit** layer. If we want to separate things like you have proposed, we can:

- Let the **Wrapper** layer as is, it already have lot of pointers and ulong types that we rarely use in C#
- Limit the **Core** layer to be a consumable C# wrapper over all the GNOME libs we provide, by following tightly the official API. We can at least add INPC and signal/properties indexers.
- Implement MVVM, commands, custom widgets and more in the **Toolkit** layer.

So, yes, I think with this we have a solid foundation for what Gir.Core can be in the future...

badcel:
I completely agree to your last posts @firox263  and @na2axl . I think from a technology point of view it is the best solution to go with a 3-Layer design. Than it is decided, we do 3 layers. :rocket: 

So we can continue to work on the cleanup and I can continue with the generator but we have to fix the naming issues and define how much "features" the 2nd level should have and what is part of the 3rd layer. I think we should continue our discussion on #68 so we don't loose our comments if this gets merged.

badcel:
Closed in favor of #73

PR #73, opened 2020-09-26
Generate v2

badcel:
## Checklist

- [ ] Generate properties
- [ ]  Support Write only properties
- [ ] Generate events
- [ ] Support events via property indexer


na2axl:
Hello @badcel, well done with this 🚀. I have a question with properties generation, seems that the generator can handle read-only and read-write properties, but what about write-only properties ? Like the `Child` property of the `Container` class...

badcel:
> Hello @badcel, well done with this . I have a question with properties generation, seems that the generator can handle read-only and read-write properties, but what about write-only properties ? Like the `Child` property of the `Container` class...

You are right, support for write only properties is missing right now. I addeed it to the todo list.

PR #74, opened 2020-09-30
Generate gobject

badcel:
I started with the GObject generation. This is compiling without the samples but a lot of things are still missing. 

I deleted the old Core / Wrapper folders and started porting the code over to the new files.

If you want to you can work on this, too and directly commit into this branch. I think we need to reach feature parity with the GenerateV2 branch before taking a look at the next library.

- [x] Merge Wrapper and Core
- [x] Generate global functions in Global.cs
- [x] Keep the class names and type hierarchy as they are defined in the gir files. (autogenerated)
- [x] Support INPC
- [x] Support events / signals via descriptors, use the prefix On e.g. OnAdd (autogenerated)
- [x] Put native methods in an internal class to get the following Syntax: Container.Native.add(widget)
- [x] Put generated core code in a *.Generated.cs


firox263:
Pushed some small formatting tweaks to the class template.

Everything's looking good to me. Not sure I'll have the time to code this week, but is there anything you'd like help with
in particular if I get the chance?

badcel:
@firox263 you could try to generate events with the "on" prefix.

I think right now we should not introduce the "Native" class because we don't know how we will handle methods in the end.

I'm going to try to generate more projects.

If this is done we can try to get the samples working again.

firox263:
@badcel I've got signals generating and it seems to be correct. One problem however is that the Signal.cs in this branch is quite different to the one in `CustomGObjects` (isn't generic, doesn't support SignalArgs-derived types, etc).

I haven't been keeping up with our signal API that much, so I wanted to double check whether this was intentional (and by extension, which version I want to be using).

badcel:
@firox263 great! The generic version should be the correct one.

I just got GTK compiling (but without any custom code) :rocket: 

If you have some time left you could check if your code still works with the generation of events for GTK and merge it into this if it is working. It looks good to me. If you have some more time left you could even start adding back custom code to the GTK project :fireworks:

badcel:
I got the quickstart sample working again :rocket:

PR #92, opened 2020-10-28
Restore gstreamer functionality

badcel:


na2axl (commented):
Hi @badcel, the code side looks good to me. But I have some problem to execute the sample, there is the stacktrace:

```
(Gstreamer.exe:18348): GLib-GObject-CRITICAL **: 21:53:58.666: g_object_weak_ref: assertion 'G_IS_OBJECT (object)' failed

(Gstreamer.exe:18348): GLib-GObject-CRITICAL **: 21:53:58.667: g_object_ref_sink: assertion 'G_IS_OBJECT (object)' failed

(Gstreamer.exe:18348): GStreamer-CRITICAL **: 21:53:58.668: gst_element_set_state: assertion 'GST_IS_ELEMENT (element)' failed

(Gstreamer.exe:18348): GStreamer-CRITICAL **: 21:53:58.669: gst_element_get_bus: assertion 'GST_IS_ELEMENT (element)' failed
Unhandled exception. System.InvalidCastException: Unable to cast object of type 'Gst.Element' to type 'Gst.Bus'.
   at GObject.Object.GetObject[T](IntPtr handle, T& obj) in D:\Aliens_Group\Gir.Core\gir.core\Libs\GObject\Classes\Object.cs:line 247
   at Gst.Element.GetBus() in D:\Aliens_Group\Gir.Core\gir.core\Libs\Gst\Classes\Element.cs:line 11
   at Sample.Gst.Play() in D:\Aliens_Group\Gir.Core\gir.core\Samples\GStreamer\Play.cs:line 12
   at Samples.Program.Main(String[] args) in D:\Aliens_Group\Gir.Core\gir.core\Samples\GStreamer\Program.cs:line 7
```

I try to run it on Windows, maybe it is an OS compatibility problem ?

firox263:
> @na2axl I try to run it on Windows, maybe it is an OS compatibility problem ?

It's working for me (Windows 10 Home, 64-bit), so I think it's Gst related. How are you getting GStreamer? I've got mine from MSYS2 using the `mingw-w64-x86_64-gstreamer`, `mingw-w64-x86_64-gst-plugins-bad`, and `mingw-w64-x86_64-gst-plugins-good` packages.

According to the GStreamer docs for [gst_parse_launch](https://gstreamer.freedesktop.org/documentation/gstreamer/gstparse.html#gst_parse_launch), the return value is either a `GstElement` or `NULL` depending on the result of the function. What this tells me is that for some reason (maybe missing gst plugins?), `playbin` has failed to launch for you, and since we're not handling this, it tries to continue and crashes (hence the GLib assertions).

It should be a simple fix, but we should look at more descriptive error messages.

I have absolutely no idea why it's trying to convert between `Gst.Element` and `Gst.Bus` though. I've found a few nasty bugs with the type dictionary yesterday, so I suspect it's related. Will open an issue soon.

Generator/Templates/GObject/class.sbntxt
```diff
@@ -27,7 +21,7 @@ namespace {{ namespace }}
 {
     // AUTOGENERATED FILE - DO NOT MODIFY
     
-    public{{get_abstract}}partial class {{ name }} {{ (': ' + parent) | get_if_parent }}
+    public partial class {{ name }} {{ (': ' + parent) | get_if_parent }}
```
firox263, line 24: What's the reasoning behind removing abstract classes? Not that I'm against it, but I'm trying to remember what we used them for.

Libs/Gst/Classes/Parse.cs
```diff
@@ -1,16 +1,16 @@
-using Gst;
+using System;
 
 namespace Gst
 {
     public class Parse
     {
         public static Element Launch(string pipelineDescription)
```
firox263, line 7: We need to handle the case where `ret` is NULL, and return appropriately. Should probably be `Element?`.

Samples/GStreamer/Play.cs
```diff
@@ -7,11 +7,11 @@ public class Gst
         public static void Play()
         {
             Application.Init();
-            var ret = Parse.Launch("playbin uri=playbin uri=http://ftp.halifax.rwth-aachen.de/blender/demo/movies/ToS/tears_of_steel_720p.mov");
+            Element ret = Parse.Launch("playbin uri=playbin uri=http://ftp.halifax.rwth-aachen.de/blender/demo/movies/ToS/tears_of_steel_720p.mov");
```
firox263, line 10: See above comment. We need to ensure that `Element` is not null before using it.

Samples/GStreamer/Program.cs
```diff
@@ -1,10 +1,10 @@
...
-        }
-    }
-}
+﻿namespace Samples
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Sample.Gst.Play();
```
firox263, line 7: This all looks good, but I've noticed `Sample.Gst.Play()` takes a while to startup for me. Could we get a `Console.WriteLine` here just to indicate that the program has started?

firox263 (changes requested):
Looks good to me for the most part. Needs a few changes with `Gst.Launch` failing when `playbin` can't be run.

badcel:
@na2axl @firox263 Thanks for your reviews. I'm going to respond later this day.

@na2axl Perhaps you keep your setup so we can verify if the coming bug fixes create better error messages?

na2axl:
Thanks for clarification @firox263, you are right I am missing the good/bad plugins from gst. @badcel I will keep my setup at is to test error messages.

Generator/Templates/GObject/class.sbntxt
badcel: I think we did not use it at all. We inserted it "just because". The problem is that we can receive objects of abstract classes from GTK.

In this PR there is the Gst.Parse class which returns an `Element`. In the gir file this class is marked as abstract:

```XML
<class name="Element" c:type="GstElement" parent="Object" abstract="1"
...
```

To be able to create an instance of this object our class can't be abstract.

The global function `parse_launch` returns this `<type name="Element" c:type="GstElement*"/>`. So our best bet is to create an `Element` even if it is abstract.

Samples/GStreamer/Program.cs
badcel: I put the `Console.WriteLine` into the `Play` method, like in the DBus samples.

Libs/Gst/Classes/Parse.cs
badcel: As `Global.parse_launch` returns an error pointer I'm now throwing an Exception in case of an error. Otherwise we would lose the error information.

Samples/GStreamer/Play.cs
badcel: Not needed as `Parse.Launch` now throws an exception which would lead to a program crash if thrown. But in case of an exception the users receives a detailed message what went wrong.

firox263: This is fine for me, but w.r.t what I described in #93 about GError. If I intentionally break this (e.g. rename `playbin` to `plybin`), I get the following error message:

```
PS D:\gir.core\Samples\GStreamer> dotnet run
Starting to play tears of steal. Please wait while file is beeing loaded...
Unhandled exception. GLib.GException: ?????????"?
   at Gst.Parse.Launch(String pipelineDescription) in D:\gir.core\Libs\Gst\Classes\Parse.cs:line 22
   at Sample.Gst.Play() in D:\gir.core\Samples\GStreamer\Play.cs:line 13
   at Samples.Program.Main(String[] args) in D:\gir.core\Samples\GStreamer\Program.cs:line 7
```

The exception message is garbled because of the way we're marshalling it. This problem is unrelated to this PR, but means we definitely need to look into `Error` and `PtrToStringAnsi` vs `PtrToStringAuto`.

firox263 (approved):
This looks good to me. @na2axl could you test this again (+maybe with the change to `Error.cs`) to see if this works on your setup.

na2axl (commented):
Hi @badcel, I've seen that you have marked some features as TODO, to move it on the Toolkit layer later. AFAIK I don't think that theses shortcuts have to be implemented in the Toolkit layer, because this will mean that we will have another class named `Bus` in the Toolkit layer which will wrap the `Bus` class of the Core layer (because don't forget that the Toolkit layer will cover/wrap the Core layer), and this sounds like a bad design for me...

From my POV I prefer that theses methods are in the Core layer, since they are directly related to the GStreamer library, and also because the Core layer want to be a C# friendly wrapper of GNOME libraries, so we can reduce to a bare minimum the usage of native functions (through shortcuts like theses ones) without disallowing the end user to access them.

na2axl:
I've checked the code again without changes in my setup, this is what I have:

```
PS D:\Aliens_Group\Gir.Core\gir.core\Samples\GStreamer> dotnet run
Starting to play tears of steal. Please wait while file is beeing loaded...
Unhandled exception. GLib.GException: no element "playbin"
   at Gst.Parse.Launch(String pipelineDescription) in D:\Aliens_Group\Gir.Core\gir.core\Libs\Gst\Classes\Parse.cs:line 22
   at Sample.Gst.Play() in D:\Aliens_Group\Gir.Core\gir.core\Samples\GStreamer\Play.cs:line 13
   at Samples.Program.Main(String[] args) in D:\Aliens_Group\Gir.Core\gir.core\Samples\GStreamer\Program.cs:line 7
```

After installing good/bad GST plugins everything works as expected 🚀 Nice work !

PR #93, opened 2020-10-29
Generate Functions + Other Tweaks

firox263:
Draft for now as needs to be rebased on develop after #92.

This PR:
 * Moves function generation code from `GRecord` to `GClass` in order to generate static methods (C functions) for classes like `GdkScreen`.
 * Fixes a bug where functions were only generated when 'moveTo' was *not* null (whereas we want to generate functions when it *is* null - i.e. functions that haven't been moved).
 * Re-adds `WrapPointerAs()` as a wrapper around `TryWrapPointerAs<T>()`.
 * Makes GObject's `Handle` public, as we need to access it from outside the class in order to implement functions like `Box.PackStart(widget, ...)`, where it needs the handle of the widget passed to it.


badcel:
> @firox263 Makes GObject's Handle public, as we need to access it from outside the class in order to implement functions like Box.PackStart(widget, ...), where it needs the handle of the widget passed to it.

Can't we use `protected static IntPtr GetHandle(Object obj)`? I did not check the code but with this already available function every GObject can get Handles from each other GObject. If we can use this we don't need to make `Handle` public. It is already used in some places, too.

firox263:
> @badcel: Can't we use `protected static IntPtr GetHandle(Object obj)`? I did not check the code but with this already available function every GObject can get Handles from each other GObject. If we can use this we don't need to make `Handle` public. It is already used in some places, too.

Ah, my bad. I think I wrote some of this code before we added GetHandle (or maybe I just didn't notice 😅). This is definitely a much better way of doing things - will update now.

Libs/GObject/Classes/Object.cs
```diff
@@ -26,7 +26,7 @@ public partial class Object : INotifyPropertyChanged, IDisposable
 
         #region Properties
 
-        protected internal IntPtr Handle { get; private set; }
+        public IntPtr Handle { get; private set; }
```
badcel, line 29: Use GetHandle instead of making Handle public

Libs/GObject/Classes/Object.cs
```diff
@@ -255,6 +255,14 @@ protected internal void OnPropertyChanged([CallerMemberName] string? propertyNam
         // This function returns the proxy object to the provided handle
         // if it already exists, otherwise creats a new wrapper object
         // and returns it.
+        public static T WrapPointerAs<T>(IntPtr handle)
+        {
+            if (TryWrapPointerAs<T>(handle, out T obj))
+                return obj;
+
+            throw new Exception("Could not wrap handle");
```
badcel, line 263: Include typeof(T) in exception message?

Libs/GLib/Structs/Error.cs
```diff
@@ -7,14 +7,14 @@ public partial struct Error
     {
         #region Properties
 
-        public string Message => Marshal.PtrToStringAuto(message);
+        public string Message => Marshal.PtrToStringAnsi(message);
```
badcel, line 10: Why is Ansi better compared to Auto?

badcel (commented):
Looks good for me, great that you found the "moveto" bug, it's quite a huge mistake.

I left some questions, too.

Libs/GObject/Classes/Object.cs
firox263: Fixed

Libs/GObject/Classes/Object.cs
firox263: Fixed

Libs/GLib/Structs/Error.cs
firox263: This was a really bizarre issue I found while playing around with CssProvider. Take
the following method as an example (implementation not important):

```cs
// This function loads an embedded stylesheet and creates a CSS
// provider from it.
private static CssProvider CreateGlobalProvider(Assembly asm)
{
    var provider = new CssProvider();

    var cssStr = GetStylesheetFromAssembly(asm);
    provider.LoadFromData(cssStr, out GError error);

    if (error != null)
        throw new Exception(error?.Message); // <-- THIS IS THE RELEVANT PART

    return provider;
}
```

When we marshal `message` as either Auto or Uni, and throw an exception like in the snippet above:

```
Unhandled Exception: System.Exception: ?????????????????????
```

While if we marshal as Ansi:

```
Unhandled Exception: System.Exception: <data>:2:16expected '}' after declarations
```

I have no clue why this happens, but PtrToStringAnsi has been working consistently for me. I assumed it was because GError was simply encoded in ANSI, but in hindsight this doesn't make any sense, since Unicode should work too. Any ideas why this could be?

badcel: Interesting, I just did a test with fedora and for me both methods work. So I think we should use the "Ansi" method.

On another sidenote: I saw your work for heyena and I think it is great :+1: . There is a lot to do for gir.core and with the samples running we can check how to improve the generation process. But with hyena we have some real world project which we can bring to live. For now I will focus on gir.core as there is more or less nothing working here :wink:. But I think we can profit from your early work and tests with hyena as you will discover real world problems using gir.core. If you have some difficulties or need support just ask I'm willing to help :rocket:

badcel (approved):
Looks good to me.

firox263:
@badcel For `struct.sbntxt` do we want `Native` to be a `struct` or a `class`? I'm not sure if it makes a difference since all of the methods inside are static.

For reference:
image

badcel:
@firox263 : I made a quick test with this code: https://gist.github.com/badcel/53391d0bc37224c692c2c0ed4b79274f

It looks like both structs have the same size. Even if we add fields to the class / struct it does not change which makes kind of sense because the struct is just a container for data. The `Native` class or structure is just another type, which is nested inside the enclosing type. As long as the enclosing type does not have any field of this data type there is no data to transfer, because the type hierarchy is not encoded in the data at all.

So we can probably choose which type to use. I would prefer the class as it is inline with the generated classes and does not allow to generate regular fields as it should only connect us with the native world via methods.

PR #95, opened 2020-10-31
GtkMinimal: Reenable sample

badcel:
Still work in progress

### DIFF
--- a/Samples/Gtk3/GtkMinimal/DemoWindow.cs
+++ b/Samples/Gtk3/GtkMinimal/DemoWindow.cs
@@ -16,13 +16,13 @@ namespace GtkDemo
         public DemoWindow(Application application) : base(application, "demo_window.glade")
         {
             // Connect Button
-            Button.Clicked += Button_Clicked;
+            Button.OnClicked += Button_Clicked;
 
             // Create Notebook
             notebook = new Notebook();
             Box.PackStart(notebook, true, true, 0);
 
-            var image = new FileImage("data/gtk.png");
+            var image = Image.FromFile("data/gtk.png");
             notebook.InsertPage("Image", image, 0);
 
             var label = new Label("Gtk and C# - Very exciting isn't it?");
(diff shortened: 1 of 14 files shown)


firox263:
I know this is still WIP, but a heads-up about some of the constructors:

```cs
public Application(string applicationId, ApplicationFlags flags = ApplicationFlags.FlagsNone) 
    : this(Native.@new(applicationId, flags)) { }
```

From what I recall, the `this(Native.@new)` pattern breaks when subclassing, as we need to be going through either `Object()` or `Object(ConstructParameter[])` for the type integration code to run. The `IntPtr` constructor would simply wrap the `GtkApplication` as a foreign-created object.

badcel:
The sample is working again. I think we can start with this simple Builder support, as it is not infiltrating our inheritance chain like the original concept. I hope to see the template support in the future.

Samples/Gtk3/GtkMinimal/DemoWindow.cs
```diff
@@ -13,37 +13,41 @@ public class DemoWindow : ApplicationWindow
...
         {
+            Application = application;
+            builder.Connect(this);
+            
             // Connect Button
-            Button.Clicked += Button_Clicked;
+            Button.OnClicked += Button_Clicked;
 
             // Create Notebook
             notebook = new Notebook();
```
na2axl, line 26: @badcel can we use the object initializer (with signal indexers) here ?

na2axl (approved):
LGTM, only the little enhancement I think we can add in the DemoWindow.cs. 👍🏾

Samples/Gtk3/GtkMinimal/DemoWindow.cs
badcel: Hm, I like to see that the "classic" style is possible, too. We should probably show the possiblities which are available.

I don't like how the samples are named, but this is for another PR.

PR #96, opened 2020-11-01
API Coverage: Add Various Methods + Properties

firox263:
This PR adds various methods, properties, etc to Gtk3 and Gdk3.
 - This is enough to port over ExceptionDialog.cs from Hyena.
 - Additional minor changes to WINDOWS.md.

Currently, we don't support defining properties on interfaces. This is needed for things like `GtkOrientable` (see `Orientable.cs`) which is used in boxes. I've temporarily defined it on GtkBox so box is usable, but we should get rid of this sooner rather than later.


firox263:
Needs to be rebased on `develop` after #94

na2axl:
Hi @firox263, nice work with this, maybe we can continue here to add every other missing properties/method ? If yes I will be able to help 🙂.

For the GInterface problem... I think the only solution to make them support our property model is to convert them to **abstract classes** in C#. I think with abstract class we can achieve the same behavior of a GInterface. What do you think ?

badcel:
@na2axl I think we can't use abstract class for GInterfaces, as a class can implement several interfaces, but not several abstract classes.

I made a test for interfaces with default implementations here: https://github.com/gircore/gir.core/pull/99

firox263:
@na2axl Resolved the conflicts, should be good to merge now. I'll leave it to you to do some stylistic cleanup/reorganisation.

PR #99, opened 2020-11-01
Feature/support interfaces

badcel:
This is just a quick try. It uses default interface implementations to inject code into classes. The big drawback is, that the interface is not visible on the class. The object must be cast to the interface to access the value.

I think we don't want to follow this aproach or? Do you have any ideas @na2axl @firox263 ?

Edit: On the other hand: I want to introduce interfaces for all types. E.g. if there is an IBox interface which implements the Orientable interface than there is no need to cast the object because all instances are just used as an interface. (I did not test this, it is my assumption as I did not use default interface implementations until now).

Alternatively we could try to redirect properties from the class to the interface, meaning we would generate less code because we do not need the property descriptor for all implementations.


badcel:
I added a converter method and an interface. It works like I thought. The big bonus on this implementation is that it solves the diamond problem for us and we have very little code to generate. As programming against interfaces is a much better way than using classes this would be an acceptable solution for me.

How to access the interface properties:

```C#
Box myBox = new Box(Orientation.Horizontal, 5);

if (myBox is Orientable ot)
{
    var bla = ot.Orientation;
}

var blu = myBox.AsOrientable().Orientation;

IBox myIBox = new Box(Orientation.Vertical, 5);
var fu = myIBox.Orientation;
```

The next question would be if this plays nice with out type integration?

na2axl:
@badcel this is very interesting, I think with your implementation we (and also end users) can achieve other high level things like IoC, Dependency Injection, even better MVVM (maybe with third party library like MVVMCross) and more (things that are definitely not possible with GtkSharp) !

I will try to test this in deep. Really nice work 👏🏾

na2axl (approved):

firox263:
> As programming against interfaces is a much better way than using classes this would be an acceptable solution for me.

I think this is the most elegant solution so far. I'm assuming the plan is to define `OrientationProperty` as a static property descriptor in the interface itself (can we do that? I think with C# 8 we can). The question is, do we add `Orientation` exclusively in the interface or in the type too?

Because if we define it in the type, we still get the benefit of doing the following:

```cs
// Can access Orientation from both Box...
var box = new Box(Orientation.Horizontal);
box.Orientation = Orientation.Vertical;

// And also IOrientable:
var orientable = box.AsOrientable();
orientable.Orientation = Orientation.Horizontal;
```

The downside is the static descriptor is only available from the interface Orientable, although that would happen regardless. Thoughts?

> The next question would be if this plays nice with our type integration?

What would be the problem here? I can't think of anything of the top of my head.

---

A few other unrelated thoughts:
 - I've noticed we don't prefix interfaces as `IOrientable`. Maybe we should do this, to be consistent with C# convention.
 - Some food for thought: The C# team are going to implement static interface members/type classes (Called "Roles and Shapes") similar to the way GObject works. When this is implemented, it could greatly simplify the bindings and would fix just about all of our pain points. Unfortunately, this probably won't be a thing until C# 11.0 (2022). Possible opportunity down the line.

na2axl:
>  The question is, do we add Orientation exclusively in the interface or in the type too?

Yes @firox263 I am thinking about that, maybe we can generate proxies in the concrete type? But this can also end in the diamond problem 🤔 , or it will just make us write twice the same code.

> The downside is the static descriptor is only available from the interface Orientable, although that would happen regardless. Thoughts?

I think this doesn't have a noticeable difference, since the static property descriptors always remain in the type where the properties they describe are... So even if the user don't want to cast every time his types, he will always be able to do:

```cs
Box box = new Box(Orientation.Vertical, 10);

// Somewhere when the time comes to change the orientation
IOrientable.OrientationProperty.Set(box, Orientation.Horizontal);
```

> What would be the problem here? I can't think of anything of the top of my head.

😂 me too.

> I've noticed we don't prefix interfaces as IOrientable. Maybe we should do this, to be consistent with C# convention.

Yes I just seen that, I think we must follow C# convention also for interfaces (we are already removing the **G** prefix on classes, so adding the **I** prefix on interfaces is not a problem 🙂)

na2axl:
@badcel, @firox263 I'm thinking that if we want to continue in this way we will have to move all properties and also signals and indexers in interfaces, and oblige end users to work with interfaces (this is not really a problem), is it true? I think with that we can avoid our class name - property name conflicts problem:

```cs
public interface ILabel : IObject
{
    public static readonly Property<string> LabelProperty = Property<string>.Register<ILabel>(
        "label",
        nameof(Label),
        (o) => o.Label,
        (o, v) => o.Label = v
    );

    public string Label
    {
        get => GetProperty(LabelProperty);
        set => SetProperty(LabelProperty, value);
    }
}

public partial class Label : ILabel { }

ILabel label = new Label("Gir.Core is awesome");
label.Label = "Gir.Core is REALLY awesome";
```

What do you think ?

firox263:
@na2axl I'm a bit confused with how this would work. What exactly are the benefits of moving properties to interfaces? Also, would we generate an interface for every type to contain the properties?


Since it's an interface, does that mean we could still do the following, using the `Label` type rather than `ILabel`:

```cs
public class Label : ILabel { ... }

// Using Label directly
Label label = new Label("Some text");
label.Text = "Change the text";
label.XAlign = 1.0f;
```

This is one of the strong points of our current design, where it matches the GObject type system very nicely (e.g. GObject=object, GInterface=interface, etc). I'm not too enthused about this from a API point of view, but if it has clear benefits, I'm open to it.

I think I still prefer @badcel's `AsOrientable()` method which simply returns the `Orientable` interface. Another messier solution, but one that "feels" better for the end user, could be to simply define the property twice for both the class and the interface. If we have some dictionary or list that stores property descriptors (we might already?), we could share the one descriptor across all the places where it's declared. Since we're generating, duplication wouldn't be as much of an issue.

badcel:
**Type System integration**
Regarding type system integration I was thinking about the `get_type` method. Interfaces in `GObject` have this method and I don't know if this can get us into some trouble? If we have interfaces with a type descriptor for the type? What if we get an `IntPtr` of an interface? how do we create an object of it? Can we? Can we receive such a pointer?

-----
**Interface naming**
Actually we never did change the names of high level data structures. We did change method names / properties, but never the structures itself. We are taking the names the gir file provides. The gir itself is making the split between the namespace and the class, struct, enum etc. 

This could get very hard if there are references from other libraries to e.g. gtk. There would be s.th. like "Gtk.Orientable" as a type string inside the gir. If we change the names and want to generate the bindings for this library we would need to know that this type is an interface, just from the name itself. As there is no clue for this, we would need s.th. like a global type registry during generation. If we know generate the 3rd party bindings, we would need to load the Gtk gir, check if this type is an interface and remember that we would need to prefix it if we generate the 3rd party bindings.

This is quite a lot of work. The prefixing with the "G" was there as an idea but it was not really working out, because of this huge mess it would make. At least right now I don't want to do all this work. But we need to decide how to handle interfaces names before we make our first release that's for sure.

I would welcome if we rewrite the whole generator as a team, with a clear concept as it currently quite hacky. But this could be combined with the interface prefixing, if we decide to do it. For me this is a topic if we are happy with all the rest of the generated code. And there is much more to think about before I would takle this one.

------------
**Expand the concept of interfaces with default implementations**
Regarding the interfaces I agree that its quite elegant. With the default interface implementations we are like injecting code. Like I mentioned before I would like to see interfaces for all types, to allow better unit testing. So if we grab @na2axl thought, we could generate all properties of an class in an interface including a default implementation and we would probably solve the problem automatically.

If we do id like this i would create factory classes like:

```C#
public class Button
{
  public static IButton Create(string label) => new Button(label);
  private Button(string label)
  {
....
  }
}

```

In this way the user would not have the problem to chosse which type to use (class vs. interface), because the only way to create a Gtk.Button is to use the factory method of the Gtk.Button which actually returns an IButton. I like this very much but it has one drawback which is that we loose the property initialization syntax as this creates a compiler error.

```C#
var myButton = Button.Create("myFirstLabel")
{
  Label = "mySecondLabel"
};
```
An alternative would be:

```C#
public class Button
{
  public static IButton Create(Action<Button> initalizer)
  {
    var btn = new Button();
    initializer(btn);
    return btn;
  }
}

var myButton = Button.Create(b =>
{
    b.Label = "myLabel";
    b.UseUnderline = true;
    b.OnClicked += OnClicked;
});
```
With this we could probably replace our object initializer syntax but it is by far not as nice.

Effectively classes would become factories and interfaces would contain the actual code. I don't know if it is a good idea to use default interface implementations in such an extensive way. Currently I would agree to @firox263 that we should keep classes as classes and interfaces as interfaces. With the current .NET interface default implementations we have an easy way to automagically implement the interface and keep things more in line with each other.

**Wrapping it up**
As I see it know I would probably go the route of default interface implementations. For now we could keep the `AsOrientable` method as it is easy to generate.

In a later step I would suggest that we create the property but using the PropertyDescriptor of the interface:

```C#
public partial class Box
{
  public Orientation Orientation
  {
    get => GetProperty(Orientable.OrientationProperty);
    set => SetProperty(Orientable.OrientationProperty, value);
  }
```

This would require some more work on the generator, too, because we have to load data which we are currently not generating.

If we later need to create an object for an interface but we don't know the actual type, we could probably just create a class and directly return it casted as an interface:

```C#
public class OrientableImpl : Object, Orientable
{
  internal OrientableImpl(IntPtr ptr) : base(ptr) { }
}
```
Do you guys think we can start with this?

na2axl:
> Regarding type system integration I was thinking about the get_type method. Interfaces in GObject have this method and I don't know if this can get us into some trouble? If we have interfaces with a type descriptor for the type? What if we get an IntPtr of an interface? how do we create an object of it? Can we? Can we receive such a pointer?

I think a solution could be to have a proxy class which will be used to create C# object from GObject interfaces with an IntPtr, maybe something like:

```cs
// The interface
public interface Orientable : IObject {}
// The class
internal class OrientableImpl: Object, Orientable
{
  public OrientableImpl(IntPtr ptr) : base(ptr) { }
}
```

So in the type system is the `OrientableImpl` which is registered for the `GType` of the `Orientable` interface. Is it possible ?

> This could get very hard if there are references from other libraries to e.g. gtk. There would be s.th. like "Gtk.Orientable" as a type string inside the gir. If we change the names and want to generate the bindings for this library we would need to know that this type is an interface, just from the name itself. As there is no clue for this, we would need s.th. like a global type registry during generation. If we know generate the 3rd party bindings, we would need to load the Gtk gir, check if this type is an interface and remember that we would need to prefix it if we generate the 3rd party bindings.

Never minded that this will require a lot of work 😅. Personally my thoughts was to only add an **I** prefix in the `interface.sbntxt` template file 😂

> Regarding the interfaces I agree that its quite elegant. With the default interface implementations we are like injecting code. Like I mentioned before I would like to see interfaces for all types, to allow better unit testing. So if we grab @na2axl thought, we could generate all properties of an class in an interface including a default implementation and we would probably solve the problem automatically.

@badcel we can ignore my previous proposition, just too much enthused 😅. After some tests I have seen some drawbacks of using interfaces to store default implementations:

- The awesome MVU-like style we use now to initialize a whole widget tree will be dropped... But this initialization style is a major argument to make users comes here.
- Like you said, we will need to use factories and not constructors, this can break several things, and even discourage ones which want to move from GtkSharp to Gir.Core.
- Our API may be less plugable... For example we will be unable to implement a XAML-based UI framework, because the XAML language work with classes, not interfaces.
- The end user can be too confused to switch between classes and interfaces.

I think for that I agree with @firox263 to go with what you have already implemented, and generate the `AsInterface()` methods. 

But... Another solution could be also to implement GInterfaces in GObjects ourselves, so it's in the `Box` class that the `Orientation` property and it's descriptor remain, and in the `Orientable` interface we have only the requirement of an `Orientation` property.

```cs
public interface Orientable: IObject
{
  Orientation Orientation { get; set; }
}

public interface IBox: Orientable { }

public partial class Box: Object, IBox
{
  public static Property<Orientation> OrientationProperty = ...;

  public Orientation Orientation { get => ...; set => ...; }
}
```

I think with that:

- We don't loose our MVU-style widget initialization
- We always have the possibility to do IoC, dependency injection, mocking in unit tests, etc because we always generate interface of each objects (ILabel, IWidget, IWindow, etc...)
- We avoid some headaches with the generator and the API design...

Finally I'm much more favorable with this design because it's simple and do the thing we want.

badcel:
@na2axl : Actually I think you are right. The right way is to do it the old fashined way. With default interface implementations it is just easier to generate currently, because we can put the code where it is defined (inside the interface). So probably we need to improve the generator first, before supporting interfaces. As a start we could just create the property descriptors inside the interface, which actually makes sense. And start hand writing the interface implementations which are important for now (e.g. get Hyena to work).

badcel:
> As a start we could just create the property descriptors inside the interface, which actually makes sense.

I'm not sure of this anymore. If we don't inject the properties, why should we inject the descriptors as they won't be visible, too. So I just removed default interface imlementations for now. The rest of the changes can be applied.

In this way we are generating empty interfaces, and have support for enums and some constructor improvements. This is at least s.th. from this PR.

PR #102, opened 2020-11-04
Feature/type integration

badcel:
Still work in progress

Documentation
- https://developer.gnome.org/gobject/stable/pt02.html
- https://developer.gnome.org/gobject/stable/chapter-gobject.html
- https://gi.readthedocs.io/en/latest/annotations/giannotations.html

Class struct contains pointer to overwrite virtual functions


badcel:
I added the code to load template without reflection on the GTK side. I think the code is quite clean. Yes it is boilerplate, but it needs no tricks to initialize the template.

The callback of the button click is currently missing and need's to be implemented.

**What would be needed to do it with reflection and attributes?**
If the code could not reside inside the `CompositeWidget` the type initialization must be done inside the `Widget` class. For this to work there needs to be some trigger after object initialization to call `InitTemplate` inside `Widget` for the `CompositeWidget` class under certain conditions (child class must use a template). Furthermore the classinit of the `CompositeWidget` must be used to call `SetTemplate` in the `Widget` class automatically.

badcel:
@firox263 + @na2axl : Can you take a look at Button.cs: https://github.com/gircore/gir.core/blob/461b7aeba68047433b5842d6b8c654bd19ee21a2/Libs/Gtk3/Classes/Button.cs#L96

I wonder if we could create s.th. like a descriptor for virtual methods to make generation easier. The code in Button shows how we could allow the user to override a method. In general it works like this:

- In the class constructor of our native types (not subclasses) we save the pointer of the original virtual function and replace it with a pointer to our own `private static` method which has the same signature like the original function.
- We create a new virtual method (non static) which just calls the original code.
- If we receive a function call from the C world, we just forward the call to our own virtual method, if we know the object. In case the caller did not call our implementation (via `base.Pressed();`) the original code is not executed. If we don't know the object it must be some other object which is not ours and we just call the standard implementaion.

_Possible optimization: Probably we have to check if the instance is a subclass. In case it is not a subclass we always call the original implementation._

Do you have any ideas how it could be more straight forward to generate the code without much duplication? Or perhaps you have other ideas to achive the same goal?

na2axl:
Hi @badcel, I've just checked the code, seems very promising! I'm ok with the process, now I think about the generation of this code using descriptors like properties/signal are currently done.

I'm currently sure that generate this code like what is done in the Button class is possible, maybe for that the generator will need a kind of dictionary to remember what variable is storing what default/original method, or something else, and then generate the bunch of `public virtual` methods which will all just make a call to the native default/original function through the variable... 

But...

From my side I think it will also be possible to introduce a new descriptor for virtual methods (said, method descriptors). A benefit with that (added to the possibility to generate code an easier code) is just we can allow GTK developers to really create widgets like they do in C, because in C we can't create a _subclass_ of a `GtkButton`, so we have to replace the `pressed` method with another one, in C# we can achieve a similar behavior with our lovely **indexers**. An example can be:

```cs
new Button
{
  [Button.PressedMethod] = (Button b) => Console.WriteLine("The button has just be pressed!"),
}
```

For now this feature is just an idea resulting of what we are doing now. To really achieve this we have to be sure that a virtual method can be overridden for an already created object (I really don't know if its possible).

> Possible optimization: Probably we have to check if the instance is a subclass. In case it is not a subclass we always call the original implementation.

I don't think if this is the good behavior. If I understood well, for a subclass named `FakeButton` for example, and implemented like this:

```cs
public class FakeButton : Button {}
```

In this case if the `pressed` function get called is it normal to not call the _original_ function ?

badcel:
If the method is not overridden we call the default implementation of our wrapper which calls the original code: meaning the fakebutton will work as expected.

I would like to see some descriptor. But there is one problem: the c world expects a static method to call. For me it is okay to use a descriptor to override but I would like to see normal overriding behaviour, too (which is working with my sample code).

I wonder if we can achieve both things?

firox263:
> For now this feature is just an idea resulting of what we are doing now. To really achieve this we have to be sure that a virtual method can be overridden for an already created object (I really don't know if its possible).

GObject's "virtual methods" are effectively just function pointers in the object's Class struct. This means that in order to override a given virtual method, you have to change the Class struct, which changes *all* instances of a given type. So from a technical standpoint, this isn't really possible AFAIK.

> In the class constructor of our native types (not subclasses) we save the pointer of the original virtual function and replace it with a pointer to our own private static method which has the same signature like the original function.

I wonder what the overhead is for routing through the CLR each time? It's probably negligible, but perhaps using reflection to selectively override the class could end up *more* performant. Either way, the static function + virtual method approach sounds great to me, and we get to keep native C# syntax.

badcel:
Yeah you are right. Routing every event through dotnet is not very clever. If we use generators we could try to save the original pointer during class initialization. But we only override the PTR in the subclass class struct if there is some override method in our subclass.

Libs/Gtk3/Classes/Widget.cs
```diff
@@ -39,6 +40,57 @@ protected static void BindTemplateChild2(Type gtype, string name)
...
+
+            if (targetEvent is null)
+                return;
+
+            System.Type? targetEventEventHandlerType = targetEvent.EventHandlerType;
+
+            if (targetEventEventHandlerType is null)
+                return;
+            
+            var del = Delegate.CreateDelegate(targetEventEventHandlerType, signalsender, callbackMethodInfo);
```
badcel, line 89: @firox263 + @na2axl : Here it complains that the signatures are not compatible. Any idea why? The method should connect the "button_clicked" method of the `CompositeWidget.cs` file in the "Template" project with the template.

It looks like I'm missing s.th. :see_no_evil:

badcel:
It is fixed, callbacks are working now :rocket: 

I implemented it via reflection for now.

badcel:
Superseded by #182 .

Issue #66, opened 2020-08-29
Typedict expensive lookup


badcel:
A thought:

We should probably check if the GType of generic type 'T' is equal to the gtype from the pointer. That way, we can avoid the expensive lookup in the type dictionary for cases where we know the type we want (e.g. `Dialog.GetContentArea`).

_Originally posted by @firox263 in https://github.com/gircore/gir.core/pull/65#issuecomment-683268065_

-------

I created an issue so we don't forget about this comment. I currently don't see the expensive lookup. Perhaps you could just create a pull request?

firox263:
Effectively, we'd retrieve the GType for the generic <T> and compare it with our known GType, adding it to the type dictionary in the process. That way if <T> is the true type of our ptr, we avoid an expensive type dictionary lookup, and if it isn't, we still register it in the type dict for future use.

For this we'll need a way of comparing GTypes (implementing `Equals()` and overriding `==/!=`), but I can't do it cleanly with our current implementation of GObject.Type (which wraps `Sys.Type`). I think this ties into the point @na2axl brought up about how we have `Sys.Type`/`Sys.Value`/etc in `Wrapper` instead of `Core`. If it's possible, I think we should move the contents of `Wrapper/API` into the core part and put the struct definitions in `.Abi.cs` partial classes like with `Object.Abi.cs`.

This would also have the benefit of keeping all library logic in `Core` and all wrapper functions in `Wrapper`.

PR #105, opened 2020-11-11
Type Dictionary Improvements

### LINKED ISSUE #97: Type Dictionary Tracking Issue
Found a really nasty bug with the type dictionary in our lookup code. If we don't instantiate an object before using it in `WrapPointerAs<T>`, it instead falls back (as intended) to the 'lowest common denominator' type. This is needed for Gdk objects, as a `GdkScreen` is actually a `GdkWin32Screen` (but of course, that's an internal type which is not exposed to bindings), so we default to `GdkScreen` instead.

The problem however, is we end up falling back to the last *registered* type, which in most cases ends up being either `Gtk.Widget` or a plain `GObject`. This always causes an Exception when used with `WrapPointerAs<T>` since you can't cast a true `GObject` type up to a `Gdk.Screen` for example.

Since this basically involves rewriting half the typedict, I've decided to do it all in one go and fix some of the other issues we had in the process. I'm currently working on this at [firox263->feature/typedict](https://github.com/firox263/gir.core/tree/feature/typedict).

**TODO:**
- [X] Rewrite lookup code to search registered *and non-registered* types.
- [X] ~Handle Unloaded Assemblies (e.g. Cairo, Pango, etc)~ Anti-Feature -> **No Longer Considered**
- [x] Find a better way of obtaining the Namespace vs the Type Name from GObject (Currently it just takes the first "word" so `GtkTextView` -> `Gtk` + `TextView`. This is problematic with things like `GtkSourceView` where it doesn't fall under the `Gtk` assembly but does under the namespace).
- [x] Do a check with TypeDescriptors to be sure the objects are derived.
- [x] More robust handling of full type names. `Assembly.GetType(string)` needs the type name in the form `Namespace.Type`, so maybe each assembly should define some kind of metadata with the namespace it implements.
- [x] Register recursively up the entire tree where possible, as this is much cheaper than doing it later.
- [x] Optimise this with a few "fast paths" to avoid doing expensive lookups. This fixes #66 
- [x] Investigate more optimisations

firox263:
**19/11/2020:** This PR is effectively complete now. In subsequent PRs, we should look at transitioning away from a runtime type dictionary to a static Xamarin-style typemap.

Previous Message Below:

~**See #97 for a todo list of what's remaining. Still a work-in-progress**~

Found time to work on #97 some more, so opening a Draft PR to keep track of things.

This ended up being a much-needed rewrite of the Type Dictionary, with a heavy emphasis on performance and "correctness". Note the contents of `TypeDictionary.Get(Type gtype)` is effectively placeholder code. Thankfully, the type dictionary is largely self-contained, so this doesn't break anything else (e.g. all samples run without issue, and startup might be faster too).

At the moment, this includes:
 * Rewrite of lookup code to search for non-registered types as well
 * Initial Assembly Lookup code (also placeholder code for now)
 * Split between singular and recursive registration
 * Optimisations when using `WrapPointerAs<T>`

Since #102 improves our organisation of a lot of type registration code, I think it would be easier for me to rebase this on #102 once that's finished (as this PR is quite some time off, anyway).

Fixes #97, #66


badcel:
We have to see what will happen to the type dict. I'm currently working on the type integration and did several thinks to the type dictionary (see #102):

1. Split the type dictionary into 2 classes to achive one responsibilty per class. (Typedictionary / TypeRegistration).
2. Remove the recursive registration process. I think we have to somehow get into a deterministic state of the known types. For me scanning the object hierarchies is not the right solution. But there is no solution to this in my current implementation. (E.g. for composite templates we don't know that there is a button inside the template. If no other Button is known outside of the template we won't be able to create this button.)

In general I found out, that GTKSharp is doing some tricks to get class instantiation right. I hope to get s.th. working which just matches the idea of GObject and thus will be easy to follow along and extend. 

Currently this will probably go to the expense on usability for the user of the lib. But if we have the"standard" solution working we can think of possibilities to make it easier for the user.

There is a lot of things to figure out. In general my branch can currently create objects from the template (but the demo is not working due to several other errors which are still present, but hey :-)

badcel:
> Initial Assembly Lookup code (also placeholder code for now)

If this means scanning arbitrary assemblies for types I'm strongly against it. I'm not 100% sure but I think Xamarin is doing s.th. like this and they are very unhappy with it. I think they even get rid of it with MAUI as it is one reason for slow application startups.

firox263:
> If this means scanning arbitrary assemblies for types I'm strongly against it. I'm not 100% sure but I think Xamarin is doing s.th. like this and they are very unhappy with it. I think they even get rid of it with MAUI as it is one reason for slow application startups.

I agree. This feels very sloppy doing it this way, but at the same time, I haven't yet figured out a way around this (given I believe GtkSharp does it like this too? I haven't checked actually). The point of this PR is basically to clean up our typedict stuff, so it at least works reliably for the mean time. Since we're already doing most of this, I figured it's a decent interim fix in lieu of a more extensive redesign.

I might look into how PyGObject and GJS handle this same proxy/native relationship, since I don't recall seeing a typedict.

firox263:
I think this is about done.

I expect to throw most of this away in a few months time whenever we get around to a proper static typedict/typemap Xamarin-style. In the mean time though, this is a huge improvement over what we currently have and it means Hyena/Composite Templates/etc can work without any hacks.

I'm thinking it might be a good idea to merge this now, before #102, as that gives @badcel the flexibility to reorganise this how he wants with the new type dict/type registration/type descriptor split.

I don't have actual numbers, but this should be quite a bit faster than before because of the optimisations/fast paths I added to WrapPointerAs<T> which avoids a typedict lookup most of the time.

Libs/GObject/Classes/Object.cs
```diff
@@ -25,14 +26,14 @@ public partial class Object : INotifyPropertyChanged, IDisposable
         #endregion
 
         #region Properties
-
-        protected internal IntPtr Handle { get; private set; }
+        
+        public IntPtr Handle { get; private set; }
```
na2axl, line 30: The Handle property became public (again 😅)

Libs/GObject/Classes/Object.cs
```diff
@@ -76,7 +77,7 @@ public Object(params ConstructParameter[] properties)
                     values[i] = prop.Value;
                 }
 
-                // Create with properties
+                // Create with propeties
```
na2axl: Typo regression

Libs/GObject/Classes/Object.cs
```diff
@@ -104,7 +105,7 @@ public Object(params ConstructParameter[] properties)
         }
 
         /// <summary>
-        /// Initializes a wrapper for an existing object
+        /// Initialises a wrapper for an existing object
```
na2axl, line 108: Typo regression

Libs/GObject/Classes/Object.cs
```diff
@@ -128,12 +129,12 @@ private void Initialize(IntPtr ptr)
             RegisterProperties();
             RegisterOnFinalized();
 
-            // Allow subclasses to perform initialization
+            // Allow subclasses to perform initialisation
```
na2axl, line 132: Typo regression

Libs/GObject/Classes/Object.cs
```diff
@@ -128,12 +129,12 @@ private void Initialize(IntPtr ptr)
...
             Initialize();
         }
 
         /// <summary>
-        /// Wrappers can override here to perform immediate initialization.
+        ///  Wrappers can override here to perform immediate initialisation
```
na2axl, line 137: Typo regression

Libs/GObject/Classes/Object.cs
```diff
@@ -261,19 +262,22 @@ protected internal void OnPropertyChanged([CallerMemberName] string? propertyNam
...
         // and returns it.
-        public static T WrapPointerAs<T>(IntPtr handle)
+        protected static T WrapPointerAs<T>(IntPtr handle)
+            where T : Object
         {
             if (TryWrapPointerAs<T>(handle, out T obj))
                 return obj;
 
-            throw new Exception($"Failed to wrap handle as type <{typeof(T).FullName}>");
+            throw new Exception("Could not wrap handle");
```
na2axl: I think the fully detailed exception is better...

Libs/GObject/Classes/Object.cs
```diff
@@ -284,31 +288,57 @@ protected internal static bool TryWrapPointerAs<T>(IntPtr handle, out T o)
...
 
-            // If it is not found, we can assume that it
-            // is NOT a subclass type, as we ensure that
-            // subclass types always outlive their pointers
+            // If it is not found, we can assume that it is NOT a subclass type,
+            // as we ensure that subclass types always outlive their pointers
             // TODO: Toggle Refs ^^^
 
-            // Resolve GType of object
+            // Resolve gtype of object
```
na2axl, line 295: Typo regression

na2axl (commented):
Nicely done with this @firox263, just found theses little regressions. 👍🏾

Libs/GObject/Classes/Object.cs
firox263: Funnily enough, I just pushed a commit that fixed this about ten seconds ago.

Libs/GObject/Classes/Object.cs
firox263: Yeah, this is another regression from the merge. Fixed in latest commit

Libs/GObject/Classes/Object.cs
firox263: While this is correct British English, I'm not that petty 😅.

I imagine this is another regression from the merge. I'll push a commit that reverts these.

Libs/GObject/Classes/Object.cs
firox263: This one is sort-of intentional (See: https://github.com/gircore/gir.core/pull/105/commits/9c9e45a3852eee3dba77d166bce1a4b743cd9e6d)

The commit message:

```
It turns out that we cannot use `protected internal` after all when accessing a less derived class from a more-derived
class (e.g. Accessing Widget.Handle in Box).

See CS1540: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1540

Also makes Disposed private, as it is not used outside the class.
```

I wanted to clarify this, because I could just be missing something. Could you try using `protected internal` again and see if it builds? The problem I'm experiencing is where we can't access Handle from a more-derived version of a type (e.g. we can go Widget->Box, but not Box->Widget).

na2axl: Isn't for that cases the `GetHandle` method is used ? To do `GetHandle(widget)` in the `Box` class for example.

firox263: Hmm. Seems odd that `GetHandle()` would work, but not `Handle` directly.

Maybe we should consider making `Handle` private and forcing `GetHandle()` for everything to be consistent? (For another PR, of course).

Anyways, I'll push another commit that changes this.

na2axl: > Maybe we should consider making Handle private and forcing GetHandle() for everything to be consistent? (For another PR, of course).

I also think is the better solution, I'm already using `GetHandle` for everything in #103

firox263: I edited that comment thinking we'd probably want to access a type's own Handle (hence the protected), but you raise a good point. There's no reason why we can't use the accessor method, so it might be best to enforce this.

na2axl (approved):
I'm OK with this PR, I'll let @badcel merge it if he is OK too.

Libs/GObject/Structs/Type.cs
```diff
@@ -49,6 +49,25 @@ public Type(ulong value)
 
         #endregion
 
+        // Print out the name of the gtype (for debugging purposes)
+        public override string ToString()
```
badcel, line 53: Can we use a separate function for this? Perhaps it is even only compiled into the binary with the `#define DEBUG` compile time constant?

To have a native C# behaviour I think we should keep our class names and the default implementation of the `ToString` method clean.

Libs/GObject/Structs/Type.cs
```diff
@@ -49,6 +49,25 @@ public Type(ulong value)
...
+        {
+            return Marshal.PtrToStringAnsi(Global.Native.type_name(_value));
+        }
+
+        public override bool Equals(object obj)
```
badcel, line 58: Do we have to override Equals? The type class is already a struct which is comparing its values automatically this should be obsolete?

Libs/GObject/Structs/Type.cs
```diff
@@ -49,6 +49,25 @@ public Type(ulong value)
...
+        {
+            if (obj is Type && ((Type)obj)._value == _value)
+                return true;
+
+            return false;
+        }
+
+        public override int GetHashCode()
```
badcel, line 66: This is doing nothing

badcel (changes requested):
@firox263 Great work! Thank you so much. I have some minor points we should clarify, all in all this looks already very good.

Libs/GLib/Classes/Filename.cs
```diff
@@ -0,0 +1,32 @@
...
+    public static class Filename
+    {
+        public static string ToUri(string filename, string hostname, out Error? error)
+        {
+            error = null;
+
+            var resPtr = Global.Native.filename_to_uri(filename, hostname, out IntPtr errPtr);
+
+            if (errPtr != null)
+                error = Marshal.PtrToStructure<Error>(errPtr);
```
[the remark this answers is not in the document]
firox263: Not entirely sold on the idea of throwing for everything (sometimes we might want to fail gracefully), but that's another discussion.

Switched to using exceptions for now 👍

Libs/GLib/Classes/Filename.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        public static string FromUri(string uri, string hostname, out Error? error)
+        {
+            error = null;
+
+            var resPtr = Global.Native.filename_from_uri(uri, hostname, out IntPtr errPtr);
+
+            if (errPtr != null)
+                error = Marshal.PtrToStructure<Error>(errPtr);
```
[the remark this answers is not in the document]
firox263: Fixed - see above

Libs/GObject/Classes/Object.TypeDictionary.cs
```diff
@@ -53,6 +57,48 @@ internal static void Add(System.Type type, Type gtype)
...
             }
 
+            /// <summary>
+            /// Recursively register <c>type</c> in the type dictionary. Prefer
+            /// <see cref="AddRecursive(System.Type, Type)"/> if the gtype is already known.
+            /// </summary>
+            /// <param name="type">The C# type.</param>
+            internal static void AddRecursive(System.Type type)
+            {
+                Type gtype = GetTypeDescriptor(type)!.GType;
```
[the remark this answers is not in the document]
firox263: Changed this to throw a more descriptive exception (as a type must always possess a TypeDescriptor to be registered), rather than just a null pointer exception. Prompts the user to check if they've passed a valid GObject, or some other type by accident.

Libs/GObject/Structs/Type.cs
firox263: Huh, didn't think you could do that.

That's much simpler, I've gone ahead and removed `Equals` 👍

Libs/GObject/Structs/Type.cs
firox263: This is needed for the implementation of `Equals`. Since we're not doing that anymore, it can be safely removed.

Libs/GObject/Structs/Type.cs
firox263: I'm not sure I fully agree with this.

`ToString` is generally used to give a more human readable representation of an object, so it makes sense to me to print the GType class name instead of the type id number, which doesn't tell us all that much (e.g. when passing it to Console.WriteLine, etc).

That comment is a mistake from when I first wrote it, as it's not really being used for debugging but rather as a more general logging/introspection tool.

I'm not opposed to making this a separate function, but then we lose the ability to simply plug the type into `Console.WriteLine` and/or read it in the debugger.

Thoughts?

Libs/GLib/Classes/Filename.cs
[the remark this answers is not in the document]
badcel: Yes we could always add a Try... Method which does not throw.

Libs/GObject/Structs/Type.cs
badcel: You are right. Actually I did not test the output of this. But the default implementation should just return e.g. "Button" or not? I assumed that with your change it returns s.th. like "GButton" which is not matching the class name anymore. So I can be wrong here.

Will check later or perhaps you have some more information?

firox263: Since it's the GType struct rather than the object, I imagine it would simply return the string "Type" whenever we call `ToString()` which doesn't seem all that useful (I haven't actually checked).

With this version it returns the name of the GType straight from GLib (so if we introspected a type id of e.g. 107, we'd be returned a string that says "GtkButton"), which makes the most sense for logging.

badcel: Oh I was completely in the wrong context 🙈 . This is the type class, my mind was in the object class.

badcel: I just used http://try.dot.net with `Console.WriteLine(typeof(System.IO.File).ToString());` and it returns  `
System.IO.File` So I agree to your change completely. Thanks for claryfing. I'm going to merge :+1:

badcel (approved):

PR #106, opened 2020-11-11
Fix Garbage Collection of Delegates

### LINKED ISSUE #82: Garbage Collection frees needed objects
We have several instances where delegates and callbacks get garbage collected when the computer is low on RAM. A typical error message looks something like this:

```
Process terminated. A callback was made on a garbage collected delegate of type 'GObject!GObject.WeakNotify::Invoke'.
   at Gtk.Global.main()
   at Gtk.Global.main()
   at Gtk.Global.Main()
   at GtkDemo.Program.Main(System.String[])
```

Once #52 is merged, I will try look into this along with some other changes to lifecycle code (e.g. ToggleRefs), so not very pressing at the moment.

firox263:
Keeps a reference to any delegates we use to avoid upsetting the garbage collector when low on RAM.

The culprits seem to be the following two delegates, which I've gone ahead and fixed:
 - `MarshalCallback` in `ClosureHelper.cs`
 - `OnFinalized` in `Object.cs`

I'm not finding anything else, but if we do encounter this later the fix is rather straightforward. We can avoid this by holding a strong reference to any delegates we pass to GLib.

Fixes #82

### DIFF
--- a/Libs/GObject/Classes/ClosureHelper.cs
+++ b/Libs/GObject/Classes/ClosureHelper.cs
@@ -12,6 +12,10 @@ namespace GObject
 
         private static readonly Dictionary<Delegate, ClosureHelper> Handlers = new Dictionary<Delegate, ClosureHelper>();
 
+        // We need to store a reference to MarshalCallback to
+        // prevent the delegate from being collected by the GC
+        private readonly ClosureMarshal _marshalCallback;
+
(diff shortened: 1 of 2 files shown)


badcel (approved):
Looks good to me. Thank you.

na2axl (approved):
LGTM 👍🏾

PR #113, opened 2020-11-22
Feature/cleanup templates

### LINKED ISSUE #111: Update to new scriban 3.0
There is an update of scriban to version 3.0: https://github.com/lunet-io/scriban

badcel:
Fixes #111


firox263 (changes requested):
Nice work! Looks good to me from an organisation standpoint. Some regressions:

 - We're not printing signal descriptor names anymore:
image

 - Obsolete tag breaks things:
image

 - Can we [remove this](https://github.com/gircore/gir.core/blob/feature/cleanup_templates/Generator/Templates/GObject/class.sbntxt#L44-L48)? Since we're not generating properties (only strings inside Native), this shouldn't do anything. I don't think [property.sbntxt](https://github.com/gircore/gir.core/blob/feature/cleanup_templates/Generator/Templates/Shared/property.sbntxt) gets used either in that case?
image


Also some problems with whitespace:

 - Native->Properties aren't being put on new lines
image

 - Doc comments lost the whitespace between the triple-slash and the tag (e.g. `///<summary>` vs `/// <summary>`:
image

 - Inconsistent spacing for region tags
image

na2axl:
@badcel I've fixed most of these things in the widget catalog PR maybe you should cherry pick the commit?

badcel:
I think this fixes all the points. It looks much better now. But for me all the whitespace is a little bit like magic. It somehow works but it is never 100% clear why or what is the correct way to do it. Do you guys have some idea how to get consistency into it?

I did not take a look into it, but perhaps a different templating engine would be better suited? E.g. I saw that there is s.th. like: https://github.com/toddams/RazorLight Do you have any experiences with with Razor templates? Would it be easier to create well formated code with it?

badcel:
Ah I just saw this one: https://github.com/lunet-io/scriban/issues/233

This could help us alot with the indents. I'm going to try to use it. Then the only remaining thing is the generation / avoidence of empty lines which is not very clear to me. In the documentation it looks quite easy though: https://github.com/lunet-io/scriban/blob/master/doc/language.md#14-whitespace-control

firox263:
> I think this fixes all the points. It looks much better now. But for me all the whitespace is a little bit like magic. It somehow works but it is never 100% clear why or what is the correct way to do it. Do you guys have some idea how to get consistency into it?

I'm glad I'm not the only one with this problem. I'm finding it very hard to pretty print whitepsace with Scriban, although it looks like that change to indentation should make things much easier. I'm not sure if it's worth switching to another templating engine, but with the generator rewrite we should look at decoupling templates from the generation, so it gives us that option if we want.

Anyhow, this looks very good. I'll go ahead and merge 👍

firox263 (approved):

PR #116, opened 2020-12-05
Move to DotNet 5.0

badcel:


badcel:
@firox263 @na2axl : Any comments or reservations regarding .NET 5? Can we merge this?

I think we should make use of the most recent technologies as we will probably benefit a lot from it.

na2axl (approved):
LGTM, I've checked and the modified code looks fine.

firox263:
One small point - Do we want to use `String.Empty` or empty string literals (`""`)? They both compile to the same thing so it's purely a stylistic choice. I like `String.Empty` because it's a bit more readable when you have lots of `+-=?` operators. Thoughts?

Anyhow, looks good to me 👍

na2axl:
According to @firox263 I will be favorable on the usage of `string.Empty` as a code style standard too.

badcel:
I agree with you `string.Empty` is more explicit. I updated the code accordingly. So this could be merged.

firox263:
I think this is ready to merge.

From our discussion in #103, let's merge this first, then check #117 works with it.

PR #118, opened 2020-12-06
Object.TypeDictionary: Extract TypeDescriptor

badcel:


badcel:
@na2axl @firox263 : I extracted the TypeDescriptor part and added some unit tests for the extracted code. What do you guys think?

badcel:
@na2axl @firox263 : I updated this to the current develop branch.

What do you think of it? Can we merge?

Tests/GObject.Tests/Classes/TypeDescriptorRegistryTest.cs
```diff
@@ -0,0 +1,85 @@
+﻿using System.Collections.Generic;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
```
na2axl, line 2: I've not a strong opinion on this but is it better to use Microsoft VS Test tools or use something like NUnit or xUnit ?

badcel: I don't have any practical nunit experience. I use xunit at work and it get's the job done. If it get's a little more complicated like with STA threads it get's complicated with xunit. There is even a separate nuget for it.I would like to try [mstest V2](https://github.com/microsoft/testfx) because it looks like microsoft has put some thought into the architecture and is quite extensible as you can influence how tests are called or you can pretty easily extend the Assert class.

 - Customizing: https://github.com/Microsoft/testfx-docs/blob/master/RFCs/003-Customize-Running-Tests.md
 - Custom Assert: https://github.com/Microsoft/testfx-docs/blob/master/RFCs/002-Framework-Extensibility-Custom-Assertions.md

From the basic features all 3 test frameworks offer the same features, they just differ a bit in the assertions class and in the name of attributes. For me it looks like mstest V2 is in between xunits "simplistic" approach and the "more features" approach of nunit as it looks quite simple but offers flexibility and extensibility.

For [BindingSharp.Test](https://github.com/badcel/BindingSharp/tree/master/BindingSharp.Test) I used mstest, too and for those straight forward tests it worked fine.

For me this is like a test. If it does not scale or we see problems with the framework we can switch to another one. I'm interested to see if we could profit from the extensibility of this framework, even if it is probably not as widely adopted as nunit / xunit.

na2axl: I don't really have a comparison opinion on these frameworks, I've mostly used xUnit over other tests engines, and I've kept using it as an habit... So I will go with whatever you choose 👍🏾

na2axl (approved):
LGTM 👍🏾

PR #126, opened 2020-12-12
Feature/generate documentation file

badcel:
- This generates the xml documentation file in release mode.
- As warnings are errors it supresses warning CS1591 which warns about missing xml comments for public members
- The comments from the gir files are not generated into the wrapper code if in release mode. In this way we can create a public documentation based on our own comments.
- In debug mode the the gir comments are still generated into the wrapper code.

### DIFF
--- a/Build/Program.cs
+++ b/Build/Program.cs
@@ -11,6 +11,7 @@ namespace Build
         #region Fields
 
         private static string Configuration = ConfDebug;
+        private static bool GenerateComments = false;
 
         private static readonly string[] TestProjects =
         {
@@ -100,10 +101,15 @@ namespace Build
                 (project) => DotNet.Test(project, Configuration)
             );
 
+            Target(Targets.Comments, () => GenerateComments = true);
+            
             Target(Targets.Release, () => Configuration = ConfRelease);
-            Target(Targets.Debug, () => Configuration = ConfDebug);
+            Target(Targets.Debug,
+                DependsOn(Targets.Comments),
+                () => Configuration = ConfDebug
+            );
 
-            Target("default", DependsOn(Targets.Build));
+            Target("default", DependsOn(Targets.Debug, Targets.Build));
             RunTargetsAndExit(args);
         }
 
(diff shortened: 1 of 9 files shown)


firox263 (approved):
Looks good to me. I like this approach as it keeps documentation as part of the generator.

One point, maybe before we merge: We should probably make comments explicitly opt-in, as running `dotnet run` by default generates comments (being a debug build), and someone could accidentally distribute that without knowing any better (which would be an LGPL license violation).

Maybe we only generate docs when `comments` is passed to the build tool, and explicitly mention this in the README.

Otherwise, nice work! This feels like a reasonably elegant solution for documentation going forwards.

badcel:
Yes it is better to be conservative with the comments. 

We should probably make the generation of the documentation file opt in, too. This gives us the maximum flexiblity to enable / disable the comments, documentation file in any combination with debug / release mode. 

I'm going to try a commandline parser to improve the ```Build``` commandline interface as currently everything is a target. There is the very promising looking [dragon fruit commandline parser](https://www.nuget.org/packages/System.CommandLine.DragonFruit/) which is part of the dotnet organisation and pretty easy to use in our case: https://github.com/adamralph/bullseye/blob/master/BullseyeSmokeTester.DragonFruit/Program.cs

badcel:
I updated the code to use the DragonFruit commandline parser and restructured it to better separate the concerns and prepare for unit tests.

@firox263 @na2axl What do you think?

firox263 (approved):
Really great work here.

Layout wise, this looks significantly better to me over what we had before. I think this is a big improvement in how we're structuring the build tool, and decoupling the generator now is especially nice to have, since it'll make the inevitable generator rewrite easier to work with (I have some ideas for that w.r.t. gstreamer - On that note, will open a PR soon for my [feature/gstreamer](https://github.com/firox263/gir.core/tree/feature/gstreamer) branch pending some final cleanup).

I hadn't heard about DragonFruit before, but I like it. It's a very clean way of handling command line arguments. I'm not personally familiar with Moq, so I can't comment on that, but it looks good to me.

I've tested this and everything works fine on my end, so I think we can merge 👍

Issue #119, opened 2020-12-08
Closures Handling


na2axl:
As discussed in #103, this issue is to find a solution about closures handling in Gir.Core libraries.
This issue can be related to #109 

---

_**@firox263:** Borrowing this to add a TODO_

**TODO 6/05/21:**
 - [ ] Reimplement Native->Managed marshalling (so we can call the user's provided delegate)
 - [ ] Properly exclude certain parameters like `userData`
 - [ ] Write unit/integration tests

**Completed:**
 - [X] Prototype Notified
 - [X] Prototype Async
 - [x] Prototype Default/Call
 - [x] Generator should use Native-suffix in native function calls
 - [x] Generator needs a way to marshal any arbitrary type from Native->Managed (this may require the generator rewrite to achieve?)
 - [x] Come to a decision whether DestroyNotify is useful in the end user API
 - [x] Implement Notify in Generator
 - [x] Implement Async in Generator
 - [x] Implement Call in Generator
 - [x] Fix DBusConnection and a few existing uses
 - [x] Memory Leak Testing

firox263:
As part of this, we should also tackle delegate/callback signatures. Currently we generate delegates with IntPtr for most parameters, but obviously we wouldn't want users to have to marshal themselves. We should modify the generator to generate a managed delegate and associated helper class for each callback which handles marshalling automatically (similar to how we populate TSignalArgs).

If we implement this using closures we can probably kill two birds with one stone here, and also reduce the amount of boilerplate we have to write in methods.

firox263:
Some ideas on how to handle delegate resources (to avoid the GC):

There is a scope parameter in GIR metadata for parameter.

See https://developer.gnome.org/gi/unstable/gi-GIArgInfo.html#GIScopeType

It has four possible values:
 - Invalid (not a callback)
 - Call (only used during the function - can free in-func)
 - Async (only ever called once - free after use)
 - Notified (DestroyNotify to free the callback resources)

For example, GtkTickCallback:

```xml
<parameter name="callback" transfer-ownership="none" scope="notified" closure="1" destroy="2">
    <doc xml:space="preserve" filename="gtk/gtkwidget.c" line="5301">function to call for updating animations</doc>
    <type name="TickCallback" c:type="GtkTickCallback"/>
</parameter>
```

From https://gi.readthedocs.io/en/latest/annotations/giannotations.html

 - call (default) - Only valid for the duration of the call. Can be called multiple times during the call.
[...]

Issue #110, opened 2020-11-20
Full Interface Support


mjakeman:
While trying to implement `TreeSelection`, I ran into an issue where I needed to get a GObject of some (theoretically) unknown type and return it as an interface. We should make `Try/WrapPointerAs` support returning interfaces as well as objects.

After #102, I think interfaces are the only "big thing" left in terms of type integration before we've covered most of the GObject type system. Let's use this to track interface support.

**To Do:**
 - [x] Support interface properties (#104)
 - [ ] Support interfaces in `Try/WrapPointerAs`
 - [ ] Allow user to implement GInterfaces for their subclasses 
   - This will need to be done during type registration for GObject to recognise it

**Future:**
 - [ ] Allow the user to export custom C# interfaces as GInterfaces?
   - Would be similar to object registration
 - [ ] Prefix interfaces with `I` for consistency
   - Would be done alongside the generator rewrite

na2axl:
Hi @firox263, you made a good point with this, maybe you can see if what I'm trying to do in #103 it is useful to you. I have:

1. Created an `IObject` interface that will be the real _root_ for every GObjects, so currently in #103 the `Object` class implement `IObject` and also every generated GInterfaces.
2. Used modifications from 1. to implement GInterface properties as described in #104, for now only static property descriptors, 
an empty C# property (which will require implementation in classes) and a `Native` class are generated. 

Maybe the fact that the `IObject` interface is the root node of the class/interface type hierarchy may help you ?

firox263:
@na2axl This looks very good to me. Having interface properties generated already knocks one thing off this list.

The next thing would be mapping between GInterfaces and their GTypes like we're doing for objects. I'll experiment on top of your branch a bit, and see if I can get something working.

badcel:
I did not check the code just in case it applies here: we could try do make smaller iterations on the develop branch and slowly get things working. 

@na2axl if complete parts of your code are already working as expected let's put them into develop.

I'm thinking about the same for my type integration or. It should probably be split into several smaller ones as it contains several features.

firox263:
Some example code:

Assume we have a subclass which implements the `Orientable` interface.

```cs
public class MySubclass : Gtk.Widget, Orientable
{
    public Orientation Orientation
    {
        get => GetProperty(Orientable.OrientationProperty);
        set => SetProperty(Orientable.OrientationProperty, value);
    }

    public void SetOrientation(Orientation orientation)
[...]

PR #127, opened 2020-12-16
WIP: GStreamer

### LINKED ISSUE #112: Support GStreamer (Gst Plugins)
We should support using GStreamer with Plugins (as base GStreamer isn't all that useful). I think it's best we get this working ahead of any generator rewrite, since GStreamer uses GObject directly.

 - GStreamer plugins have individual gir metadata files (see [gstreamer-rs/gir-files](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/tree/master/gir-files)).
 - It would make sense to namespace plugins under `Gst`, for example `GstAudio` -> `Gst.Audio`.
   - I'm not sure if we can infer this from gir data?
 - Parts of GStreamer rely on [first-class interface support](https://gstreamer.freedesktop.org/documentation/application-development/advanced/interfaces.html?gi-language=c), so this will include #110 and #104.
 - Outside of Gtk, GStreamer is the other big client of GObject, so there'll be [intricacies](https://gstreamer.freedesktop.org/documentation/application-development/basics/elements.html?gi-language=c#using-an-element-as-a-gobject) to work around.
   - While typed properties (via property descriptors) are great and should be used whereever possible, GStreamer relies on being able to retrieve properties from string representations too.
 - Easy GStreamer <-> Gtk interop for writing media-enabled applications.

Considerations:
 - Investigate [controlled properties?](https://gstreamer.freedesktop.org/documentation/gstreamer/gstobject.html#controlled-properties)
 - Allow for access to public struct member data (in the case of [GstObject->flags](https://gstreamer.freedesktop.org/documentation/gstreamer/gstobject.html#members))
 - Fix clash in generator between capitalisation and reserved keywords (e.g. `null`->`@null`, whereas it should be `null`->`Null`, see `Gst.State`).

firox263:
This PR implements the groundwork for GStreamer, as well as substantial amount of GStreamer itself (enough for a music player, with caveats). It implements parts of GStreamer, GstBase, GstAudio, GstVideo, and GstPbutils.

**This PR depends on:**
 - #110: Interface Support (only partially)
 - ~#119: Closures and Delegate Signatures~

**Remaining Tasks**
Here's the laundry list of TODOs before we can merge this:
 - [x] **Rebase on `develop`:** Needs changes to work with Scriban 3.0 and our reworked templates.
 - [x] **GIR Files:** Regenerate relevant GStreamer gir files and add them to [gir-files](https://github.com/gircore/gir-files).
 - [x] **Object Struct Fields:** Decide on safe vs unsafe access to GObject fields
     - See GstObject->Flags for an example of safe vs unsafe access
 - [x] **GhostPad Constructor Problems:** GhostPad's constructor implements custom behaviour which we can't easily implement in a C# constructor (validation of construct time properties).
     - For now, I've created a static `New()` method.
     - A long-term solution to this could be C# 10 Factories, which are static methods that support object initialiser syntax (see https://www.youtube.com/watch?t=6135&v=kU-t1qsEbqk)
 - [X] ~**Need design for GstIterator:** How should this work with C# Enumerable/Enumerator?~ **Tentative design in #157.**
 - [x] **Need design for GstTagList:** GstTagList needs a rethink now that implicit Values have been removed. (**Edit:** TagList's API has been simplified to use GValue directly for now. This seems to work fine).
 - [x] **Decide on Public API:** Mainly regarding access e.g. `Error`, `WrapPointerAs`, `GetHandle`, etc.
     - We also need to decide on `GStreamerGlueGetProperty`/`GStreamerGlueSetProperty`.
     - Rename `WrapPointerAs` and `GetHandle` to `WrapHandle/GetHandle`
 - [x] ~**Fix Delegate Signatures (#119):** We need a way to marshal between managed signatures and unmanaged signatures~
     - ~As part of this, remove Timeout->FuncData as it does not work~ **Dropped from this PR**

**Future Changes**
GstMessage and many other structs depends on support for custom Value types in the generator. This will allow us to use `ref struct` rather than the complicated and unsustainable (maintenance-wise) marshalling I'm using at the moment. This can reduce method bodies from >20 lines to sub-5, as we won't need to explicitly marshal anymore. It's probably much safer this way as well, as there is little chance of memory corruption from forgetting to free or update something.

Fixes #112 

@gircore/coreteam


badcel:
I updated the gir files in the develop branch. This should fix ```GIR Files: Regenerate relevant GStreamer gir files and add them to gir-files.```

firox263:
In the process of rebasing on develop. It's very messy, but I think I'm almost there.

We're currently blocking on a problem with DllImports in the new template code. I'm getting:
```
??Ignoring method because dll import is missing
??Ignoring method because dll import is missing
```
spammed in the console when generating and it's preventing interface native functions from being created.

Any idea what's causing this? I'm getting this on develop too @badcel

firox263:
Rebased on develop. Some new issues:
 - We have a problem with enums in GstVideo. For some reason, we're getting the enum member `Onefield` generated twice in two different structs. These need to be manually commented out for the solution to build. This will need some kind of generator workaround.
 - For some reason, Signal Args marshalling has broken in MarshalCallback (e.g. for Notebook's `SwitchPageSignalArgs`, we'll have a `nParamValues` of 2 but `paramValues` will only have a length of 1). This blows up as soon as we try and use it. I've added an assertion which aborts if this happens.

badcel:
> Edit: Think I've figured it out - I wasn't passing enough arguments to the new method.sbntxt. I'm remodelling interface.sbntxt on top of the new class template and it seems to be working.

Great you found something, this could be a bug in ```develop```, too. Ping me if I should check the code or check regarding your found issues.

firox263:
Ahh, looks like the SignalArgs problem is a regression in the generator. We've lost `[MarshalAs(UnmanagedType.LPArray, SizeParamIndex=2)]` along the way at some point. 

Shouldn't be too difficult to re-add.

badcel:
It fails with those messages:

```
/home/vsts/work/1/s/Libs/GObject/Classes/Object.TypeIntegration.cs(137,26): error CS1572: XML comment has a param tag for 'handle', but there is no parameter by that name [/home/vsts/work/1/s/Libs/GObject/GObject.csproj]
/home/vsts/work/1/s/Libs/GObject/Classes/Object.TypeIntegration.cs(152,26): error CS1572: XML comment has a param tag for 'obj', but there is no parameter by that name [/home/vsts/work/1/s/Libs/GObject/GObject.csproj]
```

I like the new xml-documentation build feature as it detects broken xml comments :rocket:

badcel:
@firox263  I updated the GStreamer gir files from https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/tree/master/gir-files . This seems to be a good source. The files are already in develop.

I tested your changes once without the enum workaround code and it worked out. Could you please remove the workaround and verify?

firox263:
I'm going to focus on #119, as this PR appears to reintroduce the bug with garbage collection of delegates (#82). I think we should go further and enforce that all closures/callbacks must go through ClosureHelper + a managed wrapper to ensure that we don't have non-GC-rooted delegates littered around the codebase

firox263:
Looking at what's left to get this merged. Rather than trying to get 100% of GStreamer in first try, I think it would be better to add the remaining bits in incremental chunks (that way, we can get the generator changes in develop as soon as possible).

> **Object Struct Fields:** Decide on safe vs unsafe access to GObject fields
    See GstObject->Flags for an example of safe vs unsafe access

I'm thinking for object struct fields, we should go with the safe access method I'm currently using. It'll be easy to switch to unsafe down the line if we find that performance is not up to par, but I'd like to keep as much of the bindings using safe C# as possible (making it easier to diagnose other issues - at least until we have better tests in place).

> **Need design for GstTagList:** GstTagList needs a rethink now that implicit Values have been removed.

I will probably drop TagList from this PR for now. It's not being used by anything (yet) and rather than mimicking the public API of mono gtk-sharp, it'd be best to come up with a more C#-like pattern for this.

> **Need design for GstIterator:** How should this work with C# Enumerable/Enumerator?

I have some ideas for this. I want to try and land this in this PR.

> Fix Delegate Signatures (#119): We need a way to marshal between managed signatures and unmanaged signatures
    As part of this, remove Timeout->FuncData as it does not work

I'm thinking of dropping the delegate rework from this PR and working on it next (separately). This is less of an obstacle to merging now as it also affects `develop`, and it might be a bit too much to fit in to a single PR.

> **Decide on Public API:** Mainly regarding access e.g. Error, WrapPointerAs, GetHandle, etc.
    We also need to decide on GStreamerGlueGetProperty/GStreamerGlueSetProperty.
    Rename WrapPointerAs and GetHandle to WrapHandle/GetHandle

Let's:
 * Make Object.GetProperty and Object.SetProperty have overloads which take a string (or better - support creating a descriptor from a string) as we talked about, and get rid of the GStreamerGlue functions.
 * Rename WrapPointerAs and GetHandle to WrapHandle/GetHandle and work them into using IHandle from #133.
 * I think Error is fine as it is with `HandleError`.

Thoughts?

@gircore/coreteam

badcel:
I fully agree on all points. I have one suggestion regarding the `SetProperty` methods.

We already have a `private void SetGProperty(Value value, string? name)` method and a corresponding getter method.  I would suggest that we rename those methods to `SetProperty` / `GetProperty` to create the desired overloads. Additionally we should make those methods protected so they are accessable from `Element`. This looks like the most simple and clean approach to me. I think we would avoid some kind of descriptor for now. We can add it later if we see the need.

On a side note, I don't know why the name parameter is last in the parameter list? This looks kind of strange to me. I would suggest switching the order of the parameters: `protected void SetProperty(string name, Value value)`.

From my point of view just go ahead all your suggested changes / planings are great. So let's make smaller iterations and get this into develop :+1: 

---------------------------

Less important remark regarding `Element.cs`, there is this code:

```C#
 // We intentionally throw an exception if the type of value cannot be wrapped
 // TODO: Support boxing arbitrary managed types
Value val;
if (value?.GetType().IsAssignableTo(typeof(GObject.Object)) ?? false)
    val = Value.From((Object) value!);
else
    val = Value.From(value);
                    
GStreamerGlueSetProperty(property, val);
```

I think we should in some later point in time verify if this code could somehow be part of the `GObject` library so the `protected SetProperty` API allows to set arbitrary objects and is doing the right thing. The signature would be s.th. like `protected void SetProperty(string name, object value)`.  I'm not sure if we can reach that goal, but perhaps you just add a TODO flag and we reconsider later?

Libs/Gst/Classes/Parse.cs
```diff
@@ -10,16 +10,24 @@ public static class Parse
...
         /// <returns>A new element</returns>
         /// <exception cref="GLib.GException">Throws an exception in case of an error</exception>
-        public static Element Launch(string pipelineDescription)
+        public static Element? Launch(string pipelineDescription)
         {
-            //TODO: As this method wrapps a global function it should probably be part of the toolkit layer.
-            
             IntPtr ret = Global.Native.parse_launch(pipelineDescription, out IntPtr error);
 
             if (error != IntPtr.Zero)
```
badcel, line 17: Use `Error.HandleError`.

Libs/Gst/Classes/Parse.cs
```diff
@@ -10,16 +10,24 @@ public static class Parse
         /// <param name="pipelineDescription">Description of the pipeline</param>
         /// <returns>A new element</returns>
         /// <exception cref="GLib.GException">Throws an exception in case of an error</exception>
-        public static Element Launch(string pipelineDescription)
+        public static Element? Launch(string pipelineDescription)
```
badcel, line 13: This does not return null as WrapHandle will throw an exception in case of an error? So this should be `public static Element Launch(string pipelineDescription)`?

Libs/Gst/Classes/Parse.cs
```diff
@@ -10,16 +10,24 @@ public static class Parse
...
 
-            return new Element(ret); //Parse always returns a new object
+            return GObject.Object.WrapHandle<Element>(ret);
+        }
+
+        public static Element? BinFromDescription(string binDescription, bool ghostUnlinkedPads)
+        {
+            IntPtr ret = Global.Native.parse_bin_from_description(binDescription, ghostUnlinkedPads, out IntPtr error);
+            
+            if (error != IntPtr.Zero)
```
badcel, line 27: Use `Error.HandleError`.

Libs/Gst/Classes/Parse.cs
```diff
@@ -10,16 +10,24 @@ public static class Parse
...
+                GLib.Error.HandleError(error);
+            
+            return GObject.Object.TryWrapHandle<Element>(ret, out Element? element) ? element : null;
```
badcel, line 30: This should probably not return null but throw an exception to be in line with `Launch`.

I would suggest that we are throwing exceptions in the "standard" case for methods with an error. In case we don't want to throw I would suggest `TryXXX` as a name with a bool return and an `out` parameter to make the behaviour of the methods absolutely clear and to be able to differentiate them.

In this method there are sometimes exceptions thrown and sometimes a `null` is returned which should not be possible as it makes using the API very hard.

Libs/Gst/Gst.csproj
```diff
@@ -5,5 +5,7 @@
 
   <PropertyGroup>
     <PackageId>Gir.CS.Gst</PackageId>
+    <LangVersion>9.0</LangVersion>
```
badcel, line 8: `LangVersion` should not be needed as `Directory.Build.props` defines .NET 5 which uses C# 9 by default

Libs/Gst/Gst.csproj
```diff
@@ -5,5 +5,7 @@
...
+<!--    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>-->
```
badcel, line 9: Remove comment.

Libs/Gst/Structs/Message.cs
```diff
@@ -0,0 +1,141 @@
...
+    {
+        public MessageType Type => type;
+
+        public Structure GetStructure()
+        {
+            // Marshal this structure
+            IntPtr thisPtr = Marshal.AllocHGlobal(Marshal.SizeOf(this));
+            Marshal.StructureToPtr(this, thisPtr, false);
+            
+            // Native Call
```
badcel, line 25: I think we should avoid redundant comments. It is obvious that the following call makes a native call. The same goes for the rest of the comments of this method (except the "TODO" one). Comments should add additional value to the code which is not obvious for the reader. Even in case we need such a comment we should probably wonder why the code is not selfexplanatory enough. And if we think that we need the comment after this consideration we should keep it.

We can keep them for now but I would like to get rid of them bit by bit.

Libs/Gst/Structs/Message.cs
```diff
@@ -0,0 +1,141 @@
...
+            // Update structures (TODO: Check for NULL)
+            this = (Message) Marshal.PtrToStructure(thisPtr, GetType())!;
+            debug = Marshal.PtrToStringAnsi(strPtr);
+            error = Marshal.PtrToStructure<GLib.Error>(errPtr);
+
+            // Free Memory
+            Marshal.FreeHGlobal(thisPtr);
+        }
+
+        public Gst.Object Src
```
badcel, line 135: Properties should be at the top of the file.

Libs/Gtk3/Classes/Widget.CompositeWidget.cs
```diff
@@ -57,21 +57,21 @@ protected static void BindTemplateSignals(Type gtype, System.Type t)
...
-            if(!TryWrapPointerAs<Widget>(@object, out var eventSource))
+            if(!TryWrapHandle<Widget>(@object, out var eventSource))
                 return;
 
-            if(!TryWrapPointerAs<Widget>(connect_object, out var compositeWidget))
+            if(!TryWrapHandle<Widget>(connect_object, out var compositeWidget))
                 return;
 
-            MethodInfo? compositeWidgetCallbackMethodInfo = compositeWidget.GetType().GetMethod(
+            MethodInfo? compositeWidgetCallbackMethodInfo = compositeWidget?.GetType().GetMethod(
```
badcel, line 66: The question mark should not be needed if `TryWrapHandle` is fixed with the correct attribute (see corresponding comment).

Libs/Gtk3/Classes/Widget.CompositeWidget.cs
```diff
@@ -57,21 +57,21 @@ protected static void BindTemplateSignals(Type gtype, System.Type t)
...
                 name: handler_name, 
                 bindingAttr: BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance
             );
             
             if (compositeWidgetCallbackMethodInfo is null)
                 return;
 
-            EventInfo? sourceEvent = eventSource.GetType().GetEvent(
+            EventInfo? sourceEvent = eventSource?.GetType().GetEvent(
```
badcel, line 74: The question mark should not be needed if `TryWrapHandle` is fixed with the correct attribute (see corresponding comment).

Libs/Gtk3/Classes/Window.cs
```diff
@@ -1021,7 +1021,7 @@ public static void SetDefaultIconList(GLib.List list)
         public static bool SetDefaultIconFromFile(string filename)
         {
             var result = Native.set_default_icon_from_file(filename, out IntPtr e);
-            HandleError(e); // Throws if error
+            GLib.Error.HandleError(e); // Throws if error
```
badcel, line 1024: Remove comment. If we think the "throw information" should be there we should consider renaming the method to `Error.ThrowOnError` or s.th. similar. In this case we fix the ambiguity and add the information to all places in the code which use this method.

Libs/Gtk3/Classes/Window.cs
```diff
@@ -1043,7 +1043,7 @@ public void SetIconList(GLib.List list)
         public bool SetIconFromFile(string filename)
         {
             var result = Native.set_icon_from_file(Handle, filename, out IntPtr e);
-            HandleError(e); // Throws if error.
+            GLib.Error.HandleError(e); // Throws if error.
```
badcel, line 1046: Remove comment.

README.md
```diff
@@ -15,13 +15,13 @@ This project aims to provide:
...
-* [GTK] (wip): UI-Toolkit (GTK3 + experimental GTK4)
-* [WebKitGTK] (planed): Browser-Engine
-* [JavaScriptCore] (planed): Javascript integration for [WebKitGTK]
-* [libhandy] (planed): Convergent UI-Widgets for [GTK] to support mobile phones
-* [dbus] (wip): Library for inter-process communication via [GIO]
-* [libchamplain] (planed): Library to display maps
-* [gstreamer] (wip): Multimedia Framework
+* [GTK] (partial support): UI-Toolkit (GTK3 + experimental GTK4)
+* [GStreamer] (partial support): Multimedia Framework
+* [DBus] (wip): Library for inter-process communication via [GIO]
```
badcel, line 20: DBus should probably be marked with "partial support", too?

Generator/Templates/GObject/class.sbntxt
```diff
@@ -67,6 +67,22 @@ namespace {{ namespace }}
         #endregion
 
         {{~ end ~}}
+        
+        #region Fields
+                
+        [StructLayout(LayoutKind.Sequential)]
+        protected {{'new'|get_if_parent}} struct Fields : IObjectStruct
+        {        
+{{ # TODO: Do we want to use field structs or put the fields in objects directly?
```
badcel, line 76: I like the `Fields` struct :+1: Remove the TODO?

Generator/Templates/GObject/class.sbntxt
```diff
@@ -67,6 +67,22 @@ namespace {{ namespace }}
...
+                
+        [StructLayout(LayoutKind.Sequential)]
+        protected {{'new'|get_if_parent}} struct Fields : IObjectStruct
+        {        
+{{ # TODO: Do we want to use field structs or put the fields in objects directly? 
+if (has_parent) && (name != "InitiallyUnowned") && (fields.size > 0)
+    fields[0].type.name = string.append fields[0].type.name ".Fields" 
+end}}
+{{~ for field in fields
+    include 'field.sbntxt' field
```
badcel, line 81: The indentation is borken. To fix it it should probably look like the following and be indented at the correct level:

```
{{~ for $field in fields }}
{{ include 'signal.sbntxt' $field }}
{{ end ~}}
```

I think it is better to use `$field` instead of `field` in the loop as this is a local variable compared to a global one.

badcel (changes requested):
Great work, this is amazing :+1:  It looks like it involved a lot of research work, too.

I did not test if it works as there are no samples. But as the code itself looks good we can fix any remaining issues as they arise.

Nevertheless I left several comments which we should discuss but there is no show stopper.

I think we should add some simple samples but this is for another PR.

Libs/GLib/Classes/Filename.cs
```diff
@@ -11,17 +11,22 @@ public static string ToUri(string filename, string hostname)
 
             if (error != IntPtr.Zero)
                 throw new GLib.GException(error);
-
+
```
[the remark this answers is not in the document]
firox263: Should be fixed

Libs/GLib/Classes/MainContext.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace GLib
+{
+    public partial class MainContext
+    {
+        public readonly IntPtr Handle;
```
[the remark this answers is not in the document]
firox263: Yeah, this was supposed to be a TODO.

I've gone ahead and added `New()` and `Default()` methods, and implemented `IHandle`.

Libs/GLib/Classes/Variant.cs
```diff
@@ -79,8 +79,8 @@ private void Init(out IntPtr handle, params Variant[] children)
         public string GetString()
         {
             ulong length = 0;
-            IntPtr strPtr = get_string(_handle, ref length);
-
+            IntPtr strPtr = get_string(_handle, out length);
+
```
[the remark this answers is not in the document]
firox263: Should be fixed

Libs/GObject/Classes/Object.Properties.cs
```diff
@@ -26,17 +26,17 @@ protected T GetProperty<T>(Property<T> property)
...
         }
 
         /// <summary>
-        /// Defines the value of a native GProperty given its <paramref name="name"/>
+        /// Assigns the value of a GObject's property given its <paramref name="name"/>
         /// </summary>
         /// <param name="value">The property name.</param>
         /// <param name="name">The property value.</param>
-        private void SetGProperty(Value value, string? name)
+        public void SetProperty(string? name, Value value)
```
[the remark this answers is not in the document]
firox263: Fixed

Libs/GObject/Classes/Object.cs
```diff
@@ -336,15 +326,36 @@ protected internal static bool TryWrapPointerAs<T>(IntPtr handle, out T o)
...
 
-            return true;
+        /// <summary>
+        /// A variant of <see cref="WrapHandle{T}"/> which fails gracefully if the pointer cannot be wrapped.
+        /// </summary>
+        /// <param name="handle">A pointer to the native GObject that should be wrapped.</param>
+        /// <param name="o">A C# proxy object which wraps the native GObject.</param>
+        /// <typeparam name="T"></typeparam>
+        /// <returns><c>true</c> if the handle was wrapped, or <c>false</c> if something went wrong.</returns>
+        public static bool TryWrapHandle<T>(IntPtr handle, [MaybeNullWhen(false)] out T? o)
```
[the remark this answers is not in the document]
firox263: Didn't realise NotNullWhen was a thing, but that makes a lot more sense. Fixed.

Libs/Gst.Audio/Gst.Audio.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <ItemGroup>
+    <ProjectReference Include="..\Gst\Gst.csproj" />
+    <ProjectReference Include="..\Gst.Base\Gst.Base.csproj" />
+    <ProjectReference Include="..\GObject\GObject.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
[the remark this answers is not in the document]
firox263: Fixed

Libs/Gst.Base/Gst.Base.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <ItemGroup>
+    <ProjectReference Include="..\Gst\Gst.csproj" />
+    <ProjectReference Include="..\GObject\GObject.csproj" />
+    <ProjectReference Include="..\GLib\GLib.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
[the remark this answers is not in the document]
firox263: Fixed

Libs/Gst.Pbutils/Gst.Pbutils.csproj
```diff
@@ -0,0 +1,15 @@
...
+  <ItemGroup>
+    <ProjectReference Include="..\Gst\Gst.csproj" />
+    <ProjectReference Include="..\GObject\GObject.csproj" />
+    <ProjectReference Include="..\Gst.Base\Gst.Base.csproj" />
+    <ProjectReference Include="..\Gst.Audio\Gst.Audio.csproj" />
+    <ProjectReference Include="..\Gst.Video\Gst.Video.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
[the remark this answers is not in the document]
firox263: Fixed

Libs/Gst.Video/Gst.Video.csproj
```diff
@@ -0,0 +1,14 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="..\Gst\Gst.csproj" />
+    <ProjectReference Include="..\Gst.Base\Gst.Base.csproj" />
+    <ProjectReference Include="..\GObject\GObject.csproj" />
+    <ProjectReference Include="..\GLib\GLib.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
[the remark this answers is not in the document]
firox263: Fixed

Libs/Gst.Video/Interfaces/Navigation.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            // Marshal message structure
+            IntPtr messagePtr = Marshal.AllocHGlobal(Marshal.SizeOf(message));
+            Marshal.StructureToPtr(message, messagePtr, false);
+
+            // Native Call
+            var result = Global.Native.navigation_message_get_type(messagePtr);
+
+            // Update message structure (Not necessary?)
+            Marshal.PtrToStructure(messagePtr, message);
```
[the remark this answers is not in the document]
firox263: Fixed

Libs/Gst/Classes/Element.cs
```diff
@@ -1,9 +1,59 @@
...
+        //     }
+        // }
+
+        #endregion
+
+        public static Element? MakeFromUri(URIType type, string uri, string elementName)
+        {
+            IntPtr result = Native.make_from_uri(type, uri, elementName, out IntPtr errPtr);
+
+            if (errPtr != IntPtr.Zero)
```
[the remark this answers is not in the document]
firox263: Fixed

Libs/Gst/Classes/Element.cs
```diff
@@ -17,7 +67,125 @@ public Bus GetBus()
...
+            return true;
+        }
+
+        public Pad? GetRequestPad(string name)
+            => TryWrapHandle(Native.get_request_pad(Handle, name), out Pad? pad) ? pad : null;
+
+        public bool SyncStateWithParent()
+            => Native.sync_state_with_parent(Handle);
+        
+        // Some older mono applications appear to use a
```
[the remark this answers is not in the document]
firox263: Fixed

Libs/Gst/Classes/Parse.cs
firox263: Fixed

Libs/Gst/Classes/Parse.cs
firox263: Fixed

Libs/Gst/Gst.csproj
firox263: Fixed

Libs/Gst/Gst.csproj
firox263: Fixed

Generator/Templates/GObject/class.sbntxt
firox263: Fixed

Generator/Templates/GObject/class.sbntxt
firox263: I've managed to fix the indentation.

There's still a problem with spacing between each field being wildly inconsistent, not really sure how to go about fixing that.

Libs/GLib/Classes/MainLoop.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+
+namespace GLib
+{
+    public sealed partial class MainLoop
+    {
+        public readonly IntPtr Handle;
```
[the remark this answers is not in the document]
firox263: Fixed

Libs/Gst/Classes/Parse.cs
firox263: The actual method gst_parse_launch can return null, so it might be worth switching this to TryWrapHandle.

Libs/Gst/Structs/Message.cs
firox263: This is a leftover from when I was trying to fix memory corruption issues with GstMessage. I had comments and code everywhere, so it was just indicating that 'x' section was for native calling, 'y' was for marshalling, etc.

We can get rid of it.

firox263: Fixed in latest commit

Libs/Gst/Structs/Message.cs
firox263: Fixed

Libs/Gtk3/Classes/Widget.CompositeWidget.cs
firox263: Fixed

Libs/Gtk3/Classes/Widget.CompositeWidget.cs
firox263: Fixed

Libs/Gtk3/Classes/Window.cs
firox263: Fixed

Libs/Gtk3/Classes/Window.cs
firox263: Fixed. I've gone ahead and renamed HandleError to ThrowOnError like you suggested. I think it's more representative of what the method actually does.

Libs/Gst/Classes/Parse.cs
firox263: Fixed, see above.

README.md
firox263: I left it as-is because I'm not really sure how much/little we support it, other than the simple DBus sample. I'll go ahead and change it to partial though.

firox263:
That should be everything, although I might have missed something here and there.

I'm working on a few samples at the moment, I'll open a new PR for that once we land this one.

@badcel

Libs/Gtk3/Classes/Window.cs
badcel: :+1: 

> We should think of how to deal with non-fatal errors however. GStreamer has a few instances of "Recoverable Errors", but I haven't looked into that too much yet.

Yes there are probably a lot of those "smaller" issues. I created #159 to keep track of your thoughts.

Generator/Templates/GObject/class.sbntxt
badcel: :+1:  I think it is fine for now. Those vertical white space problems have s.th. to do with the documentation if it is generated / available or not. It is happening for fields in structs, too (but with one line less in between...).

It's not very nice but probably involves some update to Scriban. I did not take the time to reproduce the exact steps involved in it as the behavior is probably intended but not usefull for our case. So we would need to create some minimal samples and explain why it is bad for us and make some suggestions on how to improve it to the scriban project.

I opened #160 to track this.

badcel:
Great work this is quite big :+1:

badcel (approved):

Inrego:
I just now found this project, and it does indeed seem very interesting. However, still only the few first plugins seem to be implemented. Are there any plans to support more plugins, and if yes, what's the timeline? From what I gather, you've made a rather impressive code generator, so it sounds like it could be as simple as grabbing the remaining gir-files and generate source based on them?
The source of gir files you've previously linked seems dead, but here's a newer source I've found that seems rather up-to-date:
https://gitlab.freedesktop.org/gstreamer/gir-files-rs

badcel:
Thank you. Currently GStreamer is not a priority for me as there are other things missing.

For GStreamer itself it is possible that there is some manual written code missing. As I don't have any experience with GStreamer this is all a bit blurry for me. If you know your way around GStreamer I'm grateful for any contribution and willing to support new contributors to get the needed code landed. 

Adding new GStreamer bindings requires adding them to our [gir files repository](https://github.com/gircore/gir-files) first. To do this the corresponding files must be added to the [configuration](https://github.com/gircore/gir-files/blob/main/gir_files.txt) and perhaps the CI script a little bit adopted. If this is done the gir files get updated automatically.

The next step would be to use the new gir files in the generator.

If you are interested in certain plugins feel free to add them to the gir files repository like described above.

If you want to see improved GStreamer support adding more sophisticated samples would help to unveil missing bits and pieces.

PR #131, opened 2020-12-16
Build project: Code separation and unit tests

badcel:
This finishes the unit tests for the runner class which handles the targets and their dependencies. It adds some more code separation as no parameters are given from the runner to the task executing objects.


firox263:
Had a brief skim over and looks good so far.

What's the benefit of splitting interfaces like `IGenerator` into `ILibraryGenerator` and `ISampleGenerator` (as opposed to having `IGenerator LibraryGenerator` and `IGenerator SampleGenerator` using the same base)?

badcel:
> What's the benefit of splitting interfaces like IGenerator into ILibraryGenerator and ISampleGenerator (as opposed to having IGenerator LibraryGenerator and IGenerator SampleGenerator using the same base)?

There is no ```ISampleGenerator```. So I assume you mean the ```ISampleBuilder``` and ```ILibrarybuilder```? As both tasks of those are quite easy they are already in one class ```DotnetExecutor```. 

Here comes the general answer which expains my train of thought:

The benefit is exactly what you stated. Things which are independent are in independent interfaces / classes.

The long answer is that the build project needs to do several totally different tasks. Those tasks are:
- generate  libraries (```ILibraryGenerator```)
-  clean all projects (```IProjectCleaner```)
- build all libraries (```ILibraryBuilder```)
- build samples (```ISampleBuilder```)
- ...

If we have one interface per task the builder internally has one object per task which it can call. This makes it easy during runtime or unit testing to just swap out one task and leave the other tasks untouched. 
E.g. if we want to create an alternative implementation of the clean task we just need to implement one class with one method and we don't need to implement any other method. We just call the new class under certain conditions and perhaps even call the old class under certain other conditions. 
Those classes would work completely independent of each other and we would avoid side effects and gain a lot of readability and flexibility as we have different implementations of the same interface in different files.

On the class level the same separation applies. E.g. the generator class does completely different things compared to the cleaner. I explicitly avoided putting everything in one class which implements multiplie interfaces to separate the code to avoide side effects. If I change the cleaner I can be sure that it will not affect the generator. 
The only exception I made is for the dotnet executor class as it is just forwarding the calls. If the code in there gets any more complicated it should probably be refactored into several classes to have a clear separation between the tasks.

firox263:
> So I assume you mean the ISampleBuilder and ILibrarybuilder?

Sorry, yeah. I meant ISampleBuilder and ILibraryBuilder.

I don't think I've phrased my point very well. I completely agree with separating out distinct tasks and hiding them behind interfaces. What I'm struggling to see is what the benefit is of having separate `ILibraryBuilder` and `ISampleBuilder` interfaces as opposed to just `IBuilder`.

To me, I see four distinct tasks:
 - Building (`IBuilder`)
 - Generation (`IGenerator`)
 - Cleaning (`ICleaner`)
 - Testing (`ITester`)

I'm not sure there's a significant enough difference between building libraries and samples to justify having two separate interfaces when fundamentally they are the same task (building). It seems to me like an unnecessary distinction. Hopefully this illustrates what I mean:

```cs
// Runner.cs
private void BuildLibraries()
{
    _libraryBuilder.BuildLibraries();
}
```

would become
```cs
// Runner.cs
private void BuildLibraries()
{
    _builder.Build(Projects.GetLibraries() /* or something */);
}
```

I suppose what I'm saying is `DotNetExecutor` should be a very simple wrapper over dotnet, and the burden of differentiating between samples and libraries should be part of the build tool (e.g. `Runner`). This way, the implementation doesn't need to know what 'Libraries' or 'Samples' are, and custom build targets are easy to define.

In future, this means we could also add more fine-grained control over building (e.g. build gtk but not gstreamer, only build webkit2gtk on linux, etc). I think we'll need this as we add more libraries; GStreamer has already doubled generation time.

Hope this makes sense. Either way, I like the direction you're going with here. It's a much more flexible structure than what we were using before.

badcel:
Ah thank you for clarification.

>I'm not sure there's a significant enough difference between building libraries and samples to justify having two separate interfaces when fundamentally they are the same task (building)

 Yes you are right currently there is not much of a differentiation between ```ILibraryBuilder``` and ```ISampleBuilder```. 

You want some interface ```IBuild```. I would argue that we kind of have it already but only in a form of a static class ```DotNet``` ([click](https://github.com/gircore/gir.core/blob/feature/build_unit_tests/Build/DotNet.cs)) It just takes the project and a configuration and builds it.

The intention of the ```ILibraryBuilder``` and ```ISampleBuilder``` is that the runner should not care which projects must be build to build the libraries or the samples. The runner cares about running the targets and calling corresponding objects which just do the work.

If those objects need more information on how to get the work done, they should define their dependencies in their constructor (lke the ```Settings``` object in the constructor of the ```DotNetExecutor```).

Another point is what if a ```Setting``` dictates to build some additional projects during the build target? Than the runner would need to take care and interprete those settings. Perhaps this would even require to pass even more arguments to your suggested ```IBuilder``` interface. From my point of view the ```Runner``` should not care about the details of the different targets.

One plus point for this would be for example that you could call ```dotnetExecutorObject.BuildLibraries()``` and the libraries get build, even without a runner if you can satisfy the constructor of the ```DotNetExecutor```. This could be usefull for example for our dotnet tool in case it should not use the ```Runner``` (because it  heavily depends on the ```Bullseye``` library which our ```DotNetExecutor``` does not).

So in my opinion we kind of have the ```IBuild``` interface and have an _additional_ separtion of concerns in the way that the runner doesn't care what is happening during the build target and we have different classes which can fullfill the build targets independently, which requires an interface per build target.

What do you think?

firox263:
Ah, I see what you're going for. I think the name `DotNetExecutor` confused me, since I sort of assumed it was what `DotNet` actually is.

> Should I add a LibraryBuilder class, a Tester class and a SampleBuilder class instead of the DotNetExecutor?

I like this idea quite a bit. We'd have an object for each toplevel target like you suggested, which we could then extend with finer grain control over what actually gets built.

badcel:
@firox263  I updated the code with the new classes which replace the `DotNetExecutor`. I also introduced a folder `Targets` which contains all the classes which do the target handling.

PR #133, opened 2020-12-17
Feature/compositetemplates

badcel:
This adds simplistic composite template support to the widget class.

 I would like to merge this before continuing with source generators as the PR is already quite huge.


Libs/GObject/Classes/Object.TypeHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+
+                if (ptr == IntPtr.Zero)
+                    ptr = TypeClass.Native.@ref(type.Value);
+
+                return ptr;
+            }
+            
+            internal static ClassInitFunc GetClassInitFunc(IReflect type) => (gClass, classData) =>
+            {
+                Type gtype = GetGTypeFromTypeClass(gClass);
```
firox263, line 25: In the long term, we might want to look at passing the actual corresponding type class rather than the gtype. For now, this looks good to me :+1:

firox263 (commented):
Great work here!

This looks good to me, I've made a few small comments above on things I want to clarify.

I've done a write-up/proposal in #140 with some comments about reflection/source generators and our approach to integrating with GObject (including possible bidirectionality), so please have a look when you have time. I think this is a good starting point for the more language-integrated features we want to implement.

Libs/GObject/Classes/Object.cs
```diff
@@ -235,6 +240,9 @@ protected static void HandleError(IntPtr error)
 
         protected static IntPtr GetHandle(Object obj)
             => obj.Handle;
+        
+        protected static IntPtr GetHandle(Bytes bytes)
```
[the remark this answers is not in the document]
badcel: Yeah, I did not want to make the `Handle` public. This is a little bit of a crux for me. I want to have a nice API meaning in this case no `IntPtr` visible but to achieve this I had to put this very dirty method into `Object`. I'm not satisfied with it either but preferred the nice API here because we tried very hard to hide it beforehand.

I would love to see some other idea which satisfies both needs but I already thought about it quite a lot and see no solution.

I would argue if we make it public for `Bytes` than we allow to leak the native `IntPtr Handle` property into the API. In this case I would remove the `GetHandle(Object obj)` method as well. As the removal would result in cleaner / less code. The API would even be more consistent with it. We probably even should add an `IHandle` interface in this case to make the objects compatible.

Perhaps we should in general decide to go for the cleaner code way and sacrifice a little on the API? If you agree I would introduce the `IHandle` interface and remove both methods.

Libs/GObject/Extensions/AssemblyExtension.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.IO;
+using System.Reflection;
+using GLib;
+
+namespace GObject
+{
+    public static class AssemblyExtension
+    {
+        public static Bytes ReadResource(this Assembly assembly, string resource)
```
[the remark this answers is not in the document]
badcel: Yes this is already in #134 there it is `ReadResourceBytes`. Is this okay for you?

Libs/Gtk3/Classes/Widget.CompositeWidget.cs
```diff
@@ -0,0 +1,90 @@
...
+        
+        protected static void ConnectTemplateSignals(Type gtype, System.Type t)
+        {
+            var classPtr = TypeHelper.GetClassPointer(gtype);
+            
+            IntPtr ptr = Marshal.AllocHGlobal(Marshal.SizeOf(gtype));
+            try
+            {
+                Marshal.StructureToPtr(gtype, ptr, true);
+                WidgetClass.Native.set_connect_func(classPtr, OnConnectEvent, ptr, DestroyConnectData);
```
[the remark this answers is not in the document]
badcel: Yes this is an open topic. I'm going to add a TODO flag.

Libs/GObject/Classes/Object.cs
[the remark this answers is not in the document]
badcel: I added the `IHandle` interface.

Libs/GObject/Extensions/AssemblyExtension.cs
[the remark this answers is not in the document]
firox263: Looks good :+1:

Libs/GObject/Classes/Object.cs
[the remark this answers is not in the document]
firox263: This is a nice solution. Looks good to me.

Libs/GLib/Interfaces/IHandle.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace GLib.Interfaces
```
firox263, line 3: This should just be `GLib`, not `GLib.Interfaces` to be consistent.

Samples/Gtk3/Template/CompositeWidget.cs
```diff
@@ -0,0 +1,34 @@
...
+                template: Assembly.GetExecutingAssembly().ReadResource("CompositeWidget.ui")
+            );
+            BindTemplateChild(gClass, nameof(Button));
+            ConnectTemplateSignals(gClass, type);
+        }
+
+        protected override void Initialize()
+        {
+            InitTemplate();
+            BindTemplateChild(nameof(Button), ref Button);
```
firox263, line 24: What's the difference between `BindTemplateChild` and `BindTemplateChild<T>`? As we use them both for `Button`, they really should have different names.

Maybe `BindTemplateChildForClass` or `ClassBindTemplateChild` for the static one?

Samples/Gtk3/Template/Programm.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
```
firox263, line 1: Typo in the name of this file.

"Program.cs" vs "Programm.cs"

firox263 (commented):
Found a few more stylistic issues. Should be good to merge after these are fixed :+1:

Libs/GLib/Interfaces/IHandle.cs
badcel: Done

Samples/Gtk3/Template/Programm.cs
badcel: Done

Samples/Gtk3/Template/CompositeWidget.cs
badcel: I used different names:

- `ConnectTemplateChildToField` I hope this makes it clear that we "Connect" the template to the field. I think connect is a good name because the attribute to connect fields is the `ConnectAttribute`, too.
- In the `static ClassInit` I only use `BindXX` methods as those work on the native / class level.

What do you think of this names?

firox263: I like this a lot more. Much less ambiguous.

firox263 (approved):

PR #138, opened 2020-12-19
Readme: Promote discussions and good first issues

badcel:


README.md
```diff
@@ -70,7 +70,7 @@ To use the newly build libraries in your project just add a reference to the csp
     $ dotnet add reference [RepoPath]/Libs/Gtk/Gtk.csproj
 
 ## How to help
-Anyone who wants to help is very welcome. Just create a pull request for new code or create an issue to get in contact.
+Anyone who wants to help is very welcome. If you want to get into the project you can just grab some [good first issues](https://github.com/gircore/gir.core/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) or get in contact via a [discussion entry](https://github.com/gircore/gir.core/discussions).
```
firox263, line 73: "Grab" probably isn't the right word here. Maybe something like:

```suggestion
Anyone who wants to help is very welcome. If you want to get into the project take a look at our [Good First Issues](https://github.com/gircore/gir.core/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) or get in contact by starting a [Discussion](https://github.com/gircore/gir.core/discussions).
```

Looks good otherwise. I like the idea of using Discussions as a starting point because it keeps the issue tracker on actionable issues only. Plus it tends to be a 'friendlier' way of getting into a project.

badcel:
Yes this is exactly the idea we should follow along with issues und discussions. We can even convert issues to discussions which should clean up our issues.

PR #149, opened 2020-12-23
Generate the correct library names for macOS.

cameronwhite:
The libraries are named e.g. `libgtk-3.0.dylib` on macOS.

I tested with the GTK libraries installed via homebrew, and this fixes the runtime errors I hit while running the sample GTK application.


badcel:
Thank you very much. Great to know that the generated code works for mac os with this fix :+1:

PR #154, opened 2020-12-24
Feature/build os independent dlls

badcel:
This allows to use the generated DLLs in a cross platform manner. They should work on linux / windows / osx and do not depent on the host which is generating the code.

To solve this there is module initializer which calls the `DllImport.Initialize` method once which detects the os and safes the library to use for all subsequent `DllImport` calls.

- The part of the module calling the `DllImport.Initialize` is in the shared template `module_dll_import.sbntxt`.
- The `DllImport` class is generated once per library in the template `helper_dll_import.sbntxt`


badcel:
@gircore/coreteam This is ready for review.

cameronwhite:
> @cameronwhite could you test this on mac?

Yep, tested this on macOS and it's working fine for me.

badcel:
@cameronwhite  thank you very much ❤️

Generator/Templates/Shared/helper_dll_import.sbntxt
```diff
@@ -0,0 +1,60 @@
...
+        #region Methods
+        
+        public static void Initialize()
+        {
+            NativeLibrary.SetDllImportResolver(typeof(DllImport).Assembly, ImportResolver);
+        }
+        
+        private static IntPtr ImportResolver(string libraryName, Assembly assembly, DllImportSearchPath? searchPath)
+        {
+            if (NativeLibrary.TryLoad(DllName, assembly, DllImportSearchPath.System32, out IntPtr libHandle))
```
na2axl, line 37: This was not working for me until I've changed the `DllImportSearchPath` to `LegacyBehavior` (using Windows)

Generator/Templates/Shared/helper_dll_import.sbntxt
```diff
@@ -0,0 +1,60 @@
...
+                return libHandle;
```
na2axl, line 38: Maybe a way to cache the loaded library, to avoid redundant loading ?

na2axl (commented):
Hi @badcel, I've checked this on Windows using Msys64 to get GObject libraries. Found theses problems

Generator/Templates/Shared/helper_dll_import.sbntxt
badcel: Hm `LegacyBehavior` is working for me, too. I updated the branch to use this value.

@cameronwhite is it working for you, too?

Generator/Templates/Shared/helper_dll_import.sbntxt
badcel: I'm not sure if there is some caching on the side of `NativeLibrary.TryLoad`.

This reads a little bit like it: [docs](https://docs.microsoft.com/en-us/dotnet/core/dependency-loading/loading-unmanaged).

If you agreee I would avoid adding caching for now. If we see problems later or have definitive feedback that we are missing an opportunity here we can add the caching ourself.

@gircore/coreteam what do you think?

Generator/Templates/Shared/helper_dll_import.sbntxt
cameronwhite: Yes, it's still working for me on macOS with the latest changes in this branch

Generator/Templates/Shared/helper_dll_import.sbntxt
badcel: Hm I added a `Console.WriteLine` and the `ImportResolver` method is called several times. So it looks like it is not caching. But I'm not sure if we should add caching and if this could result in some unexpected results.

badcel:
@firox263 If you approve I would like to merge this without any caching (for now) and with the `LegacyBehavior` for `DllImportSearchPath`.

Generator/Templates/Shared/helper_dll_import.sbntxt
na2axl: I think we can check that by printing the returned IntPtr to the console, if the result is the same may be it's finally better to cache ?
But I've no problem to skip this for now.

firox263 (approved):
Tested this on Windows 10 and everything works nicely.

Looks good to me 👍

I've opened #161 to keep track of dll caching

PR #156, opened 2020-12-27
Build: Support  creating packages which get pushed nuget.org automatically

### LINKED ISSUE #155: Nuget Package
It would be awesome to have a nuget package bundled up.

I'm keen to find a start creating c# apps for the Linux desktop/mobile software ecosystem, and I think this would go a long way.

badcel:
This is an attempt to prepare for releasing nuget packages via github.

The idea is that we have a separate `release.yml` pipeline which runs if we create a release on the github page. This pipeline just builds the project and packs the libraries into `/Nuget` folder in the main directory of the project. The version number is taken from the release name and set in the build settings. So the release name should be s.th. like "0.1.1" or "1.2.3-alpha" for example. If it is not a version number parseable by nuget, it is going to throw an exception and fail.

Afterwards `dotnet publish` is called and pushes the packages online. I did not want to include the publishing inside the `Build` project to be sure we are never leaking the nuget api key somewhere or passing it on to some 3rd party library.

I think this is a very simple approach and we can very easily publish releases and kepp full control over version numbers.

What do you think?

@gircore/coreteam 
@archanox

Fixes #155


badcel:
@firox263 if you agree I would merge this and do a test release to see if everything works. I'm going to delete the release on github and the packages on nuget.org afterwards.

firox263 (approved):
Looks good to me.

One thing to be aware of is nuget.org [doesn't allow you to delete packages](https://docs.microsoft.com/en-us/nuget/nuget-org/policies/deleting-packages), only unlist them. Perhaps in that case we should wait to do a nuget release, until we have slightly more stable code. Something to keep in mind. 

Maybe we want to merge this without the changes to the GitHub actions yml?

badcel:
> One thing to be aware of is nuget.org doesn't allow you to delete packages, only unlist them.

Yep already tested this in the beginning of the project there are like ~15 packages unlisted already. :see_no_evil:

As the API key is going to expire in 3 month  I think it is okay to unlist the packages. They are online for like 10 minutes and if we know that it is working afterwards we have something gained and the API key will be valid for one more year.

PR #167, opened 2020-12-30
DllImport: Improvements

### LINKED ISSUE #161: Cache DLLs in DllImport
Investigate whether we can/should cache the resolved DLL between calls to avoid a lookup each time we call a native method.

From #154:

> **na2axl:** Maybe a way to cache the loaded library, to avoid redundant loading ?

> **badcel:** I'm not sure if there is some caching on the side of NativeLibrary.TryLoad.
This reads a little bit like it: docs.
If you agreee I would avoid adding caching for now. If we see problems later or have definitive feedback that we are missing an opportunity here we can add the caching ourself.
gircore/coreteam what do you think?

> **badcel:** Hm I added a Console.WriteLine and the ImportResolver method is called several times. So it looks like it is not caching. But I'm not sure if we should add caching and if this could result in some unexpected results.
Edit: I found no search results which did some caching. Perhaps there are multiple contexts involved and multiple calls are correct?

> **na2axl:** I think we can check that by printing the returned IntPtr to the console, if the result is the same may be it's finally better to cache ?
But I've no problem to skip this for now.

badcel:
- Cache library pointers
- Handle unknown libraries gracefully
- Forward given search path to `NativeLibrary.TryLoad`

Fixes #161 

@gircore/coreteam Please test again under windows as this changes the call of `NativeLibrary.TryLoad` to take the search path from the caller.

### DIFF
--- a/Generator/Templates/Shared/helper_dll_import.sbntxt
+++ b/Generator/Templates/Shared/helper_dll_import.sbntxt
@@ -1,6 +1,10 @@
 using System;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+#nullable enable
 
 namespace {{ namespace }}
 {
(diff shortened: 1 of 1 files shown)


firox263 (approved):
Runs fine on Windows 10.

Nice work!

Issue #78, opened 2020-10-16
Floating reference handling


badcel:
Floating reference handling is probably wrong.

See: https://github.com/GtkSharp/GtkSharp/issues/197 and https://github.com/GtkSharp/GtkSharp/pull/198/commits/b58235b3151b7ba496cd431f5a2c708b34e02af3

badcel:
Information on memory management: https://developer.gnome.org/programming-guidelines/stable/memory-management.html.en

Just some thoughts, there could be false assumptions in them, please correct me if I'm wrong.

Creating our own objects should be correct:
- We don't ref any pure objects (meaning if they are not derived from `InitiallyUnowned`) as if we create a new object ourself they are already owned by us and have a reference count of 1.
- If we create an instance of `InitiallyUnowned` object we sink the ref and take ownership resulting in a reference count of 1

Getting objects:
- TODO

badcel:
I wonder if we use s.th. like `GCH.Root(callback)` and are refing instances if retrieved from C world if this actually is the same like using toggle refs?

badcel:
I asked in #introspection on matrix chat:

> for c# language binding: If we have some GTK program defining a button with an image. If the user now calls gtk_button_get_image we retrieve the C widget and wrap it into a C# object. The method has transfer ownership none attribute. I'm not sure if we should ref the widget instance in this case? If the widget gets destroyed we could unref the instance on C# side afterwards to allow it to be destroyed.
> Can this behaviour be applied in general to reference counted objects with transfer ownership none? (In case of floating references i would sink it).

Zander Brown answered:
> if a function is (transfer none) then you haven't been given a ref, so if you want to keep it around you should ref() it yourself (in which case you must also unref() at some later point to avoid leaking)

I think this confirms the behavior I described above.

PR #174, opened 2021-01-01
Generator: Add ownership transfer to xml documentation

badcel:
Add xml documentation which explains if each parameter of a method should transfer the ownership or not.

This information makes it easy to write manual code to handle the ownership transfer. See  #78 

Return values are going to be added.


badcel:
@gircore/coreteam: Added the return value, this is ready for review.

firox263 (approved):
Looks good to me

PR #177, opened 2021-01-03
Feature/improve marshallig

badcel:
This should fix a lot of memory leaks. It intoduces two helper classes to simplify memory management:

1. StringHelper: Allows to convert a pointer into an ANSI string and automatically free it afterwards. The caller needs to make sure that the IntPtr has transfered it's ownership to us, so we can free it.
2. MarshalHelper: Allows to pass a struct and an action to it. The action is executed and get's a pointer to the struct. After execution the pointer is automatically freed.


badcel:
@gircore/coreteam : This is ready for review

Libs/GLib/Classes/MarshalHelper.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        public static K ToPtrAndFree<T, K>(T structure, Func<IntPtr, K> action)  where T : struct
+        {
+            K result;
+            IntPtr ptr = Marshal.AllocHGlobal(Marshal.SizeOf(structure));
+            try
+            {
+                Marshal.StructureToPtr(structure, ptr, false);
+                result = action(ptr);
+            }
```
firox263, line 16: Since we're putting this in a try block, can we add some kind of logging if an exception is thrown?

e.g. maybe:

```suggestion
            }
            catch (Exception e)
            {
                Console.WriteLine($"Exception thrown while marshalling: {e.Message}");
            }
```

Libs/GLib/Classes/MarshalHelper.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+        
+        public static void ToPtrAndFree<T>(T structure, Action<IntPtr> action) where T : struct
+        {
+            IntPtr ptr = Marshal.AllocHGlobal(Marshal.SizeOf(structure));
+            try
+            {
+                Marshal.StructureToPtr(structure, ptr, false);
+                action(ptr);
+            }
```
firox263, line 32: See above

Libs/Gst/Structs/Structure.cs
```diff
@@ -11,6 +11,7 @@ public partial struct Structure
             IntPtr thisPtr = Marshal.AllocHGlobal(Marshal.SizeOf(this));
             Marshal.StructureToPtr(this, thisPtr, false);
             
+            //Do not free result as ownership is not transered!
```
firox263, line 14: Typo:

```suggestion
            // Do not free result as ownership is not transferred!
```

firox263 (approved):
Looks good to me.

The `ToPtrAndFree` method is a big improvement in terms of freeing memory appropriately. We should enforce this wherever ownership is tranferred.

Just have one point around exceptions and logging, then this is good to merge.

Libs/GLib/Classes/MarshalHelper.cs
badcel: Actually I don't want to catch the exception on our side of the code. But if the user is catching it the memory should be freed. This is why the try block is missing the catch part: 

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/try-finally

Do you think we should catch the exception?

firox263: Ah, is the exception still propagated to the user then? As in, does finally run if the user catches the exception higher up the call stack? The docs don't make that very clear:

> Usually, when an unhandled exception ends an application, whether or not the finally block is run is not important. However, if you have statements in a finally block that must be run even in that situation, one solution is to add a catch block to the try-finally statement.

But then this implies that it does:

> Alternatively, you can catch the exception that might be thrown in the try block of a try-finally statement higher up the call stack. That is, you can catch the exception in the method that calls the method that contains the try-finally statement, or in the method that calls that method, or in any method in the call stack.

badcel: Yep it should be like this, see [stackoverflow](https://stackoverflow.com/questions/9291437/use-a-try-finally-block-without-a-catch-block) which even states that this is similar to what a using block compiles to.

badcel:
@firox263 Thank you for your review, merging :+1:

PR #184, opened 2021-01-08
Feature/fix references

### LINKED ISSUE #78: Floating reference handling

badcel:
Fixes #78 

This adds the concept of *owned refs* to the API. Every wrapped IntPtr must be defined as owned or not. If a pointer is owned it means that the ownership of this part of the memory is transferred to us. We do not have to increase the reference count as we just grab the responsibility. In case this owned reference is floating, we are `Debug.Assert`ing the program as this scenario is not possible.

In case the ownership is not transferred to us this means some other other code owns the reference. In this case we ref the object to make sure that we want to own this reference, too. Meaning as long as our C# object exists the memory can not be freed via an `unref`.

As our objects get notified in case of finalization of a gobject even if we still own a reference we remove the object from our global object array and mark ourself as disposed. 

In case of an object create directly by us via `new ...` we don't need to increase the reference count as `Object.new_with_properties` transfers the ownership to us even in case of an initially unowned object.

Changes
- Supply the `ownedRef` parameter if wrapping a pointer
- The new code fixes a bug which first set the handle to `IntPtr.Zero` and afterwards tried to remove the object via the nulled handle.
- The code for the object constructor with parameters is simplified and removes the double branches
- The code for the object constructor with parameters frees the `Values` passed into it. If we free those, we need to be sure that the caller is not keeping a reference to the `ConstructParameter`s. For this reason I marked the constructor as protected. In this way the derived classes can use ConstructProperties to provde constructors with concrete arguments to the enduser.
- Added nullable information to the xml documentation.
- Added support for wrapping of nullable pointers and adopted our current methods to use the new method if applicable

@firox263 : I changed some GStreamer method signatures which were wrong according to documentation. I assume that this should not break your work or if so at least be "the right" api. Could you verify?

### DIFF
--- a/Libs/GObject/Classes/Object.cs
+++ b/Libs/GObject/Classes/Object.cs
@@ -45,8 +45,8 @@ namespace GObject
         /// Constructs a new object
         /// </summary>
         /// <param name="properties"></param>
-        /// <remarks>This constructor is protected to be sure that there is no caller (enduser) keeping a reference tothe
-        /// construct parameters as the contained values are freed at the end of this constructor.
+        /// <remarks>This constructor is protected to be sure that there is no caller (enduser) keeping a reference to
+        /// the construct parameters as the contained values are freed at the end of this constructor.
         /// If certain constructors are needed they need to be implemented with concrete constructor arguments in
         /// a higher layer.</remarks>
         protected Object(params ConstructParameter[] properties)
--- a/Libs/GObject/Classes/Object.cs
+++ b/Libs/GObject/Classes/Object.cs
@@ -58,11 +58,7 @@ namespace GObject
             var names = new IntPtr[properties.Length];
             var values = new Value[properties.Length];
 
-            IntPtr ptrFirstName = IntPtr.Zero;
             IntPtr handle;
-            
-            if (properties.Length > 0)
-                ptrFirstName = names[0];
 
             try
             {
@@ -74,6 +70,8 @@ namespace GObject
                     names[i] = Marshal.StringToHGlobalAnsi(prop.Name);
                     values[i] = prop.Value;
                 }
+               
+                IntPtr ptrFirstName = properties.Length > 0 ? names[0] : IntPtr.Zero;
 
                 // Create with properties: The new object is owned by us even in case of an object
                 // which inherits GInitially unowned.
@@ -103,12 +101,6 @@ namespace GObject
         /// <param name="ownedRef">Defines if the handle is owned by us. If not owned by us it is refed to keep it around.</param>
         protected Object(IntPtr handle, bool ownedRef)
         {
-            //This is never called if an object is created with new... from C#. It is only called for wrapped pointers.
-            
-            // TODO: Check to make sure the handle matches our
-            // wrapper type.
-            Initialize(handle);
-
             if (!ownedRef)
             {
                 // - Unowned GObjects need to be refed to bind them to this instance
@@ -124,6 +116,10 @@ namespace GObject
 
                 Debug.Assert(!Native.is_floating(handle), "Owned floating references are not possible.");
             }
+            
+            // TODO: Check to make sure the handle matches our
+            // wrapper type.
+            Initialize(handle);
         }
 
         ~Object() => Dispose(false);
@@ -361,7 +357,7 @@ namespace GObject
         /// <param name="ownedRef">Specify if the ref is owned by us, because ownership was transferred.</param>
         /// <typeparam name="T"></typeparam>
         /// <returns><c>true</c> if the handle was wrapped, or <c>false</c> if something went wrong.</returns>
-        public static bool TryWrapHandle<T>(IntPtr handle, bool ownedRef, [NotNullWhen(true)] out T? o)
+        public static bool TryWrapHandle<T>(IntPtr handle, bool ownedRef, [NotNullWhen(true)] out T? o) 
             where T : Object
         {
             o = null;
@@ -406,9 +402,9 @@ namespace GObject
 
         protected virtual void DisposeManagedState()
         {
-            Handle = IntPtr.Zero;
             Objects.Remove(Handle);
-            
+            Handle = IntPtr.Zero;
+
             // TODO: Find out about closure release
             /*foreach(var closure in closures)
                 closure.Dispose();*/
--- a/Libs/GObject/Classes/Object.cs
+++ b/Libs/GObject/Classes/Object.cs
@@ -45,7 +45,11 @@ namespace GObject
         /// Constructs a new object
         /// </summary>
         /// <param name="properties"></param>
-        public Object(params ConstructParameter[] properties)
+        /// <remarks>This constructor is protected to be sure that there is no caller (enduser) keeping a reference tothe
+        /// construct parameters as the contained values are freed at the end of this constructor.
+        /// If certain constructors are needed they need to be implemented with concrete constructor arguments in
+        /// a higher layer.</remarks>
+        protected Object(params ConstructParameter[] properties)
         {
             // This will automatically register our
             // type in the type dictionary. If the type is
--- a/Libs/GObject/Classes/Object.cs
+++ b/Libs/GObject/Classes/Object.cs
@@ -77,14 +77,15 @@ namespace GObject
 
                 IntPtr zero = IntPtr.Zero;
 
-                // Create with properties: The new object is owned by us even in case of an object
-                // which inherits GInitially unowned.
                 handle = Native.new_with_properties(
                     typeId.Value,
                     (uint) properties.Length,
                     ref (properties.Length > 0 ? ref names[0] : ref zero),
                     values
                 );
+
+                if (Native.is_floating(handle))
+                    Native.ref_sink(handle); //Make sure handle is owned by us
             }
             finally
             {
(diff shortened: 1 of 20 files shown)


badcel:
@firox263 I did a short test with Tinyshee and after updating like 4 `WrapHandle` calls in the `PlayerEngine` it did play back a random mp3. So this should work in general.

Libs/Gst/Classes/GhostPad.cs
```diff
@@ -12,7 +12,7 @@ public partial class GhostPad
         // which *must* be run. Therefore our existing constructor design is
         // inadequate.
         public static GhostPad New(string name, Pad target)
-            => WrapHandle<GhostPad>(Native.@new(name, target.Handle));
+            => WrapHandle<GhostPad>(Native.@new(name, target.Handle), false);
```
firox263, line 15: Should this be `WrapNullableHandle` as constructor functions can fail? Although, we assume that object creation always succeeds when using normal constructors, so perhaps not. Thoughts?

Libs/Gst/Structs/Message.cs
```diff
@@ -19,7 +19,7 @@ public partial struct Message
         
         public Gst.Object Src
         {
-            get => GObject.Object.WrapHandle<Gst.Object>(src);
+            get => GObject.Object.WrapHandle<Gst.Object>(src, false);
```
firox263, line 22: This one will need some investigation. It's probably better to drop the `Src` field altogether and just force the user to go through `Message.GetStructure` because we at least have introspection data for it. Let's not block on this for now though.

Libs/Gtk3/Classes/FileChooserButton.cs
```diff
@@ -110,6 +105,6 @@ public FileChooserDialog Dialog
             ConstructParameter.With(DialogProperty, dialog)) {}
 
         public static FileChooserButton New(string title, FileChooserAction action)
-            => new FileChooserButton(Native.@new(title, action));
+            => new FileChooserButton(Native.@new(title, action), false);
```
firox263, line 108: I believe this should be `true`?

[`gtk_file_chooser_button_new`](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkfilechooserbutton.c#L3016) doesn't have introspection attributes (which is probably a documentation issue), but as it's simply a wrapper over `g_object_new`, it should return an owned reference (e.g. `transfer full`).

Libs/Gtk3/Classes/Window.cs
```diff
@@ -913,8 +910,8 @@ public void GetDefaultSize(out int width, out int height)
         public bool GetHideTitlebarWhenMaximized() =>
             Native.get_hide_titlebar_when_maximized(Handle);
 
-        public GdkPixbuf.Pixbuf GetIcon() =>
-            WrapHandle<GdkPixbuf.Pixbuf>(Native.get_icon(Handle));
+        public GdkPixbuf.Pixbuf? GetIcon() =>
+            WrapNullableHandle<GdkPixbuf.Pixbuf>(Native.get_icon(Handle),false);
```
firox263, line 914: Style: add space

```suggestion
            WrapNullableHandle<GdkPixbuf.Pixbuf>(Native.get_icon(Handle), false);
```

Samples/Gtk3/GtkMinimal/DemoWindow.cs
```diff
@@ -14,7 +14,7 @@ public class DemoWindow : ApplicationWindow
         private Notebook notebook;
 
         public DemoWindow(Application application) : this(application, new Builder("demo_window.glade")) { }
-        private DemoWindow(Application application, Builder builder) : base(builder.GetObject("root"))
+        private DemoWindow(Application application, Builder builder) : base(builder.GetObject("root"), false)
```
firox263, line 17: This sample should be updated to use the source generator `[Template(...)]` attribute, but that's for another PR.

firox263 (commented):
Great work! This looks significantly more reliable.

I've made a few comments, only one real issue where the ownership in `FileChooserButton.New()` might be wrong. Other than that, some minor style changes and I think this can be merged 👍

**Edit:** I'm getting some interesting errors when trying to run the demo player. I'll investigate some more, I think there might've been some breakage in creating properties.

firox263:
Found an issue with the changes to the constructor. When running Tinyshee with this PR, I get the following output:

```
(dotnet.exe:12632): GLib-GObject-CRITICAL **: 23:17:41.710: g_param_spec_pool_lookup: assertion 'param_name != NULL' failed

(dotnet.exe:12632): GLib-GObject-CRITICAL **: 23:17:41.710: g_object_new_is_valid_property: object class 'GtkBox' has no property named '(NULL)'
```

From what I can tell, it seems that:
```cs
IntPtr ptrFirstName = (properties.Length > 0) ? names[0] : IntPtr.Zero;

// ...

ref ptrFirstName
```

is not the same as

```cs
ref names[0]
```

Only the first property seems to be getting marshalled when using `ref ptrFirstName`.

I assume this is because the runtime has decided to only marshal the first 64-bits of memory as ref (the size of an IntPtr) rather than (64 * n_properties - the size of the IntPtr array).

I was able to fix this by using `ref names[0]` and ensuring that names always had at least one element. A bit messy, but you get the idea:

```cs
var names = new IntPtr[(properties.Length > 0) ? properties.Length : 1];
names[0] = IntPtr.Zero;

// names[0] gets overwritten with the correct value later
// ...

handle = Native.new_with_properties(
    typeId.Value,
    (uint) properties.Length,
    ref names[0],
    values
);
```

Libs/Gst/Classes/GhostPad.cs
badcel: I think it does not matter a lot currently especially as it is not the constructor but some helper method. Perhaps we need to put a bit more thought in it if we generate methods like this later. 

I updated it, just to match the documented behavior as this is the only argument I currently have.

Libs/Gst/Structs/Message.cs
badcel: Okay :+1: I don't have any opinion here as I do not have any experience with the GStreamer API. But like you said as this is probably fine tuning let's discuss this later. I added a TODO flag to the property.

Libs/Gtk3/Classes/FileChooserButton.cs
badcel: It is not documented in code, which looks like resulting in some default values in the gir file which states `transfer-ownership="none"` which is correct as it is derived from `GInitiallyUnowned `. In this case it looks like it is just a wrapper so you are probably right. I asked this in the #introspection chat. Let's see what is happening.

Libs/Gtk3/Classes/Window.cs
badcel: Done

Samples/Gtk3/GtkMinimal/DemoWindow.cs
badcel: Yes this is for another PR. But we need to decide if we want to have a sample using the `Builder` API. This API does currently not support the `TemplateAttribute` as it is not using composite widgets.

badcel:
Hm this is strange. I verified again (recompiling the gir.core code inside `Tinyshee` and making sure it is the right code) and it is working for me.

I added a little bit different workaround compared to your solution. I think it is not as quirky but I don't know if it is actually the same like yours.

Libs/Gtk3/Classes/FileChooserButton.cs
badcel: The answer that the documentation is correct. g_object_new is not always transfering the full ownership. So my assumption was false but this concrete piece of code is correct.

Libs/GObject/Classes/Object.cs
```diff
@@ -77,29 +74,25 @@ public Object(params ConstructParameter[] properties)
                     names[i] = Marshal.StringToHGlobalAnsi(prop.Name);
                     values[i] = prop.Value;
                 }
+               
+                IntPtr ptrFirstName = properties.Length > 0 ? names[0] : IntPtr.Zero;
```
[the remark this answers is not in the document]
badcel: I created another workaround for your runtime error.

badcel:
@firox263 I added the `ref_sink` call if we are creating our objects with `g_object_new`. I think with this we do everything we can. If it is floating we sink it if not we assume that we own the instance.

This should be good to merge now.

Libs/Gtk3/Classes/FileChooserButton.cs
firox263: I think your solution to `ref_sink` if floating should be enough for now. We can revisit this later if it becomes relevant, but I don't think we'll need much more than that.

firox263:
> I added a little bit different workaround compared to your solution. I think it is not as quirky but I don't know if it is actually the same like yours. Could you verify again under windows?

Works perfectly on Windows now. I like your workaround, makes more sense.

> I added the ref_sink call if we are creating our objects with g_object_new. I think with this we do everything we can. If it is floating we sink it if not we assume that we own the instance.

I think this handles `g_object_new` pretty much completely. I'd be surprised if anything else is needed.

Will go ahead and merge.

PR #187, opened 2021-01-13
GtkMinimal sample: Rework to be more tailored to a gtk builder sample

### LINKED ISSUE #186: Rework GtkMinimal sample
The GTK Minimal sample currently uses the `Builder` API which is not supporting composite templates and therefor no `Template` attribute.

I would suggest that we rename the sample to s.th. like `Builder` and keep it as an example on how to use the builder API.

badcel:
This simplifies the GtkMinimal sample to be a sample demonstrating the builder API.

Fixes #186


badcel:
@na2axl I removed the code regarding property descriptors because I think those should be in a separate sample which is focusing on them and not mixing it with the builder API. Perhaps some markdown document is even better suited to explain the details?

@gircore/coreteam this should be ready to merge.

firox263:
Looks good to me on the whole, although I wonder whether we really want to demonstrate our Builder API with this:

```cs
public DemoWindow(Application application) : this(application, new Builder("demo_window.glade")) { }

private DemoWindow(Application application, Builder builder) : base(builder.GetObject("root"), false)
{
    // ...
}
```

While this technically works, it seems a bit redundant given we support proper composite templates, so you'd always prefer the `[Template(...)]` source generator attribute over this.

Maybe we could instead show how to create a dialog/window entirely with builder and extract objects from it. As in we'd use `builder.GetObject("dialog")` to get a `Gtk.Dialog` object, and then run that, maybe retrieving a few child objects as well to setup signal handlers. Something like this, with a bit more detail:

```cs
void RunDialog()
{
    Dialog dlg = (Dialog)builder.GetObject("dialog");
    Button btn = (Button)builder.GetObject("main_btn");
    btn.OnClicked = (btn, args) => Console.WriteLine("Hello World!");

    dlg.Run();
}
```

I don't have any problems with merging this at the moment, since it's an improvement over `GtkMinimal` at the moment, but it feels a bit like we have two APIs which do the same thing.

Thoughts? @badcel

firox263 (approved):
See above comment. I think this is fine to merge for now, but maybe we want to revisit the Builder sample later to avoid confusion with Composite Templates.

badcel:
Hm in general I agree, to you. But if we have the API available I think we should demonstrate all the use cases. E.g. if someone want's to convert from GTKSharp this would probably be the code they want to check out as it is quite similar there.

So I would say that we can expand on this sample with your dialog addition in a separate PR. Perhaps we should even start with a separate sample to keep things small and understandable.

If you think we should not have competing APIs I wold probably go and delete the `Builder` class completely. But I would have mixed feelings about this. I think users would complain that it is missing so I would like to keep it around.
Builder could be very handy for example for F# as it does not require the user to write the boiler plate code which is generated via a source generator. So the builder API would be the go to way if reflection should be used.

We could add some comment to the Builder sample(s), which informs about the "Composite Templates" and that they probably should be preferred.  What do you think @firox263

firox263:
I think we should look at reducing the scope of `Builder` so it doesn't 'compete' with composite templates. I've opened #202 so we can discuss this further.

Let's merge this now while we still have no merge conflicts.

PR #201, opened 2021-01-21
Feature/safe handles

### LINKED ISSUE #198: Consider use of SafeHandles
If we keep some `IntPtr` code we should use safe handles instead of them.

SafeHandles wrap an IntPtr and allow to use them like managed code.

Some examples which I got during a discussion regarding object disposal:
- https://github.com/Techsola/InstantReplay/blob/main/src/Techsola.InstantReplay/Native/Gdi32.cs#L50
- https://github.com/Techsola/InstantReplay/blob/main/src/Techsola.InstantReplay/Native/Gdi32.GdiObjectSafeHandle.cs
- https://github.com/Techsola/InstantReplay/blob/77862658f3bd5c1a50c048095768150d268bf308/src/Techsola.InstantReplay/Native/Gdi32.cs#L104

If we use safe handles for our handles, we can omit Destructors and the "GC.SupressFinalize" completely as the handles are like managed code which get freed by the SafeHandle class:

- https://github.com/dotnet/runtime/discussions/41493#discussioncomment-58355
```cs
class Foo : IDisposable
{
    private FooSafeHandle thing;

    public void Dispose() => thing.Dispose();
}
```

badcel:
Fixes #198. Followup for #188

I prefer this code a little over the original one as it is a bit cleaner and should be less error prone than using finalizers according to documentation.


firox263:
@badcel Can you rebase this on develop? I've merged #188 so there are a few conflicts in `Object.cs`.

badcel:
This could even be more "invasive" if we use the new generator to make one safe handle per type and use them in the "native" class as it would bring us type safety. 

I'm not sure how equality is implemented on the safe handle class but if it just compares the internal intptr we could replace our public handle property with different implementations of safe handle and everything should just work.

But like i think I would tackle this with the new generator.

badcel:
@firox263 I fixed the conflicts. I think we should stick with this simple solution if you agree to use SafeHandles.

Libs/GLib/Classes/Exception.IDisposable.cs
```diff
@@ -1,30 +1,24 @@
...
+    public partial class GException
     {
-        #region IDisposable Implementation
-
-        private bool _disposedValue;
-
-        ~GException() => Dispose(false);
-
-        public void Dispose()
+        private class GExceptionSafeHandle : SafeHandle
```
firox263, line 8: This file should be renamed from `Exception.IDisposable.cs` to `Exception.SafeHandle.cs`.

Libs/GObject/Classes/Object.ClosureHelper.cs
```diff
@@ -16,12 +16,13 @@ private class ClosureHelper : IDisposable
...
             private readonly ActionRefValues? _callback;
+            private readonly Closure.ClosureSafeHandle _safeHandle; 
             
             #endregion
             
             #region Properties
-            
-            public IntPtr Handle { get; private set; }
+
+            public IntPtr Handle => _safeHandle?.DangerousGetHandle() ?? IntPtr.Zero;
```
firox263, line 25: It might be worth checking if the handle is invalid before returning.

From the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.safehandle.dangerousgethandle?view=net-5.0#System_Runtime_InteropServices_SafeHandle_DangerousGetHandle):

> Using the DangerousGetHandle method can pose security risks because, **if the handle has been marked as invalid with SetHandleAsInvalid, DangerousGetHandle still returns the original**, potentially stale handle value. The returned handle can also be recycled at any point. At best, this means the handle might suddenly stop working. At worst, if the handle or the resource that the handle represents is exposed to untrusted code, this can lead to a recycling security attack on the reused or returned handle.

I think it should be sufficient to return null if the handle is invalid.

Libs/GObject/Classes/Object.cs
```diff
@@ -31,8 +32,8 @@ public partial class Object : IObject, INotifyPropertyChanged, IDisposable, IHan
         #endregion
 
         #region Properties
-        
-        public IntPtr Handle { get; private set; }
+
+        public IntPtr Handle => _safeHandle?.DangerousGetHandle() ?? IntPtr.Zero;
```
firox263, line 36: See above comment about invalid handles.

firox263 (changes requested):
I like this approach quite a lot. It's more understandable than `Dispose` and it opens the door to generating safe handles for GType boxed structs which need custom memory management.

A few safety-related issues around using DangerousGetHandle, but otherwise looks good to me.

Libs/GObject/Classes/Object.ClosureHelper.cs
badcel: Yes I'm going to change it as it is making the intend clearer.

I think it does improve security only marginally. If I understood the docs correctly the root problem is that we should not expose the intptr at all but only the safe handle.

So this PR improves only the memory management aspect . Every user of a GOject could safe it's handle (as it's a struct) and access it after the object was disposed / freed.

I'm going to open a new issue to verify/ improve the security aspect as part of the new generator.

badcel:
@firox263 I updated the code. Should be good to merge.

firox263 (approved):

Issue #100, opened 2020-11-02
Generator rewrite


badcel:
If the generator is rewritten the following features should be thought of:

- Prefixing of interfaces (requires loading of other libraries while one library is generated, to detect if a type is an interface)
- Generation of properties which are defined in an interface (requires loading of a different XML than the one which is currently generated)

firox263:
I have a few ideas for this. I think the main point is to give us access to richer information about GObject types by processing the data beforehand (which is necessary for e.g. prefixing interfaces).

Preferably, we'd split the generator up into a few passes:
 - **Load:** Read in XML file for library and all dependencies
 - **Preprocess:** Resolve aliases, Obtain rich type information->for SignalArgs, GVariant, GValue, interface prefix, etc. By loading the entire tree and preprocessing, we also get dependencies' type information, which we'll need for renaming interfaces.
 - **Validation:** (Optional) Make sure gir data is complete/correct; more for third party libraries
 - **Writing/Generation:** Async generation of all source files (since we have all the data we need in Preprocess, we can simply write in any order. Should speed things up a lot I imagine). We also get the benefit of decoupling Scriban and templates from our logic.

The first three steps would be done synchronously, while the last step could use async/await so we're not waiting on File I/O.

I think you mentioned wanting to make this a `dotnet tool`, and I like this idea a lot. If we rewrite the generator, we should implement this at the same time. Would we also merge the build tool into the generator in that case?

badcel:
> I think you mentioned wanting to make this a dotnet tool, and I like this idea a lot. If we rewrite the generator, we should implement this at the same time. Would we also merge the build tool into the generator in that case?

I think we would not merge the build project. The `dotnet tool` is just able to generate one binding. As we are supporting several libraries we need to call the `dotnet tool` several times. This is probably the job of the Build project. Other 3rd party library authors perhaps want to build their project in a different way, so we just generate source files with the generator and don't force someone that we build their projects, too.

firox263:
Generator wishlist/Adding for future reference:
 - We need to properly implement `in/out/ref` based on the direction annotation from gir
 - PascalCase for struct fields
 - Fix inconsistency where `null` becomes `@null` instead of `Null`.

firox263:
From #127:

[...]

PR #220, opened 2021-01-26
WIP: Staged Generator

mjakeman:
## New Staged Generator
Tracking PR for the generator rewrite. The new generator is being developed in parallel in branch [`feature/generator-v2`](https://github.com/gircore/gir.core/tree/feature/generator-v2). See referenced PRs and commit messages for more information.

---

_Original Post:_

This is a "proof of concept" staged generator like I described in https://github.com/gircore/gir.core/issues/100#issuecomment-721087887. It only generates a subset of the current generated code to focus on the overall architecture instead.

This is a complete rewrite - The git diff isn't too helpful for understanding what's going on here. The source can be found on the branch [`feature/generator-v2`](https://github.com/gircore/gir.core/tree/feature/generator-v2) in the `Generator/` and `Repository/` folders. Note the build tool doesn't compile, so CI will fail.

To try out, run the generator project as an executable, using `Generator/` as your working directory. Some dependencies like `GModule-2.0.gir` are not satisfied, so those need to be manually copied into `gir-files` for the time being. The generator output is currently directed to `Generator/output/*` to avoid clashing with the `Libs/` folder. See `Tool.cs` for a good starting point. An example run might look something like:

```
$ cd Generator
$ dotnet run -- ../gir-files/Gst-1.0.gir
```

### Architecture
The generator is divided into two projects:
 - `Repository` contains loading, parsing, and type dictionary/resolution code. It functions similarly to `libgirepository` and creates a database-style reference of all symbols in the project. 
 - `Generator` meanwhile builds upon `Repository` to create a streamlined code-generation tool.

This separation allows for _significantly_ cleaner generator code, specifically:
 - Using strong references between types, rather than string-based references (See `TypeReference`). This makes renaming and processing far more straightforward.
 - Separate from file format (works around limitations of XmlSerializer), which then enables:
     - Using enums instead of strings for attributes like Direction/Transfer/CalleeAllocates/etc
     - Control over mutability (I use the new `init` accessor extensively to decide what can/cannot be modified during processing).

### Stages
```
Repository
 1. Loader (includes Graph, Serialisation/Parsing, resolves dependencies)
 2. Type Resolution (resolves string-based references into strong references, accounting for namespacing and internal/external types).
---------------------------------------------------------------------
Generator
 3. Processing/Analysis (transforming data, renaming, reparenting, aliasing, etc)
 4. Services (Small non-mutable/static helper objects which do generation logic)
 5. Writer (Asynchronously outputs source files)
```

All five stages are namespaced and reasonably separate. 

A short description of each stage:
 - **Loader** takes in a list of "top-level" projects. It serialises the GIR introspection data and resolves any dependent library GIR files (e.g. Gtk depends on Gdk, Cairo, Pango). Finally, using a very simple graph algorithm (located in `Repository/Graph`), it orders dependencies such that symbols will always be defined before they are used. This is necessary for shared dependencies.
 - **Type Resolution** includes a type dictionary ("Symbol Dictionary") that stores symbols and their names/namespaces. We iterate over all 'String References' and resolve them using the symbol dictionary. This not only ensures the GIR file is valid (all symbols are defined) but allows for renaming by simply changing the property of one single struct instance.
 - **Processing** allows us to transform a subset of data ahead of generation. Prefixing and PascalCase conversions will happen here, as well as "reparenting" (which simply changes the managed name of a type to include the parent class name at the front).
 - **Services** contains small, maintainable **non-mutable** (so they can be used asynchronously) utility classes which implement the generation logic, so as to decouple it from the template backend (e.g. Scriban will only organise the code into source files, the actual generation logic will be handled by the generator using the data from the Analysis stage).
 - **Writer** is a simple Asynchronous wrapper around Scriban which allows source files to be written concurrently (as the introspection data is already finalised in the processing/analysis stage).

### Features
Repository Features:
 - Loading GIR files and their dependencies automatically (recursive load)
 - Serialising GIR files into a raw structure
 - Parsing the raw structure into a nice, readable processed structure (e.g. enums over strings, etc)
 - Directed acyclic graph for ordering logic/avoiding circular dependencies
 - TypeDict (Symbols and Aliases)
 - Resolves strings into strong references, done from the perspective of the current namespace

Generator Features:
 - Printing basic delegate signatures
 - Interface renaming
 - Object reparenting
 - Folder names based on type and version

@badcel @na2axl


badcel:
Great work! :heart: This is quite impressive and like you promised quite a great leap forward.

I just stepped through your code one time (meaning I have only limited overview). I like the general concept of the interpretaton of the GIR data and the split into new structures with the `ManagedName` / `NativeName`.

Just from my "feeling" I think how the code is intertwined is not optimal at the moment. Perhaps it is just some small adjustments, perhaps a little more refactor or even some dependency injection. I need go deep to answer this question. But this will take a bit of time (perhaps even some weeks).

But we should probably work on the general architecture before adding more features. We could try to add architectural improvements via pull requests into your branch.  What do you think?

firox263:
> Just from my "feeling" I think how the code is intertwined is not optimal at the moment. Perhaps it is just some small adjustments, perhaps a little more refactor or even some dependency injection. I need go deep to answer this question. But this will take a bit of time (perhaps even some weeks). A possible solution to this is working on it together step by step.

I agree. The generator I've written here is more of a "Look, you can write a generator with this!" than an actual generator. It was more to figure out the needed features and constraints for the repository side of things.

> But we should probably work on the general architecture before adding more features. We could try to add architectural improvements via pull requests into your branch. I would try to make more smaller PRs to give more frequent feedback and not do s.th. like "V2" after a long time. In this way we can discuss the changes step by step. What do you think?

Yeah, this sounds like a good idea. I think `Repository` is in a good state to iterate on. I don't mind what happens with Generator since it's really just bits and pieces of the last generator copy/pasted over. My preference would be to make this a parallel branch alongside develop and spend the next few months making small PRs into it until we're ready to merge.

mjakeman:
Closed in favour of #274

PR #239, opened 2021-02-18
Feature/generator v2 cleanup

badcel:


Repository/Services/ClassStructResolverService.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Repository.Model;
+
+namespace Repository.Services
+{
+    internal class ClassStructResolverService
```
firox263, line 7: I'd be in favour of moving this to Generator (At least the reparenting part). Otherwise, it feels like the Repository library is gaining too much responsibility over generator-specific decisions.

No reason to block however, we could merge now and change this later.

badcel: I agree with you. We need to find a proper solution for this. Let's clarify in the chat. I think we merge your pr #240 in here and in the meantime we figure out how to solve this.

PR #240, opened 2021-02-19
Integrate Build Tool with new Generator

firox263:
Allows for the generator to output to the `Libs` directory so we can begin compiling libraries again.

We will probably want to refactor/rewrite the build tool at some point, but this should be enough to start getting CI checks.

Edit: CI now fails "successfully" (as in, the build tool works and a CI failure indicates an actual build failure)


Generator/Generator.cs
```diff
@@ -1,25 +1,37 @@
...
+        public static void Write(IEnumerable<string> projects, string outputDir = "output")
         {
             var repository = new Repository.Repository();
             var loadedProjects = repository.Load(FileResolver.ResolveFile, projects).ToList();
 
             Log.Information("Ready to write.");
 
-            var writerService = new Container().Resolve().Value;
-            foreach (LoadedProject proj in loadedProjects)
+            try
```
badcel, line 21: Please don't add exception handling here. This is the task of the caller. Even if it is the build project.

After merging this branch into develop  we should probably remove the build project and  enhance the build pipeline to use the same tool which 3rd party devs would use. This would require to add the different parameters of the build project to the tool which itself then uses the generator.

This means that the top level caller can decide how to handle exceptions and not the generator in between as all non critical exceptions are already handled inside the generator (like detecting var args arguments for example).

badcel (changes requested):
Great work :+1: Just one comment regarding error handling.

badcel (approved):
Great work thank you!

PR #243, opened 2021-03-01
Generate Delegates correctly

mjakeman:
Various fixes to allow delegates to work properly.
 * Implements `Timeout.cs` using the new delegate handlers.
 * Moves processing out of repository and into generator
 * Makes NativeName immutable again

It's a little bit messy at the moment. Generated delegate file names have a 'Native' suffix as I'm modifying the ManagedName property to get things to work (which I don't think we want).


badcel:
We can move processing into the generator. In my last commits I added a new `Name` Property to `Symbol` to allow us to change the `NativeName` and `ManagedName` without loosing any reference.

I think we don't need metadata information if we just want to rename a symbol. 

My interpretation of the names in `Symbol` class:
- The `Name` property contains the original name of the symbol and can not be altered.
- The `NativeName` is the name of the symbol which should be used to make calls into the C libraries. It is part of the `Native` class and should be settable in the `Generator` project.
- The `ManagedName` is the name of the `Symbol` which should be part of the public API and visible to the user. It is settable in the `Generator` project.

PR #272, opened 2021-04-07
Generator Rewrite: Managed Method Generation

mjakeman:
Generates a managed API wrapper around the low level functions. Aiming to cover ~70-80% of the API in this first pass.

Not supported (yet):
 - Static methods (no instance parameters)
 - Async methods
 - Methods with in/out/ref
 - Methods with records as parameters/return
 - Methods with arrays of GObjects as parameters/return
 - Methods with union parameters/return
 - Methods which return delegates

Supported:
 - Everything else

Slowly working towards having GObject compile. Needs a lot of work around marshalling records, as well as freeing in-scope resources. Very messy

**Current State (Does X project compile?):**
 - [X] GLib
 - [X] GObject
 - [X] Gio
 - [X] Cairo
 - [X] Pango
 - [X] GdkPixbuf
 - [x] Atk - ~Strange generation behaviour for native methods~
 - [x] Gdk - ~Needs record support~
 - [x] Gtk - Done
 - [ ] GStreamer, Misc - **N/A**


mjakeman:
Creates compilable, reproducible output for GLib->Gtk. Capable of running the `Window` sample.

I think it's worth merging without GStreamer support (given that GstVideo is broken anyway), so we can use CI to help avoid breakage as we expand support for the excluded method types (particularly we need to support Records and in/out/ref parameters - possibly unions too?).

PR #282, opened 2021-04-20
Run Integration Tests on CI

mjakeman:
An experiment with GitHub Actions


mjakeman:
All tests that don't require system services (like an X11 server) pass without hassle.

If we can separate out the GTK and DBus tests, we can run all the others on CI.

badcel:
We could just add a new test category and a corresponding build target.

 For example "systemtest" which depends on "integrationtest" and contains only the things which require system services.

badcel:
As GitHub supports matrix builds we can build and test the code in parallel for Linux / Mac / windows.

PR #288, opened 2021-04-25
Generator V2: Reflection-based subclassing

mjakeman:
This reintroduces the old reflection-based subclassing that is still used in `develop`. Like develop, it only handles object instantiation and does not support registering interfaces, new gobject properties, etc. Additionally, it doesn't (yet) support overriding the ClassInit function which is needed for templates.

The next steps from here are to figure out what _needs_ to be done at runtime and what lookups could optimised out via source generation (we would ideally support both source gen and fallback reflection using some kind of interface).


Libs/GdkPixbuf-2.0/Classes/Pixbuf.cs
```diff
@@ -14,8 +14,11 @@ public partial class Pixbuf
 
         public static Pixbuf NewFromFile(string fileName)
         {
-            IntPtr handle = Native.Pixbuf.Instance.Methods.NewFromFile(fileName, out var error);
+            var error = new GLib.Native.Error.Handle(IntPtr.Zero);
```
badcel, line 17: Why do we need to define the error beforehand? The SafeHandle should free the memory correctly?

I did some tests and it looks like the error handle is not freed (immediately) if the `GException` saves the error safe handle. But if we do not hold on to it, the reference it is freed immediately. I wonder if the garbage collector would collect the safe handle and thus free the error at a later point in time?

If possible I want to keep the original code. The big point for safe handle is that it removes the need to care about the unmanaged memory freeing (e.g. `error.Dispose()`).

Libs/GdkPixbuf-2.0/Classes/PixbufLoader.cs
```diff
@@ -13,11 +13,13 @@ public static Pixbuf FromBytes(byte[] data)
             {
                 using var bytes = Bytes.From(data);
 
-                Native.PixbufLoader.Instance.Methods.WriteBytes(handle, bytes.Handle, out var error);
+                var error = new GLib.Native.Error.Handle(IntPtr.Zero);
```
badcel, line 16: See comment regarding error safe handle.

Libs/Gio-2.0/Classes/DBusConnection.cs
```diff
@@ -16,8 +16,10 @@ public partial class DBusConnection
 
         public static DBusConnection Get(BusType busType)
         {
-            var handle = Native.Functions.BusGetSync(busType, IntPtr.Zero, out var error);
+            var error = new GLib.Native.Error.Handle(IntPtr.Zero);
```
badcel, line 19: See comment regarding error safe handle.

Libs/Gio-2.0/Classes/DBusConnection.cs
```diff
@@ -34,7 +36,8 @@ public static DBusConnection Get(BusType busType)
             void Callback(GObject.Object sourceObject, AsyncResult res)
             {
                 // TODO: Make sure this is correct (can we assume res is a GObject?)
-                var ret = Native.DBusConnection.Instance.Methods.CallFinish(sourceObject.Handle, (res as GObject.Object).Handle, out var error);
+                var error = new GLib.Native.Error.Handle(IntPtr.Zero);
```
badcel, line 39: See comment regarding error safe handle.

Libs/Gio-2.0/Classes/DBusConnection.cs
```diff
@@ -51,7 +54,8 @@ void Callback(GObject.Object sourceObject, AsyncResult res)
         public Variant Call(string busName, string objectPath, string interfaceName, string methodName, Variant? parameters = null)
         {
             var parameterHandle = parameters?.Handle ?? GLib.Native.Variant.Handle.Null;
-            var ret = Native.DBusConnection.Instance.Methods.CallSync(Handle, busName, objectPath, interfaceName, methodName, parameterHandle, GLib.Native.VariantType.Handle.Null, DBusCallFlags.None, 9999, IntPtr.Zero, out var error);
+            var error = new GLib.Native.Error.Handle(IntPtr.Zero);
```
badcel, line 57: See comment regarding error safe handle.

Generator/Templates/native.safehandle.sbntxt
```diff
@@ -46,7 +46,7 @@ namespace {{ namespace.native_name }}
                 else
                 {
                     var str = new {{symbol_name}}();
-                    Marshal.StructureToPtr(str, ptr, true);
+                    Marshal.StructureToPtr(str, ptr, false);
```
badcel, line 49: Did this change fix some error for you?

Libs/GLib-2.0/Classes/Filename.cs
```diff
@@ -8,17 +8,21 @@ public static class Filename
     {
         public static string ToUri(string filename, string hostname)
         {
-            var uri = Native.Functions.FilenameToUri(filename, hostname, out Native.Error.Handle error);
+            var error = new Native.Error.Handle(IntPtr.Zero);
```
badcel, line 11: See comment regarding error safe handle.

Libs/GLib-2.0/Classes/Filename.cs
```diff
@@ -8,17 +8,21 @@ public static class Filename
...
             Error.ThrowOnError(error);
+            error.Dispose();
 
             return uri;
         }
 
         public static string FromUri(string uri, out string? hostname)
         {
-            var fileName = Native.Functions.FilenameFromUri(uri, out hostname, out Native.Error.Handle error);
+            var error = new Native.Error.Handle(IntPtr.Zero);
```
badcel, line 22: See comment regarding error safe handle.

Libs/Gst-1.0/Classes/Element.cs
```diff
@@ -46,9 +46,11 @@ public partial class Element
 
         public static Element MakeFromUri(URIType type, string uri, string elementName)
         {
-            IntPtr result = Native.Element.Instance.Methods.MakeFromUri(type, uri, elementName, out var error);
+            var error = new GLib.Native.Error.Handle(IntPtr.Zero);
```
badcel, line 49: See comment regarding error safe handle.

Libs/Gst-1.0/Classes/Parse.cs
```diff
@@ -13,16 +13,20 @@ public static class Parse
         /// <exception cref="GLib.GException">Throws an exception in case of an error</exception>
         public static Element Launch(string pipelineDescription)
         {
-            IntPtr ret = Native.Functions.ParseLaunch(pipelineDescription, out var error);
+            var error = new GLib.Native.Error.Handle(IntPtr.Zero);
```
badcel, line 16: See comment regarding error safe handle.

Libs/Gst-1.0/Classes/Parse.cs
```diff
@@ -13,16 +13,20 @@ public static class Parse
...
             GLib.Error.ThrowOnError(error);
+            error.Dispose();
 
             return ObjectWrapper.WrapHandle<Element>(ret, false);
         }
 
         public static Element BinFromDescription(string binDescription, bool ghostUnlinkedPads)
         {
-            IntPtr ret = Native.Functions.ParseBinFromDescription(binDescription, ghostUnlinkedPads, out var error);
+            var error = new GLib.Native.Error.Handle(IntPtr.Zero);
```
badcel, line 26: See comment regarding error safe handle.

Libs/GObject-2.0/Native/Classes/TypeDictionary.Registration.cs
```diff
@@ -0,0 +1,95 @@
+﻿using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace GObject.Native
+{
+    public static partial class TypeDictionary
+    {
+        // TODO: This stuff needs to be hidden behind a ITypeInitialiser interface
```
badcel, line 9: I think we are starting to leave behind our prototype phase as we already know that we will get the code working, as it is working already in the develop branch. This is why I want to start with code as clean as possible now to avoid another rewrite not to far in the future.

Can you please cleanly separate the lazy loading and the TypeDictionary in different classes as both are different things which should not influence each other. 

One possible idea is that the TypeDictionary itself gets to be a static wrapper around some object which can resolve types. And then we can have different implementations of this object. One beeing lazy and another one expecting everything to be present at program startup. 
This would mean we actually have two different type dictionaries and the static wrapper decides which one to use.

Then we will probably see, that both implementations share some common code which we can refactor into the TypeDictionary itself to get the essential interface for "different kinds of loading types".

badcel (changes requested):
I think we have to separate the different code concerns as we are leaving prototyping phase. This is why I like to have this cleaned up before merging.

Libs/GdkPixbuf-2.0/Classes/Pixbuf.cs
mjakeman: We had quite a lot of trouble with `SafeHandle` + direction attributes when getting `TypeQuery` to work (`AccessViolationException`). The solution was to not generate direction attributes for SafeHandle-type parameters (i.e. records). The downside is this means that error handling can't use `out` anymore as GError is a record.

The calls to Dispose can be removed (in fact, they should happen automatically when the variable leaves the block IIRC).

Generator/Templates/native.safehandle.sbntxt
mjakeman: Changing this (in addition to stopping usage of `ref safehandle`) fixed getting `AccessViolationException` and other heap corruption related problems.

Reading the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.structuretoptr?view=net-5.0), I get the impression it should indeed be false:

> **fDeleteOld**
> true to call the DestroyStructure(IntPtr, Type) method on the ptr parameter before this method copies the data. The block must contain valid data. Note that passing false when the memory block already contains data can lead to a memory leak.

As we create the (blank) pointer above with `AllocHGlobal`.

Libs/GdkPixbuf-2.0/Classes/Pixbuf.cs
badcel: Did the problem regarding SafeHandles occur only with the error argument or with other arguments, too?

I reread the documentation regarding error handling: https://developer.gnome.org/glib/2.64/glib-Error-Reporting.html

I read it like it is correct to pass a pointer. But I assume that if we mark a safe handle as `out` the marshaller is implicitly creating a pointer because C does not know of `out` parameters and then returning the result into our managed world. This can be the reason why it is working. So it could be correct to do it in this way as in some cases it is more "idiomatic" like in the error case.

Parameters are described like this:

```xml
<parameter name="padding_top" direction="out" caller-allocates="0" transfer-ownership="full" optional="1" allow-none="1">
```

What happens if we use `out` parameters only if direction is `out` and `caller-allocates` is 0? In case `caller-allocates` is 1 we just generate a regular parameter.

Just for reference, the error parameter is generated here as `out` and `caller-allocates` = 0
https://github.com/gircore/gir.core/blob/50d797b6e01c1caec9304fd03833ace359626598/Repository/Factories/Model/ParameterListFactory.cs#L37

Can you verify this? If this works out I think our native code could be quite a bit cleaner.

Libs/GObject-2.0/Native/Classes/TypeDictionary.Registration.cs
badcel: The code is not really separate. E.g.: `RegisterSubclass` is part of `TypeRegistrar` but in case of "preregistered" subclasses there is no need for the `TypeRegistrar` to have this method.

If we have some kind of fallback mechanic it is not the task of the type dictionary to do this. The task of the type dictionary is to provide types. If a type is not found it is the task of the caller to decide what to do. Perhaps we need to add another layer on top of the type dictionary.

I try to adhere to the SOLID principles and I think the type dictionary is not following the single responsibility principle.

I'm still in the phase of internalize those principles but I think they help a lot creating better code.

If you prefer we can merge the code but it probably can not stay like it is now. On the other hand I think it is a good way to practice "clean code" if we work out the architecture for this together in this PR. What do you think?

Generator/Templates/native.safehandle.sbntxt
badcel: Isn't it the other way around? We create a pointer to some memory location. What if this memory location still holds old data which was not cleared. But I could be wrong about this.

Perhaps it is working again with my other comment regarding `caller-allocates`?

We probably should understand what is the correct way of doing it and why, add a comment and be done with it.

One of us could ask on gitter in #csharplang. I made very good experiences there. The people are quite helpful.

badcel (approved):

PR #299, opened 2021-04-29
WrapHandle + Support `IHandle` on interfaces

mjakeman:
- Generates interfaces with support for `IHandle`.
- Moves to the new WrapHandle method in `ObjectWrapper` and removes the old one.
- Fixes type lookup
- Adds various constructors as a temporary measure until we can generate them


Libs/GObject-2.0/Native/Classes/ObjectWrapper.cs
```diff
@@ -27,7 +33,14 @@ public static class ObjectWrapper
             // We need to find the "true type" of the object so that the user
             // can upcast/downcast as necessary. Retrieve the gtype from the
             // object's class struct.
-            Type gtype = GetTypeFromHandle(handle);
+
+            Type gtype = GetTypeFromInstance(handle);
+
+            Debug.Assert(
```
badcel, line 39: I think we do not need this assert as the `GetTypeFromInstance` method retrievs the type from the handle. The old implementation was just a bug. Otherwise we could write debug asserts for all kind of in between steps.

Libs/GObject-2.0/Native/Classes/ObjectWrapper.cs
```diff
@@ -39,11 +52,16 @@ public static class ObjectWrapper
...
-        private static Type GetTypeFromHandle(IntPtr handle)
+        private static Type GetTypeFromInstance(IntPtr handle)
         {
             TypeInstance.Struct instance = Marshal.PtrToStructure<TypeInstance.Struct>(handle);
-            TypeClass.Struct classStruct = Marshal.PtrToStructure<TypeClass.Struct>(instance.GClass);
-            return new Type(classStruct.GType);
+            TypeClass.Struct klass = Marshal.PtrToStructure<TypeClass.Struct>(instance.GClass);
+            var typeid = klass.GType;
+
+            if (typeid == 0)
```
badcel, line 61: In which case do you think the type id could be zero? Would this even be an error for this method as a type id of 0 is valid for the `Type` struct.

If the check is necessary I would suggest checking for `Types.Invalid` as it explains the intend better.

Libs/GObject-2.0/Native/Classes/TypeDictionary.cs
```diff
@@ -75,9 +75,12 @@ internal static System.Type GetSystemType(Type gtype)
             // tree until we find a type that is. As all objects are
             // descended from GObject, we will eventually find a parent
             // type that is registered.
-            
+
             while (!_reverseTypeDict.TryGetValue(gtype, out sysType))
+            {
                 gtype = new Type(Functions.TypeParent(gtype.Value));
+                if (gtype.Value == 0) throw new Exception("Could not retrieve parent type - is the typeid valid?");
```
badcel, line 82: We could add a method to the `Type` struct so this reads:

```cs
if (gtype.IsInvalid()) throw new Exception("Could not retrieve parent type - is the typeid valid?");
```

I think this makes the intend clearer.

Libs/Gtk-3.0/Classes/CssProvider.cs
```diff
@@ -2,37 +2,32 @@
...
+using GObject.Native;
 
 namespace Gtk
 {
-    /*
     public partial class CssProvider
     {
-        public CssProvider() { }
+        public static CssProvider New()
+            => ObjectWrapper.WrapHandle<CssProvider>(Native.CssProvider.Instance.Methods.New(), false);
```
badcel, line 14: The wrapper checks if an object is already known for this pointer and afterwards calls the constructor per reflection. But as we are creating a new instance we can omit this check as we now there is no other instance and direcly call a constructor.

Libs/Gtk-3.0/Classes/Expander.cs
```diff
@@ -0,0 +1,10 @@
+﻿using GObject.Native;
+
+namespace Gtk
+{
+    public partial class Expander
+    {
+        public static Expander New(string text)
+            => ObjectWrapper.WrapHandle<Expander>(Native.Expander.Instance.Methods.New(text), false);
```
badcel, line 8: The wrapper checks if an object is already known for this pointer and afterwards calls the constructor per reflection. But as we are creating a new instance we can omit this check as we now there is no other instance and direcly call a constructor.

Libs/Gtk-3.0/Classes/Image.cs
```diff
@@ -0,0 +1,13 @@
+﻿using GObject.Native;
+
+namespace Gtk
+{
+    public partial class Image
+    {
+        public static Image New()
+            => ObjectWrapper.WrapHandle<Image>(Native.Image.Instance.Methods.New(), false);
```
badcel, line 8: The wrapper checks if an object is already known for this pointer and afterwards calls the constructor per reflection. But as we are creating a new instance we can omit this check as we now there is no other instance and direcly call a constructor.

Libs/Gtk-3.0/Classes/ScrolledWindow.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using GObject.Native;
+
+namespace Gtk
+{
+    public partial class ScrolledWindow
+    {
+        public static ScrolledWindow New(Adjustment? hAdjustment = null, Adjustment? vAdjustment = null)
+        {
+            IntPtr ptr = Native.ScrolledWindow.Instance.Methods.New(
```
badcel, line 10: The wrapper checks if an object is already known for this pointer and afterwards calls the constructor per reflection. But as we are creating a new instance we can omit this check as we now there is no other instance and direcly call a constructor.

Libs/Gtk-3.0/Classes/TextView.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using GObject.Native;
+
+namespace Gtk
+{
+    public partial class TextView
+    {
+        public static TextView New()
+            => ObjectWrapper.WrapHandle<TextView>(Native.TextView.Instance.Methods.New(), false);
```
badcel, line 9: The wrapper checks if an object is already known for this pointer and afterwards calls the constructor per reflection. But as we are creating a new instance we can omit this check as we now there is no other instance and direcly call a constructor.

Libs/GObject-2.0/Native/Classes/ObjectWrapper.cs
mjakeman: I like having Debug asserts as it helps us ensure that the type integration's state is exactly how we expect it to be. At the same time, unlike exceptions it won't degrade performance on a release build which is another benefit.

I'd be in favour of writing more debug asserts in the type integration part of GObject so we don't have any surprises.

Libs/GObject-2.0/Native/Classes/ObjectWrapper.cs
mjakeman: I think I got confused between `g_type_parent` returning zero and type itself being zero. In hindsight, this doesn't make sense - I'll remove it.

Libs/GObject-2.0/Native/Classes/TypeDictionary.cs
mjakeman: Looking at it again, if TypeParent returns zero, it simply means the call failed. So instead of wrapping it in a `Type`, maybe it should read something like this:

```cs
var result = Functions.TypeParent(gtype.Value);
if (result == 0) throw new GTypeLookupException();
gtype = new Type(result);
```

Libs/Gtk-3.0/Classes/CssProvider.cs
mjakeman: How would this work? Would we just add a `WrapNewHandle<>()` that bypasses all the checks?

Libs/GObject-2.0/Native/Classes/ObjectWrapper.cs
badcel: Okay, we can keep it as it is not hurting like you said.

Libs/GObject-2.0/Native/Classes/TypeDictionary.cs
badcel: Yes I like your 2nd thought a litte better than the first. In general I think both approaches are valid.

Nevertheless I don't want any "magic values" inside the code, even if it is the number zero as it does not express the intend and makes it hard to look for all places where we are interested in certain basic types.

In addition to my first suggestion I have some more:

 ```cs
var result = Functions.TypeParent(gtype.Value);
if (result == (nuint) BasicTypes.Invalid) throw new GTypeLookupException();
gtype = new Type(result);
```

 ```cs
//AsNuint is an extension method on the enum.
var result = Functions.TypeParent(gtype.Value);
if (result == BasicTypes.Invalid.AsNuint()) throw new GTypeLookupException(); 
gtype = new Type(result);
```

What do you think? Do you have another suggestion?

Libs/Gtk-3.0/Classes/CssProvider.cs
badcel: It would be as easy as this:

```cs
public static CssProvider New()
    => new CssProvider(Native.CssProvider.Instance.Methods.New(), false);
```

Libs/GObject-2.0/Native/Classes/TypeDictionary.cs
mjakeman: We've decided on moving `BasicTypes` to Native and then using them in `Type`'s static values

PR #303, opened 2021-05-02
Build: Support async doc generation

badcel:


mjakeman (approved):
Looks good to me. At some point, we should decide if we want to use the `Parallel` or `async/await` API (as they each have their own benefits/tradeoffs). For now, as this is consistent with our existing usage, let's merge.

PR #310, opened 2021-05-06
Update README.md

mjakeman:
Noticed the README still contains some references to pre-Generator V2, where we e.g. didn't generate managed methods, etc.

I've updated a few parts and given the supported libraries section a facelift by adding a table.

Thoughts?


badcel:
Can you add a target text for system integration?

badcel (approved):

badcel (approved):

PR #318, opened 2021-05-12
Start refactoring repository project

badcel:


badcel:
The window sample still outputs some critical errors during runtime. The rest of the samples seem to work fine.

badcel:
As the errors are already present in the current develop branch I would like to merge this as it improves the file structure.

We can improve / fine tune the structure in a separate PR in a next step.

GirLoader/Output/Model/Alias.cs
```diff
@@ -1,16 +1,16 @@
 ﻿using System.Collections.Generic;
 
-namespace Repository.Model
+namespace GirLoader.Output.Model
 {
     public class Alias : Element
     {
-        public Namespace Namespace { get; }
+        public Repository Repository { get; }
```
mjakeman, line 7: I'm not a fan of using Repository and Namespace interchangeably like this. I think it's an unnecessary indirection, given that the Generator will only ever use the Namespace field. I would like to keep using Namespace here like before.

GirLoader/Output/Model/NamespaceFactory.cs
```diff
@@ -0,0 +1,125 @@
...
+        private readonly AliasFactory _aliasFactory;
+        private readonly CallbackFactory _callbackFactory;
+        private readonly EnumerationFactory _enumartionFactory;
+        private readonly InterfaceFactory _interfaceFactory;
+        private readonly RecordFactory _recordFactory;
+        private readonly MethodFactory _methodFactory;
+        private readonly ConstantFactory _constantFactory;
+        private readonly UnionFactory _unionFactory;
+
+        public NamespaceFactory(ClassFactory classFactory, AliasFactory aliasFactory, CallbackFactory callbackFactory, EnumerationFactory enumartionFactory, InterfaceFactory interfaceFactory, RecordFactory recordFactory, MethodFactory methodFactory, ConstantFactory constantFactory, UnionFactory unionFactory)
```
mjakeman, line 18: typo: `enumartionFactory` -> `enumerationFactory`

mjakeman (changes requested):
I think this is a good start. It's an improvement in terms of getting rid of cruft, but I have a few comments/issues with readability:
 - I don't think it makes sense to swap Namespace with Repository. They aren't really the same concept, and it makes the Generator code more verbose with little benefit.
 - I'm not really a fan of `GirLoader` as the project name. It describes what the library does, but doesn't make much sense in terms of it being a representation of introspection data.  Maybe `Introspection` could be a better name? e.g. `GirLoader.Output.Model.Record` vs `Introspection.Model.Record`.
 - Having `Input.Model` and `Output.Model` is nice in terms of symmetry but it reduces readability. It's not immediately clear what type is being referred to when the namespace is not qualified.
 - Similarly, I would like to avoid having multiple `Loader.cs` files.
 - I think Loading, Transforming, and Resolving are too interconnected. My preference would be to split out Model and have Input and Output interface only by the contents of Model (i.e. no reference to Input from Output).

I'm not sure how much is worth addressing in this PR. We can change/tweak most of this in later PRs. I would like to revert the Namespace->Repository change before merging though.

badcel:
Yes I think we should merge as it is now and improve later in little steps. I think this makes like the 90% step into the right direction. The rest is really fine tuning and not very urgent and probably needs a lot of discussion. 

I would even consider deferring the discussion at a later point in time as I would like to concentrate on bringing the `Generator` forward. I did not expect the refactoring of the `Repository` getting this huge. So the `Generator` refactoring will probably be much more work.

In case you want to improve the structure yourself I would suggest you open a new PR and we discuss the topics there. But I would recommend to not mix different topics like name of the project file with for example the file hierarchy as we could have different opinions and sorting it all out in one step could be more difficult (except the topic are dependent on each other from your point of view).

My opinions regarding your remarks: 

> I don't think it makes sense to swap Namespace with Repository. They aren't really the same concept, and it makes the Generator code more verbose with little benefit.

I did not do the changes from the view point of the `Generator` project, but from the view point of what is right in my opinion from the `Repository` projects view. I did the change for the same reason you bring forward: `Namespace` and `Repository` are different concepts. A `Namespace` does not define any dependencies. Instead a `Repository` defines the dependencies of all the `Namespaces` it is containing. As we require the dependencies on the `Type` level there are two possibilities to solve this:
 * Switch the `Namespace` property for the `Repository` property
 * Introduce a new a type which just holds a list of the dependencies.

I decided to not use a new Type because if we for example would need another information from the repository we would need to transfer it to the new type, too. So I'm left with the first solution.
In general I think the best solution would be to have no dependency from the `Type` level into higher levels (neither `Namespace` nor `Repository`).
But as this requires a lot more brain thought and concepts I went with the "simple" solution. Any idea into this direction is more than welcome. _<5 minutes later>_ Hmmm, while thinking about it I checked the `Method` generation and it uses the "current namespace" which is available during generation. Perhaps we can really get rid of the reference into higher levels completely if we do it in the same way? But this would be another PR, too.

> I'm not really a fan of GirLoader as the project name. It describes what the library does, but doesn't make much sense in terms of it being a representation of introspection data. Maybe Introspection could be a better name? e.g. GirLoader.Output.Model.Record vs Introspection.Model.Record.

I'm open to improvement here but for me the "introspection data" is part of the "gir" = "GObject Introspection". I think if we omit something like "Loader" it is not clear for a 3rd party developer what this is for and digging through code would be necessary to understand what is happening especially in our case as we have lots of projects in the git repository.

> Having Input.Model and Output.Model is nice in terms of symmetry but it reduces readability. It's not immediately clear what type is being referred to when the namespace is not qualified.

This is why I would qualify the name if there is any risk of confusing them, and at the same time avoiding duplicated potentially missleading information. It is quite common to have conflicting names which are only separated by the namespace. Consider Frameworks like Xamarin Forms. There is the Gtk.Button, the Xamarin button, the WPF button and so on. They need to get everything handled somehow but they don't prefix / suffix their classes. It does not add any value.

>Similarly, I would like to avoid having multiple Loader.cs files.

See above.

> I think Loading, Transforming, and Resolving are too interconnected. My preference would be to split out Model and have Input and Output interface only by the contents of Model (i.e. no reference to Input from Output).

Yeah like one year ago I would agree without hesitation. I think it is very hard to find the line between "overengeneering" and "future proof work which is as easy as possible". For me separating this is not needed. The model classes do not have any relevant logic. They are just there to help other classes do their work. The important logic is in the "Loader" / "Resolver" things. Thus I think the focus should be on those. If the model would be needed separately by some other project this would perhaps qualify to move it into a top level folder. But in the same moment I think it should be moved into a separate project file to stand for itself. So currently it is a bit: [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

I think you are already aware that some of my opinions are not the idiomatic C# ones (like not suffixing class with the namespace they live in, interfaces which are not starting with an "I", or using a little more uncommon libs like MSTest, Serilog and so on). I don't want to go the idiomatic way but the way which makes sense from a data point of view and test things out to improve on the standard:
 * E.g. GObject does not prefix their interfaces with an "I". 
 * E.g. Vala is not adding suffixes to their classes.

Ultimatively I think this is from a pure technical language independent point of view the right decision. This is why I try to test thinkgs out as I want to see if it is benefical to do so with C#. If it is not we can always change but learned a lesson for our professional live.

So let's merge? :wink:

PR #322, opened 2021-05-19
Delegate Marshalling

mjakeman:
Reorganises delegate generation into a new `DelegateGenerator.cs` class, reusing method generation code where possible. Uses the standard Native<->Managed conversion logic from `Convert.cs` over whatever custom logic we had before.

~Currently only GLib builds. Some string-related problems are breaking method generation for GObject and up.~

This is part 1/2 for fully supporting signals again (custom SignalArgs generation is also broken).

~Note, I've temporarily disabled GTK/GStreamer to speed up iteration - maybe we should add a setting for this?~

**Edit:** Now working with all projects enabled under CI.


Generator/DelegateGenerator.cs
```diff
@@ -0,0 +1,126 @@
...
+            // We only support a subset of delegates at the
+            // moment. Determine if we can generate based on
+            // the following criteria:
+
+            // No in/out/ref parameters
+            if (parameterList.SingleParameters.Any(param => param.Direction != Direction.Default))
+                return false;
+
+            // No union parameters
+            if (parameterList.SingleParameters.Any(param =>
```
badcel, line 99: Can be shortened to:

```cs
if (parameterList.SingleParameters.Any(param =>
    param.TypeReference.GetResolvedType() is Union))
    return false;
```

PR #323, opened 2021-05-19
Signals, Cairo, Events + General Cleanup

mjakeman:
This restores our old support for signals now that delegates work as expected.

This is technically done, but needs some further testing.

It also includes:
- Basic implementation of Cairo
- Support for GdkEvent
- A complex sample demonstrating custom drawing code
- Fixes to union and bitfield generation

---

TODO (before merge):
 - [x] `Value.Extract` must work for arbitrary records
 - [x] Fix bitfield size (long vs uint vs int?)
 - [ ] Fix issue where DrawingArea occasionally stops drawing on low memory (GC?)
 - [X] Handle unions (e.g. GdkEvent / GdkEventKey)
     - [X] Special case Gdk.Event ~/ Gst.MiniObject / etc~ (MiniObject is for another PR)
     - (funnily enough, Gst.MiniObject is the example of [what not to do](https://gi.readthedocs.io/en/latest/writingbindableapis.html) when writing bindable APIs)
 - [X] We need a solution for cross-platform cairo (handwritten bindings)
 - [X] ~Tests~ (Next PR: Unit Test all of GLib/GObject)
 - [x] Port a subset of [text editor](https://github.com/mjakeman/bluetype/tree/wip/mjake/generator-compat) as a sample project
     - [x] Either fix deletion or remove it from the text editor demo (currently poorly implemented, prone to crashing)


badcel:
We need to check the documentation generation as the old union generation behavior is restored which could break the documentation again.

mjakeman:
Had this error when pressing a button after the drawing area stopped drawing:

> (dotnet.exe:17948): GLib-GObject-CRITICAL **: 18:25:20.068: g_object_remove_toggle_ref: assertion 'G_IS_OBJECT (object)' failed

So maybe it isn't signal related and in fact a sign some underlying object is being freed.

badcel:
To fix the error it looks like in the class "PrimitiveValueType.cs" the order of lines 15+16 must be swaped: https://github.com/gircore/gir.core/blob/cde3d7b3f90fdbc41c7061c57b6fbdf607eebb4d/GirLoader/Output/Model/PrimitiveValueType.cs#L15

Afterwards there are some follow up errors which must probably be resolved.

Build/Properties/launchSettings.json
```diff
@@ -0,0 +1,9 @@
+{
```
badcel, line 1: I think we can remove this file for now. I don't know how to generate universal usable "launchSettings"

Generator/Convert.cs
```diff
@@ -71,15 +71,15 @@ internal static string NativeToManaged(TransferableAnyType transferable, string
                 { Type: String } when (transfer == Transfer.None) && (transferable is ReturnValue) => $"GLib.Native.StringHelper.ToStringUtf8({fromParam})",
 
                 // Record Conversions (safe handles)
-                ArrayTypeReference { Type: Record r, TypeReference: { CTypeReference: { IsPointer: true } } } when useSafeHandle => $"{fromParam}.Select(x => new {r.Write(Target.Managed, currentNamespace)}(x)).ToArray()",
-                ResolveableTypeReference { Type: Record r, CTypeReference: { IsPointer: true } } when useSafeHandle => $"new {r.Write(Target.Managed, currentNamespace)}({fromParam})",
-
+                ArrayTypeReference { Type: Record r, TypeReference: { CTypeReference: { IsPointer: true } } } when useSafeHandle => $"{fromParam}.Select(x => {r.Write(Target.Managed, currentNamespace)}.__FactoryNew(x)).ToArray()",
+                ResolveableTypeReference { Type: Record r, CTypeReference: { IsPointer: true } } when useSafeHandle => $"{r.Write(Target.Managed, currentNamespace)}.__FactoryNew({fromParam})",
```
badcel, line 75: `__FactoryNew` is not idiomatic C#. It should not be available in the public API.

Generator/Templates/module_type_registration.sbntxt
```diff
@@ -7,16 +9,40 @@ namespace {{ namespace.name }}
...
+                Console.WriteLine($"Could not register class type '{{ $class.name }}': {e.Message}");
             }
             {{ end }}
+            
+            {{ for $record in records }}
+            try
+            {
+                {{ $record | write_type_registration_record }}
+            }
+            catch (Exception e)
```
badcel, line 29: Why do we need a catch here? If it fails the application should crash. It just maps the types in a dictionray.

badcel: If there are records / classes / unions without a `GetGType` method we should not fail with an exception but gracefully handle this case.

badcel: The try catch is needed because there are some GTK classes like "GtkPlug" which are not available on all operating systems. The "GtkPlug" got removed in Gtk4. which should solve this issue. So just add a TODO which states that the try catch should be removed if we switch to Gtk4.

Generator/Templates/record.sbntxt
```diff
@@ -21,12 +21,18 @@ namespace {{ namespace.name }}
...
         partial void Initialize();
         
-        public {{ $record_name }}({{ $safe_handle }} handle)
+        private {{ $record_name }}({{ $safe_handle }} handle)
         {
             _handle = handle;
             Initialize();
         }
         
+        public static {{ $record_name }} __FactoryNew({{ $safe_handle }} handle)
```
badcel, line 30: Not idiomatic C#, should not be inside public API.

Libs/GModule-2.0/GModule-2.0.csproj
```diff
@@ -0,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
badcel, line 1: GModule is not used anywhere. Either delete it completely or add to projects which should be generated.

Libs/GObject-2.0/Classes/Object.Properties.cs
```diff
@@ -55,7 +55,7 @@ protected Value GetProperty(string name)
             var handle = Native.Value.ManagedHandle.Create();
             Native.Object.Instance.Methods.GetProperty(Handle, name, handle);
 
-            return new Value(handle);
+            return Value.__FactoryNew(handle);
```
badcel, line 58: Remove

Libs/GObject-2.0/Classes/Object.cs
```diff
@@ -97,8 +99,11 @@ protected internal void OnPropertyChanged([CallerMemberName] string? propertyNam
 
         public virtual void Dispose()
         {
+            Debug.WriteLine($"Disposing Object: Address {_handle.Handle}, Type {GetType()}.");
             _signalRegistry.Dispose();
             _handle.Dispose();
         }
+
+        ~Object() => Dispose();
```
badcel, line 107: This should not be needed. The GC should automatically garbage collect our instances. I wonder how it works for the signalRegistry in the `Dispose` method. Needs to be evaluated.

Libs/GObject-2.0/Native/Classes/ObjectMapper.ToggleRef.cs
```diff
@@ -45,12 +46,14 @@ public ToggleRef(IntPtr handle, object obj, bool ownedRef)
...
                 RegisterToggleRef();
+                
+                Debug.WriteLine($"Created ToggleRef: {GetLogState()}.");
             }
 
             private void RegisterToggleRef()
             {
                 Native.Object.Instance.Methods.AddToggleRef(_handle, _callback, IntPtr.Zero);
-                Native.Object.Instance.Methods.Unref(_handle);
+                // Native.Object.Instance.Methods.Unref(_handle);
```
badcel, line 56: This should be correct. Verify again.

Libs/GObject-2.0/Native/Classes/ObjectMapper.ToggleRef.cs
```diff
@@ -45,12 +46,14 @@ public ToggleRef(IntPtr handle, object obj, bool ownedRef)
 
                 OwnReference(ownedRef);
                 RegisterToggleRef();
+                
+                Debug.WriteLine($"Created ToggleRef: {GetLogState()}.");
```
badcel, line 50: `GetLogState` is executed even in release mode.

Libs/GObject-2.0/Native/Classes/ObjectMapper.ToggleRef.cs
```diff
@@ -80,17 +83,31 @@ private void ToggleReference(IntPtr data, IntPtr @object, bool isLastRef)
...
                     else
                         throw new Exception("Could not toggle reference to strong. It is garbage collected.");
-
                 }
                 else if (isLastRef && _reference is not WeakReference)
                 {
                     _reference = new WeakReference(_reference);
                 }
+                
+                Debug.WriteLine($"Toggled ToggleRef: {GetLogState()}.");
```
badcel, line 92: `GetLogState ` is executed in release mode. If we got this right one time it is obsolete.

Libs/GObject-2.0/Native/Classes/ObjectMapper.ToggleRef.cs
```diff
@@ -80,17 +83,31 @@ private void ToggleReference(IntPtr data, IntPtr @object, bool isLastRef)
...
             }
 
             public void Dispose()
             {
-                Native.Object.Instance.Methods.RemoveToggleRef(_handle, _callback, IntPtr.Zero);
+                Debug.WriteLine($"Disposing of ToggleRef: {GetLogState()} (note: pre-disposal state).");
+                
+                // Native.Object.Instance.Methods.RemoveToggleRef(_handle, _callback, IntPtr.Zero);
```
badcel, line 99: Why did this get removed?

Libs/GObject-2.0/Native/Classes/RecordWrapper.cs
```diff
@@ -0,0 +1,44 @@
...
+
+            object? result = factory.Invoke(null, new object[] { handle });
+            
+            if (result == null)
+                throw new Exception($"Type {trueType}'s factory method returned a null object. This could mean improperly defined bindings");
+
+            return result;
+        }
+        
+        private static MethodInfo? GetRecordFactory(System.Type type)
```
badcel, line 32: We do not want reflection during runtime.

badcel: Perhaps this class should be removed completely?

Libs/GObject-2.0/Native/Classes/TypeDictionary.cs
```diff
@@ -97,6 +97,9 @@ internal static System.Type GetSystemType(Type gtype)
             if (_reverseTypeDict.TryGetValue(gtype, out System.Type? sysType))
                 return sysType;
 
+            if (Functions.TypeIsA(gtype.Value, Type.Boxed.Value))
+                throw new Exception($"Record not registered: {gtype.ToString()}");
```
badcel, line 101: This message is irritating. The message says that the record is not registred but the condition resulting in the error checks that boxed types are not supported in general? Either the condition is wrong or the message must be adjusted.

badcel: Either way why do we need this at all?

Libs/GObject-2.0/Records/Value.cs
```diff
@@ -145,21 +145,32 @@ public void Set(object? value)
...
+            // TODO: It would be nice to support boxing arbitrary managed types
+            // One idea for how to achieve this is creating our own 'OpaqueBoxed' type
+            // which wraps a GCHandle or similar. We can then retrieve this at runtime
+            // from a static dictionary, etc. Alternatively, perhaps we want to find a
+            // method which plays nice with AOT compilation.
+            
+            // TODO: Should this be GetBoxed/TakeBoxed/DupBoxed? 
+            return RecordWrapper.WrapHandle(Native.Value.Methods.GetBoxed(Handle), new Type(type));
+
+            throw new NotSupportedException($"Can't get boxed value. Type '{new Type((nuint)type)}' is not supported.");
```
badcel, line 173: this line is obsolete.

Libs/Gdk-3.0/Records/Event.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    public partial class Event
+    {
+        public static Event __FactoryNew(IntPtr ptr)
+        {
+            var ev = Marshal.PtrToStructure<Native.EventAny.Struct>(ptr);
+            switch (ev.Type)
+            {
+                case EventType.Expose:
+                    return EventExpose.__FactoryNew(ptr);
```
badcel, line 14: Fix MethodName

Libs/Gdk-3.0/Records/Event.cs
```diff
@@ -0,0 +1,84 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace Gdk
+{
+    public partial class Event
+    {
+        public static Event __FactoryNew(IntPtr ptr)
```
badcel, line 8: Fix Name. Should it be public?

Libs/Gio-2.0/Classes/DBusConnection.cs
```diff
@@ -37,7 +37,7 @@ void Callback(GObject.Object sourceObject, AsyncResult res)
                 var ret = Native.DBusConnection.Instance.Methods.CallFinish(sourceObject.Handle, (res as GObject.Object).Handle, out var error);
                 Error.ThrowOnError(error);
 
-                tcs.SetResult(new Variant(ret));
+                tcs.SetResult(Variant.__FactoryNew(ret));
```
badcel, line 40: Revert

Libs/Gio-2.0/Classes/DBusConnection.cs
```diff
@@ -55,7 +55,7 @@ public Variant Call(string busName, string objectPath, string interfaceName, str
 
             Error.ThrowOnError(error);
 
-            return new Variant(ret);
+            return Variant.__FactoryNew(ret);
```
badcel, line 58: Revert

Libs/Gtk-3.0/Classes/IMContextSimple.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Gtk
+{
+    public partial class IMContextSimple
+    {
+        public static IMContextSimple New()
+            => new IMContextSimple(Native.IMContextSimple.Instance.Methods.New(), false);
```
badcel, line 6: According to documentation we own the reference returned by the new method: https://docs.gtk.org/gtk3/ctor.IMContextSimple.new.html

So `false` -> `true`?

Libs/Gtk-3.0/Classes/TreeSelection.cs
```diff
@@ -11,7 +11,7 @@ public void GetSelected(out TreeModel model, out TreeIter iter)
             Native.TreeSelection.Instance.Methods.GetSelected(Handle, out var modelPtr, iterHandle);
 
             model = ObjectWrapper.WrapHandle<TreeModel>(modelPtr, false);
-            iter = new TreeIter(iterHandle);
+            iter = TreeIter.__FactoryNew(iterHandle);
```
badcel, line 14: Revert

Libs/xlib-2.0/xlib-2.0.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
badcel, line 1: As nobody is using xlib we should either include it in the build projects or remove it.

Libs/GObject-2.0/Native/Classes/RecordWrapper.cs
badcel: Currently necessary because we do the same for objects. Perhaps we can get rid of this via #397

Libs/GObject-2.0/Classes/Object.cs
badcel: The signal registry is a local object of the instance itself. It just gets garbage collected.

PR #326, opened 2021-05-23
Generator: Differentiate between pointed / unpointed records

badcel:
Unpointed records (value based)
* For non array records in the gir their original record structure must be used.
* For arrays of records in the gir their original record structure must be used inside an array.

Pointed records (pointer based)
* For non array records in the gir a pointer must be used. This could either be an SafeHandle (regular case) or an IntPtr (e.g. in case of free methods).
* For arrays of records always IntPtr[] must be used as arrays of SafeHandles are not supported by the runtime.

The method and delegate generators currently do not differentiate between pointed and unpointed records. To avoid compilation errors those cases are currently passing default values to methods and marked with a TODO flag.

This fixes the window sample.

### DIFF
--- a/Generator/Convert.cs
+++ b/Generator/Convert.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using GirLoader.Output.Model;
-using Array = System.Array;
 using String = GirLoader.Output.Model.String;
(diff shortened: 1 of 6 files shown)


badcel:
Merged as it is a bug fix.

PR #338, opened 2021-05-27
Fix Generator on Windows

mjakeman:
For some reason, the generator has stopped working on Windows (despite working on Linux and containing no platform-specific code). This change really shouldn't make a difference, but somehow it does. Perhaps we've encountered a runtime bug?

This PR:
 - Changes how we handle Repository to make the generator work again
 - Adds CI Matrix builds for Windows and MacOS


badcel:
Superseeded by #344

PR #400, opened 2021-08-10
GirLoader refactorings

badcel:
- Provide non static public interface
- Flatten folder hierarchy


GirLoader/TypeReferenceResolver.cs
```diff
@@ -32,7 +32,7 @@ public bool Resolve(Model.TypeReference typeReference, Model.Repository reposito
             return false;
         }
 
-        private bool TryResolveAlias(Model.TypeReference typeReference, Model.Repository repository, [MaybeNullWhen(false)] out Model.Type type)
+        private bool TryResolveAlias(Output.TypeReference typeReference, Output.Repository repository, [MaybeNullWhen(false)] out Output.Type type)
         {
             type = null;
```
mjakeman, line 37: Is `type = null` necessary? It seems a bit redundant if RecursiveResolveAlias returns null anyway.

mjakeman (commented):
Great work with the structure!

It's much better now than before and resolves my issue over having multiple files of the same name nested in subdirectories.

A few small issues + CI doesn't seem to be running?

PR #558, opened 2022-02-16
Improve reference resolving

badcel:


badcel:
The current version of dotnet 6.0.200 seems to have a bug in `dotnet format` as setting the dotnet-version to 6.0.102 does not fail the `dotnet format call opposed to 6.0.200.

Just compare the builds of the last two commits of this branch.

Issue: https://github.com/dotnet/format/issues/1519

PR #583, opened 2022-03-14
Require a foreign record's OwnedHandle to manually implement ReleaseHandle()

cameronwhite:
- For foreign records, ReleaseHandle() is marked partial and not implemented, so the manual bindings are required to fill it in manually with the appropriate cleanup function
- Implement ReleaseHandle() for all of the foreign records. Added a constant for the Cairo library name in the DllImportOverride class to avoid duplicated string constants
- The one exception is Cairo.Matrix, which doesn't have a cleanup function since it's a plain C struct that is always allocated and managed by the caller, so I think only MatrixManagedHandle will ever be used.

Related to PR #578


badcel:
This looks very good thank you.

I have one question regarding the `destroy` methods: What is the reason to use the abstract base safe handle as a parameter instead of the concrete owned handle?

Shouldn't the bindings only be allowed to decrement the reference count of instances it owns?

cameronwhite:
Good catch, yeah I agree it should use the OwnedHandle type to be more correct 👍 .
I had just been switching from IntPtr's (the original changes in #578) to SafeHandle's to add some type safety

PR #621, opened 2022-05-08
Add bindings for Cairo.Region

cameronwhite:
This also adds some additional wrapper code for `Cairo.RectangleInt` in order to implement the unit tests (e.g. constructing a rectangle, comparing equality, and field accessors) - this is the main thing I'd like to discuss since it's a more general topic on the generator side of things!

The extra level of indirection to access the fields (`RectangleInt` vs `RectangleIntData`) feels kind of awkward - my current changes have some rough code that attempts to hide this. I feel like it would be more natural if the generator could just directly emit a `struct` type here, since on the C side this is a [plain struct](https://cairographics.org/manual/cairo-Types.html#cairo-rectangle-int-t) rather than an opaque / handle type

I think it might also make something like [cairo_region_create_rectangles()](https://cairographics.org/manual/cairo-Regions.html#cairo-region-create-rectangles) simpler to wrap as well since the memory layout would match 
The old cairo bindings from Mono / GtkSharp also made this a struct type, so changing from value to reference semantics would also make it tricky to port existing code correctly

### DIFF
--- a/src/Libs/cairo-1.0/Internal/Records/Region.cs
+++ b/src/Libs/cairo-1.0/Internal/Records/Region.cs
@@ -5,7 +5,68 @@ namespace Cairo.Internal
 {
     public partial class Region
     {
+        // Not currently wrapped:
+        // - cairo_region_create_rectangles (needs to be an array of RectangleIntData?)
+        // - cairo_region_equal (if wrapped, need a hash function to go along with it)
+
+        [DllImport(CairoImportResolver.Library, EntryPoint = "cairo_region_contains_point")]
+        public static extern bool ContainsPoint(RegionHandle handle, int x, int y);
+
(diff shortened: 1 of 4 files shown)


badcel:
Should wait until #622 is fixed to rely on improved struct generation.

cameronwhite:
I've rebased this on top of the latest changes - the `RectangleInt` struct is much more usable now without any manual bindings 👍 
The two awkward things I encountered are:
- Constructing it requires `new RectangleInt(Internal.RectangleIntManagedHandle.Create())` for now, as was noted in the release notes
- Equality comparison doesn't work, I think because it includes the `Handle` member. So code which could have been written as  `rect_copy.Should().Be(rect_init)` requires comparing all the members explicitly. Mostly just an annoyance for writing unit tests :)

badcel:
Ah good point! I will add some equality check to the generator which just compares the pointers.

PR #625, opened 2022-05-10
Add the remaining general Cairo.Context bindings

cameronwhite:
This is a followup to #624 and adds the remaining bindings from https://cairographics.org/manual/cairo-cairo-t.html.
The one exception is `cairo_copy_clip_rectangle_list()`, which probably needs some special wrapping since `cairo_rectangle_list_t` isn't very friendly to expose directly in C#


badcel:
Thank you very much :+1:

PR #632, opened 2022-05-29
Add bindings for Cairo.Pattern

cameronwhite:
This follows the suggested API structure from https://cairographics.org/manual/bindings-patterns.html. The "mesh" pattern type will be left for a separate PR

The unit tests revealed that Cairo.Internal.MatrixData was the wrong size (empty) since the fields aren't in the gir file, so those have been manually added

### DIFF
--- a/src/Libs/cairo-1.0/Records/SolidPattern.cs
+++ b/src/Libs/cairo-1.0/Records/SolidPattern.cs
@@ -1,0 +1,19 @@
+﻿namespace Cairo
+{
+    public class SolidPattern : Pattern
+    {
+        private SolidPattern(Internal.PatternHandle handle)
+            : base(handle)
+        {
+        }
+
+        public static SolidPattern CreateRgb(double r, double g, double b)
+            => new SolidPattern(Internal.Pattern.CreateRgb(r, g, b));
+
+        public static SolidPattern CreateRgba(double r, double g, double b, double a)
+            => new SolidPattern(Internal.Pattern.CreateRgba(r, g, b, a));
+
+        public Status GetRgba(out double r, out double g, out double b, out double a)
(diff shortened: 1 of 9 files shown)


src/Libs/cairo-1.0/Records/SolidPattern.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Cairo
+{
+    public class SolidPattern : Pattern
+    {
+        private SolidPattern(Internal.PatternHandle handle)
+            : base(handle)
+        {
+        }
+
+        public static SolidPattern CreateRgb(double r, double g, double b)
```
badcel, line 10: Why does this class has static create methods but other classes use a constructor?

badcel (commented):
Looks good to me except one small question. See comment.

src/Libs/cairo-1.0/Records/SolidPattern.cs
cameronwhite: In the cairo docs there was a suggestion to do this:

> The functions that should map to constructors or static methods for the various types are listed after the type, methods on that type are listed below. Note that cairo_pattern_create_rgb() and cairo_pattern_create_rgba() should not be overloaded with each other as a SolidPattern() constructor, but should appear as static methods instead. This is to maintain code clarity by making it clear how the arguments relate to color components

But it does feel inconsistent so I'm not 100% sure if we should follow that.
FWIW, in GtkSharp these were done as overloaded constructors (along with further overloads for the `Cairo.Color` struct that they added for convenience)

badcel: I think if it matches the official documentation it is okay 👍

badcel:
Thank you for the good work 🚀

PR #635, opened 2022-05-30
Add Cairo.Context text-related methods

cameronwhite:
This adds the methods from https://cairographics.org/manual/cairo-text.html, minus the `cairo_glyph_t` methods which require bindings for additional types.
The methods related to the toy font face API (`cairo_toy_font_face_create()` etc) have also been added

### DIFF
--- a/src/Libs/cairo-1.0/Internal/Records/Context.Text.cs
+++ b/src/Libs/cairo-1.0/Internal/Records/Context.Text.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace Cairo.Internal
+{
+    public partial class Context
+    {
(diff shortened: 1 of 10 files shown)


badcel:
Thanks for your work again :+1: 

The "toy" api is currently kind of hidden inside the internal namespace. What do you think of creating a "ToyFontFace" internal class?

badcel:
> That sounds ok, so adding e.g. `Internal/Records/ToyFontFace.cs` to have the bindings for methods like `cairo_toy_font_face_get_family()`?

Yes I think it is worth a try as the documentation says that it is for testing only and one should avoid it for production use. So I think a clean separation could be useful.

badcel (approved):
LGTM 🚀

PR #649, opened 2022-07-24
Changed member generation formatting to use invariant culture

### LINKED ISSUE #648: Negative values in enums use an incorrect unicode character as the minus sign
I don't know if this is my setup or the generator but the following lines in all files get reported as errors (I just cloned the repo today and ran the generation):
```c#
        Nothing = −1,
```
in fact all numbers with the minus sign in front of them seem to be reported like this. In my IDE that character looks exactly like the minus sign (`-`) but seems like here on Github the character actually looks distinct from the minus sign.

Using emacs I confirmed that the actual characters in the files are not the normal minus sign but a unicode character that looks the same:
Kuvakaappaus - 2022-07-23 13-46-43

Here's what it looks like after I corrected the character:
Kuvakaappaus - 2022-07-23 13-46-56

So it seems the generator writes out a "MINUS SIGN" character instead of a "HYPHEN-MINUS" like it should.

hhyyrylainen:
this fixes generating enums in different user locales

closes #648

### DIFF
--- a/src/Generation/Generator3/Renderer/Public/Member.cs
+++ b/src/Generation/Generator3/Renderer/Public/Member.cs
@@ -1,8 +1,10 @@
-﻿namespace Generator3.Renderer.Public
+﻿using static System.FormattableString;
+
+namespace Generator3.Renderer.Public
 {
     public static class Member
     {
         public static string Render(this Model.Public.Member member)
-            => $"{ member.Name } = { member.Value },";
+            => Invariant($"{ member.Name } = { member.Value },");
     }
 }


badcel:
Could you add a code comment which explicitly describes why the invariant culture needs to exist?

(Just to make it as easy as possible to understand the reason for the code to avoid removing it during some cleanup or similar.)

hhyyrylainen:
It is likely to come up in multiple other places, in fact I think the general styleguide should be updated to say that all computer readable text formatting must use invariant culture, because basically you are otherwise asking for trouble. That's just a general portability thing that any computer readable output generating code needs to take into account. Fixing all existing code to follow that new advice is a bit too much for me right now, especially as I ran into missing features very fast when evaluating whether I want to do a project with gir.core or not...

badcel:
Thanks for evaluating the project 👍. It's still a lot of work. I think the project could be very valueable but there are very little helping hands currently so it moves quite slowly.

I think you are right. All the renderers should probably just output invariant cultured code. 

I will merge this PR in the coming days.

hhyyrylainen:
I've been looking for a way to make cross platform GUI with C# for a while now. All options seem to run into issues very fast when I've tried them. I think I was a bit too hasty to discard gir.core, especially if making cross platform releases will be relatively simple with precompiled native binaries. I've opened a bunch of issues for what problems I ran into, which if can be resolved in a reasonable time, I think I could get behind using gir.core and fixing some stuff like the 6597 warnings I get.

badcel:
There are lots of things missing. If you increase the log level there are many messages regarding things which are not generated because something is missing in the generator. 

I think it is to early for me to check why certain methods are missing as complete areas of code are not generated yet. 

Currently I rework the generator to simplify its architecture. If this is done I will continue working on code generation. 

You are very welcome to support the project. If you are really interested I could introduce you to the generator code so you can work on things you are missing yourself.

If you don't have the time to spare you could check out GtkSharp which provides bindings for GTK3 and should be production ready from a users point of view.

hhyyrylainen:
That's understandable if things are so unfinished that specific areas don't fully make sense to have open issues for them separately. I think before I'll fix any more bugs in this library, I'll try some of the other potential alternatives I found and see if the bugs in them are easier for me to fix.

I specifically ruled out any library that doesn't support Gtk4, I want to make a future proof project so that's basically a requirement. And it seemed like GtkSharp, if I didn't mix it up with any other project, didn't even seem like they were open to Gtk4 support instead they pointed people to projects that were building Gtk4 support from the ground up.

badcel:
If you want to use Gtk4 and C# gir.core will hopefully get the goto library.

One reason I created this project is to have a future proof generator which the GtkSharp one is not from my point of view.

As far as I know there is no alternative which aims to support multiple gobject libraries including Gtk4. This is why I created this project. So I hope to get new devs on board to bring C# and the GObject type system into a useable state.

Issue #559, opened 2022-02-17
Use most recent .NET SDK Version for CI builds


badcel:
There is a bug in SDK 6.0.200 which fails the code format verification. See: https://github.com/dotnet/format/issues/1519

Revert commit d04fbfcf41596cb2255ed969ee0333335c8baccd once the bug is solved.

Failed build: https://github.com/gircore/gir.core/runs/5220806768?check_suite_focus=true
Working build only SDK-Version set to 6.0.102: https://github.com/gircore/gir.core/runs/5222828020?check_suite_focus=true

badcel:
As there is a new release but the bug is still present I opened a new bug to get this fixed: https://github.com/dotnet/format/issues/1572

badcel:
The current dotnet version 6.0.202 reports a non 0 exit code because of the warning of: #511

In case the `ModuleInitializerAttribute` is removed we can use the most recent dotnet version again.

Issue #511, opened 2021-12-31
Fix Warning CA2255


badcel:
During compilation of the libraries the CA2255 warning is raised:

`[CA2255] The 'ModuleInitializer' attribute is only intended to be used in application code or advanced source generator scenarios`

Documentation: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2255

We would need to create some Methods to initialize the modules. GTK would probably need to initialize GObject, too. GObject itself would need to check if it is already initialized and skip initialization in case there is a user initiated initialization.

PR #660, opened 2022-08-19
Remove module initializers

badcel:
- [x] Update github actions to use current dotnet again.

Fixes: #559 
Fixes: #511


src/Libs/Gtk-3.0/Public/Functions.cs
```diff
@@ -4,7 +4,7 @@ namespace Gtk
 {
     public class Functions
     {
-        public static void Init()
+        internal static void Init()
```
badcel, line 7: Keep public

PR #691, opened 2022-11-06
Render interface methods

badcel:
- [x] Add a comment to the fixer stating that it should be removed and replaced with an explicit interface implementation while rendering.
- [x] call fixer via namespace: "Fixer.Classes.Fixup(@namespace.Classes);"
- [x] call generators via namespace: "Generator.Classes.Generate(@namespace.Classes);"


nlogozzo:
Any progress on this PR. This is all i'm waiting for to port my app ([Money](https://github.com/nlogozzo/NickvisionMoney)) so i can use things like GActionMap and GtkFileChooser.

badcel:
This PR is my primary focus but I can't promise a finishing date.

I think it is mostly done probably only some corner cases remain.

If you consider porting your app it would probably make sense to familiarize with the code as I expect you will miss other things as well and as of my current situation I'm not able to code as much as I would like to support the needs of other projects.

nlogozzo:
> In case you are interested you are very welcome to join our matrix chat.

Do you have a like to the room for me to join? :)

badcel:
It is mentioned on the readme but here you go: 

[#gircore:matrix.org](https://matrix.to/#/%23gircore:matrix.org?via=matrix.org)

nlogozzo:
Seems like a lot of fun going on here :grinning:

PR #723, opened 2022-12-26
Add binding for cairo_image_surface_get_data()

cameronwhite:
This is used for low-level (via `unsafe`) modification of the image buffer

Note in GtkSharp this was called `DataPtr` because they also had a separate `Data` property that returned a `byte[]` copy (https://github.com/GtkSharp/GtkSharp/blob/develop/Source/Libs/CairoSharp/ImageSurface.cs#L85), but I think calling the property `Data` is better to match the underlying Cairo API

### DIFF
--- a/src/Libs/cairo-1.0/Internal/ImageSurface.cs
+++ b/src/Libs/cairo-1.0/Internal/ImageSurface.cs
@@ -1,4 +1,5 @@
-﻿using System.Runtime.InteropServices;
+﻿using System;
+using System.Runtime.InteropServices;
 
 namespace Cairo.Internal;
 
@@ -9,6 +10,9 @@ public class ImageSurface
 
     // Skip cairo_image_surface_create_for_data for now, as it has some
     // additional lifetime requirements for the input buffer.
+
+    [DllImport(CairoImportResolver.Library, EntryPoint = "cairo_image_surface_get_data")]
+    public static extern IntPtr GetData(SurfaceHandle handle);
 
     [DllImport(CairoImportResolver.Library, EntryPoint = "cairo_image_surface_get_format")]
     public static extern Format GetFormat(SurfaceHandle handle);
--- a/src/Libs/cairo-1.0/Public/ImageSurface.cs
+++ b/src/Libs/cairo-1.0/Public/ImageSurface.cs
@@ -1,4 +1,6 @@
-﻿namespace Cairo;
+﻿using System;
+
+namespace Cairo;
 
 public class ImageSurface : Surface
 {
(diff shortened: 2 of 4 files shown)


badcel:
Thank you for the pr. Can you describe the use case for the pointer? I would like to keep pointers to a minimum in the public API.

Pycairo returns a "memoryview" for get_data (https://pycairo.readthedocs.io/en/latest/reference/surfaces.html). I think this implicates read only access to the data without copping it. Would this be sufficient for our public API?

cameronwhite:
This is used a lot in Pinta for any image manipulation that can't be done with Cairo's drawing APIs, like many of the effects (e.g. a gaussian blur). A simple example would be something like https://github.com/PintaProject/Pinta/blob/master/Pinta.Tools/Tools/PaintBucketTool.cs#L76

The pointer provides direct access to the image buffer and allows writing to specific pixels. The memory layout depends on the image surface's format (https://cairographics.org/manual/cairo-Image-Surfaces.html#cairo-format-t), so in Pinta's case it's casted to a pointer to a BGRA pixel struct.

`Span<T>` could probably be used these days, but `unsafe` is still needed to construct that from a pointer so I'm not sure if we want to have `unsafe` enabled in gir.core?

badcel:
I think unsafe is okay and we should probably at least verify the span route.

It would be great to see how we could improve the API especially as pinta is a pretty good application to get a lot of use cases covered 👍

cameronwhite:
Sounds good, I'll do a quick test and benchmark in Pinta to verify that the span approach works well

badcel:
It would be great if you could add a simple unit test, too if it turns out to be a good solution.

cameronwhite:
Did a quick test in Pinta and the performance is actually slightly better with `Span<ColorBgra>` - the benchmark for the `RadialBlurEffect` went from 5,923.8ms to 5,757.7ms, so I think this is a valid approach

I'm thinking that `ImageSurface` can provide a `Span<byte>` with the surface's data, then Pinta can use `MemoryMarshal.Cast<byte, ColorBgra>(span)` to cast to the BGRA format that Pinta uses.

badcel:
Couldn't we provide the structs and provide extension methods like `surface.GetData().AsBgra()` or similar?

cameronwhite:
I was a bit hesitant about that because Pinta's ColorBgra has a lot of things in it that are probably beyond the scope of what gir.core should provide and belong in Pinta's codebase (https://github.com/PintaProject/Pinta/blob/master/Pinta.Core/Effects/ColorBgra.cs).
But I suppose a basic version of that struct could be provided, and then Pinta can move everything else to extension methods / utility functions

badcel:
Does Cairo provide any support for those types or is it just providing the type as an enum?

If not I agree with you. Then I would prefer to just return an "SurfaceData : Span<byte>" and leave the rest to the user e like Cairo does.

cameronwhite:
No, I don't think cairo provides any support for those types other than the Format enum - the image data is only provided as an `unsigned char *`

badcel:
In general this looks good to me. I wonder in which case we create a property and in which case a method? What do you think of only using methods in cairo and remove properties?

To merge this the branch needs to be rebased onto the current main.

PR #725, opened 2022-12-27
Fix a couple issues with string (array) parameters

cameronwhite:
This addresses a couple issues encountered with global functions (#724)

- Raise an error for out string parameters with caller-allocates=true. This occurs for `g_unichar_to_utf8()`, which expects the caller to provide a suitably sized buffer, so I don't think we can safely generate bindings automatically for functions like that

-  String arrays that are a ref/out parameter produce compile errors with the current code and require further work to support. In particular, some functions like `hb_feature_to_string()` seem to incorrectly appear with string array out parameters, but actually behave like `g_unichar_to_utf8()` above and expect a pre-allocated buffer.

- While looking into this, I noticed that string arrays should also be converted to UTF8 strings, matching regular string parameters

### DIFF
--- a/src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/String.cs
+++ b/src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/String.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using Generator.Model;
 
 namespace Generator.Renderer.Public.ParameterToNativeExpressions;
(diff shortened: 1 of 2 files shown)


src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/String.cs
```diff
@@ -10,6 +11,10 @@ public bool Supports(GirModel.AnyType type)
 
     public void Initialize(ParameterToNativeData parameter, IEnumerable<ParameterToNativeData> _)
     {
+        // This is tricky to auto-generate bindings for, since it can expect the
```
badcel, line 14: Could you update this comment to something along the lines: `//Todo the caller needs to pass in some kind of Span<T> as a buffer which can be filled by the C function`

badcel (commented):
I think if the comment is added I don't need to create a separate issue as there are many places in the code which are not implemented and don't have an issue. The todo then already mentions a possible way forward.

PR #728, opened 2022-12-28
Fix errors with marshalling record array parameters.

cameronwhite:
This occurred for `g_dbus_annotation_info_lookup()` which #724 now generates bindings for

- The public type should just be `Gio.DBusAnnotationInfo[]`, not `IntPtr[]`

- If there isn't a parameter that specifies the length, ensure that a null terminated pointer array is provided to the native function

- Fix a typo in the generated code: `DangerousGethandle` -> `DangerousGetHandle`. This code path wasn't being hit previously :)

The generated code for this function was previously:
```
public static string? DbusAnnotationInfoLookup(IntPtr[] annotations, string name)
{
    var annotationsNative = annotations.Select(record => record.Handle.DangerousGethandle()).ToArray();
    var result = Gio.Internal.Functions.DbusAnnotationInfoLookup(annotationsNative, name);

    return GLib.Internal.StringHelper.ToStringUtf8(result);
}
```
And is now:
```
public static string? DbusAnnotationInfoLookup(Gio.DBusAnnotationInfo[] annotations, string name)
{
    var annotationsNative = annotations.Select(record => record.Handle.DangerousGetHandle()).Append(IntPtr.Zero).ToArray();
    var result = Gio.Internal.Functions.DbusAnnotationInfoLookup(annotationsNative, name);

    return GLib.Internal.StringHelper.ToStringUtf8(result);
}
```

### DIFF
--- a/src/Generation/Generator/Renderer/Public/Parameter/From/ArrayRecordParameter.cs
+++ b/src/Generation/Generator/Renderer/Public/Parameter/From/ArrayRecordParameter.cs
@@ -16,9 +16,7 @@ internal static class ArrayRecordParameter
     {
         var arrayType = parameter.AnyType.AsT1;
 
-        return arrayType.Length is null
-            ? Type.PointerArray
-            : ComplexType.GetFullyQualified((GirModel.Record) arrayType.AnyType.AsT0) + "[]";
+        return ComplexType.GetFullyQualified((GirModel.Record) arrayType.AnyType.AsT0) + "[]";
     }
 
     private static string GetDirection(GirModel.Parameter parameter) => parameter switch
--- a/src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/RecordArray.cs
+++ b/src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/RecordArray.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Text;
 using Generator.Model;
 
 namespace Generator.Renderer.Public.ParameterToNativeExpressions;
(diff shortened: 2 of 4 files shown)


badcel:
Thanks for the PR. Unfortunately the changes are probably not enough. While checking out your changes I saw that there is a `zero-terminated` attribute in the [XML definition](https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/docs/gir-1.2.rnc#L459) file which is already [known in our interface](https://github.com/gircore/gir.core/blob/49eddb4871359b07a48ee49b41dd797efa830e7f/src/Generation/GirModel/ArrayType.cs#L6). The attribute is set in the [code](https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gdbusintrospection.c#L1845) and in the [documentation](https://docs.gtk.org/gio/type_func.DBusAnnotationInfo.lookup.html).

Unfortunately the attribute is missing in the GIR file. So I went ahead and asked in #introspection:gnome.org on matrix and the answer is that the gir generator is not emitting the flag as it is assumed to be true if some condition is met: 

> if zero-terminated is missing, there's no length, there's no fixed size, and the name attribute is unset, then zero-terminated is true

To code this in the correct way the "IntPtr.Zero" should probably only be added if `zero-terminated` is true. Additionally the `GirLoader` has to implement the logic mentioned above in https://github.com/gircore/gir.core/blob/49eddb4871359b07a48ee49b41dd797efa830e7f/src/Generation/GirLoader/Output/TypeReferenceFactory.cs#L63

As the C gir generator will probably some day _always_ output the `zero-terminated` information correctly there should be a TODO in this code place to remove the logic again.

The next problem is I actually don't know if the xml parser allows to change `ZeroTerminated` type from `bool` to `bool?` here: https://github.com/gircore/gir.core/blob/49eddb4871359b07a48ee49b41dd797efa830e7f/src/Generation/GirLoader/Input/ArrayType.cs#L11

This is needed to be able to differentiate between `false` and `null` in the xml. So this is not only a trivial fix but needs work in several places.

cameronwhite:
The `bool?` didn't work for the XML parsing as you thought, so I just added it as a string to be parse later, similar to how the array length etc behave

badcel:
I need to test it, but coding wise it looks perfectly fine 👍 

Can you rebase on main? I will merge after testing.

PR #733, opened 2023-01-02
Disable string parameters with a direction of inout

cameronwhite:
The default marshalling produces crashes for functions like `pango_skip_space()`, so more investigation is required.

This is related to #724, where a compile error (missing 'ref' keyword) was also produced because the surrounding code only handled 'in' and 'out' strings.
So far, `inout` string parameters only occur for deprecated pango functions.


badcel:
Could you remove the curly brackets?

Issue #505, opened 2021-12-26
Support WebKitGTK once it supports GTK4


badcel:
Support WebKitGTK once it supports GTK4

Original sample with JavaScript integration: https://github.com/gircore/gir.core/blob/v0.0.1/Samples/Gtk3/GtkApp/MyWindow.cs

mikkeljohnsen:
Could WebKitGTK for GTK3 not be wrapped also ?

Since both Gtk3 and Gtk4 is wrapped.

badcel:
No it won’t be supported as GTK3 will be removed see #396.

PR #746, opened 2023-01-08
Add webkitgtk

badcel:
Fixes: #505 

Depends on: #752


mcatanzaro:
This is pretty cool! Quick note: webkit2gtk-5.0 is/was an unstable/development API version that is removed in WebKitGTK 2.40 and replaced by webkitgtk-6.0, which will be the stable API for GTK 4. webkitgtk-6.0 is still changing in every release, but should become stable within the next few weeks in time for GNOME 44 and WebKitGTK 2.40.0.

badcel:
Thank you for the info. I opened issue #799 to keep track of this.

PR #760, opened 2023-01-14
Remove incomplete manual binding for TimeoutAddFull

cameronwhite:
Now that global functions are generated, this is available as `GLib.Functions.TimeoutAddFull()` (the generated version works fine in my testing).

GtkSharp had a similar `GLib.Timeout` class so I stumbled across this while porting Pinta, and then found it didn't work :)


cameronwhite:
Somewhat related, I noticed that https://docs.gtk.org/glib/func.timeout_add_full.html mentions `This method is renamed to [g_timeout_add()](https://docs.gtk.org/glib/func.timeout_add.html) in language bindings`.
So perhaps the `shadows` property is what indicates it should be renamed to `TimeoutAdd`?

badcel:
Oh very interesting catch. This is used for method overloading which is not yet implemented. See #761

Issue #763, opened 2023-01-16
Verify handling of out and inout array parameters


cameronwhite:
This is a future TODO since `out` and `inout` array parameters are currently skipped for public bindings, at least for primitive and string values, but I also saw an issue with the internal bindings and wanted to record my findings for future reference

The internal bindings for functions such as `g_input_stream_read()` are incorrect and result in a crash, which I also reproduced with a simple C test library
- For an `out` parameter with `caller-allocates=true` the `ref byte[]` should actually just be `byte[]` since arrays are already a reference type. This could explicitly be `[Out] byte[]` [to be safe](https://stackoverflow.com/a/14366422) in case the native function doesn't get a direct pointer to the managed array.
- Similarly an `inout` parameter would be `[In, Out] byte[]`, not `ref byte[]`
- `out` parameters with `caller-allocates=false` (if they exist) would require investigation and possibly custom marshalling to deal with the memory allocation.

badcel:
Consider using span / memory?

cameronwhite:
I think Span would make sense for `caller-allocates=true` since then a stack-allocated buffer could be passed in.
Not sure if there are any cases of `caller-allocates=false`, but in that case the native function is probably doing the allocation and passing a pointer back so Span probably doesn't make sense there

badcel:
@cameronwhite I think I would generate the internal API with `caller allocates true` similar to my HidApi implementation: [Sample](https://github.com/badcel/HidApi.Net/blob/9fb1bac6a07b14556151d482b90a3ace49fd0580/src/HidApi.Net/Internal/NativeMethods.cs#L42)

In HidApi I use a `ref byte` instead of `byte[]` as it is easy to get the reference to the first byte from a span. creating a byte array from a span would probably copy the data.

cameronwhite:
👍 that sounds like the right approach to me for the internal API

Issue #504, opened 2021-12-26
Verify how to run system tests during CI


badcel:
If docker container allows to start a display we could run the system tests during CI.

badcel:
Alternatively there could be a gobject based c test library which provides an api to test the gobject system in depth without the need of a display

badcel:
UI-Shooter has a headless mode spawning a Weston instance without a display: https://gitlab.com/rmnvgr/uishooter

Perhaps we could do the same to execute GTK unit tests?

badcel:
In regard to #764:

My first idea would be that this repository should not care about some C code. There could be a new repository inside the GirCore organization which contains the C-Test library.

The new repository could have some GitHub action to build the code for different platforms and host the binaries and gir files alongside the code.

This repo is included as a git subproject in GireCore/gir.core repository.

The binding for the new c library would be part of the GirCore solution but would not be published.

This allows us to move all tests over to the new library freeing us of the need to have GTK installed to execute the tests.

cameronwhite:
Thanks! Yeah I think moving it to another repo makes sense
The binaries should be fairly small / not frequently updated so storing them in the repo should hopefully be fine

PR #764, opened 2023-01-16
Add a C test library for easier testing of the generated bindings

cameronwhite:
This is related to #504 - I decided to experiment with this after writing a C library to test out #763, and found it really useful to have a simple C function (which I could also add printouts to, etc) to test against
I think this could make it easier to have more extensive tests for correctness, which can run without a display and without requiring a lot of boilerplate code to call a function with the parameter type we want to test (e.g. having to construct a GTK application and widgets)

Still a work in progress, but wanted to get some initial feedback on whether this would be worth adding?
Also a couple questions:
- Currently I have this as a separate script. I wasn't sure about making this part of the default build since it adds the requirement of having a C toolchain, gir-related tools, etc?
- The library is currently generated into the `Libs` folder like everything else, but it could perhaps go elsewhere. I'm not sure if there are any things to watch out for with this, e.g. would it be pulled into the Nuget packages?

TODOs
- [x] Needs testing on other platforms, currently only tested on macOS
- [x] Add to CI build script


badcel:
I did not look at the code but the idea is very welcome!

See #504

cameronwhite:
I've got the Windows build sorted out now, so this is building and running the test successfully on all platforms 🎉

badcel:
Can you update the reademe's Build section and include an `dotnet build Gir.Core.Libraries.slnf`?

Additionally I think it would be helpful to explain why the solution filter exists.

badcel (changes requested):
Thanks for your work, very well done. This helps a lot to establish a much improved testability of the generator it is very much appreciated.

Please fix the remaining remarks and it is good to go.

src/Native/GirTestLib/primitivevaluetype.c
```diff
@@ -0,0 +1,38 @@
+#include "primitivevaluetype.h"
```
badcel, line 1: Should the folder: `src/Native/GirTestLib/` be part of the solution to allow easy editing?

cameronwhite:
Thanks! I've pushed fixes for those issues

PR #790, opened 2023-02-05
gtksourceview bindings

### LINKED ISSUE #789: Bind GtkSourceView
Add bindings for GtkSourceView.

The gir file is already in gir-files repository.

hol430:
Eventually fixes #789


hol430:
@badcel - this seems to be basically working now. I've had to add support for rendering interface `GPtrArray` return types in the internal layer, but I can move this into a separate branch/PR if you prefer. Otherwise let me know if you spot anything else suspicious here.

src/GirCore.sln
```diff
@@ -204,6 +204,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "img", "img", "{15AC3143-968
 		..\docs\img\GtkApp.png = ..\docs\img\GtkApp.png
 	EndProjectSection
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "GtkSource-5", "Libs\GtkSource-5\GtkSource-5.csproj", "{9D19B388-AD8B-4DC5-B351-5D9FCEAE8FEA}"
```
badcel, line 207: Can you please add a new solution folder named "GtkSource-5" into the Samples folder of the solution file and add your sample project to it?

In case you don't have an IDE supporting this I can add this for you.

src/Samples/GtkSource-5/SourceView/SourceViewTest.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
badcel, line 1: Can you please rename your project into something like "GtkSourceView" "SimpleGtkSourceView". Please avoid the suffix "Test" as this would look like it is a unit test project.

src/Samples/GtkSource-5/SourceView/Program.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    var window = ApplicationWindow.New((Application) sender);
+    window.Title = "GtkSource5 Demo";
+    window.SetDefaultSize(300, 300);
+    var buf = GtkSource.Buffer.New(null);
+    var view = GtkSource.View.NewWithBuffer(buf);
+    view.Monospace = true;
+    view.ShowLineNumbers = true;
+    Settings? settings = Settings.GetDefault();
+    if ( (settings?.GtkApplicationPreferDarkTheme ?? false) ||
```
badcel, line 16: From my point of view the condition can be a bit simplified. If you aggree you can change it to:

```CS
    if(settings?.GtkApplicationPreferDarkTheme == true || 
       settings?.GtkThemeName?.ToLower().Contains("dark") == true)
        buf.SetStyleScheme(GtkSource.StyleSchemeManager.GetDefault().GetScheme("Adwaita-dark"));
```

badcel (changes requested):
Technically this is fine. Just some minor remarks regarding the sample and solution file.

hol430:
All good suggestions. I think I've implemented them all, though I'm unsure if I've setup the .sln file correctly, as my IDE doesn't support editing them visually, and the CLI in this area seems somewhat ambiguous.

src/Samples/GtkSource-5/SourceView/Program.cs
hol430: Yeah, I'm not sure if it's even worth having this extra code at all, or whether it would be better to keep things as minimal/clear as possible in the example. Without loading this style scheme though, the widget will be white background on gnome with dark mode enabled, so I've left it in for now.

badcel (approved):

badcel:
Yep, you have nailed it, even if done with the CLI :+1:: The folder structure looks perfectly fine in Rider.

All the other points are fixed, too. If you do a final rebase this is good to go: :rocket:

badcel:
Thank you :rocket:

PR #794, opened 2023-02-07
Rename Signal.Connect parameter

### LINKED ISSUE #793: Rename Signal.Connect parameter
The first parameter `o` of the `Signal.Connect` method is not descriptive. It should have a more descriptive name like `obj`.

This needs to be changed for all overloads and all Signal classes:
 - [SignalTSender.cs](https://github.com/gircore/gir.core/blob/main/src/Libs/GObject-2.0/Public/SignalTSender.cs)
 - [SignalTSenderTSignalArgs.cs](https://github.com/gircore/gir.core/blob/main/src/Libs/GObject-2.0/Public/SignalTSenderTSignalArgs.cs)
 - [ReturningSignalTSender.cs](https://github.com/gircore/gir.core/blob/main/src/Libs/GObject-2.0/Public/ReturningSignalTSender.cs)
 - [ReturningSignalTSenderTSignalArgs.cs](https://github.com/gircore/gir.core/blob/main/src/Libs/GObject-2.0/Public/ReturningSignalTSenderTSignalArgs.cs)

fsobolev:
Closes #793

### DIFF
--- a/src/Libs/GObject-2.0/Public/ReturningSignalTSender.cs
+++ b/src/Libs/GObject-2.0/Public/ReturningSignalTSender.cs
@@ -32,31 +32,31 @@ public class ReturningSignal<TSender, TReturn> : SignalDefinition
     /// <summary>
     /// Connects a <paramref name="signalHandler"/> to this signal.
     /// </summary>
-    /// <param name="o">The object on which connect the handler.</param>
+    /// <param name="sender">The object on which connect the handler.</param>
     /// <param name="signalHandler">The signal handler function.</param>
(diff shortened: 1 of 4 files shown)


badcel:
Hi thanks for the PR :+1: . Can you squash your two commits into one?

fsobolev:
> Can you squash your two commits into one?

done :)

badcel:
Hm I just thought about the name a bit more. The generic argument is named `TSender` perhaps the parameter should be called `sender` instead of `obj`?

Issue #756, opened 2023-01-12
Properly support GError


badcel:
Currently the GirLoader is adding a parameter in case the function can throw. The GirLoader should not care and if a function throws it should be a flag in the GirModel.

There needs to be a special renderer to render the Gerror parameter for internal API.

For public API there must be a parameter expression to convert the value into an exception if necessary.

PR #798, opened 2023-02-11
GError support

cameronwhite:
- Add a `throws` flag to the model, replacing the manually added parameter
- Generate an `out ErrorOwnedHandle` parameter for internal functions
- For public functions, throw an exception if an error was returned
- Skip properties where the getter or setter throw. We could implement this, but it only occurs in deprecated WebKit2WebExtension functions so it doesn't seem worthwhile
- Handle name collisions between interface methods and constructors. This occurs for `Gio.DBusConnection` and a couple other classes, which have a `new_finish` constructor but also have a `new_finish` method from `Gio.AsyncInitable`
- Remove manual bindings for `Element.MakeFromUri()` and `Pixbuf NewFromFile()` which are no longer needed.
- Add tests for functions, methods, and constructors that throw

Fixes: #756


cameronwhite:
I think this is working well now - one future TODO item I noticed is that callbacks can also have a `throws` flag (https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/docs/gir-1.2.rnc#L428) if they have a `GError **` parameter

src/Generation/GirModel/Constructor.cs
```diff
@@ -8,5 +8,6 @@ public interface Constructor
     ReturnType ReturnType { get; }
     string CIdentifier { get; }
     IEnumerable<Parameter> Parameters { get; }
+    public bool Throws { get; }
```
badcel, line 11: Idea to streamline the code a bit:

The [rnc file](https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/docs/gir-1.2.rnc#L501) declares an element called `Callable`. I think we should introduce an interface called `Callable` in the GirModel project to have a common base for `Method`, `Constructor` and `Function`. This interface would probably look something like:

```CS
public interface Callable
{
  string Name { get; }
  string CIdentifier { get; }
  bool Throws { get; }
  IEnumerable<Parameter> Parameters { get; }
  InstanceParameter? InstanceParameter { get; }
}
```

Then `GirModel.Function`, `GirModel.Method` and `GirModel.Constructor` could inherit this interface. The code would need to implement the new interface explicitly.

Additionally the `Method` interface would "hide" the `InstanceParameter` property:

```CS
public interface Method
{
  ...
  new InstanceParameter InstanceParameter { get; }
  ...
}
```

This allows us to be sure that methods will always have an instance parameter. `Callables` on the other hand might have an instance parameter or not.

Now you can rewrite `Generator.Renderer.Internal.Errors.Render` to accept a `Callable` and detect automatically what to do in regards to the comma.

Additionally I would create a similar renderer to the internal one for the public namespace to replace this:

```CS
if (function.Throws)
{
  string separator = parameters.Any() ? ", " : string.Empty;
  call.Append($"{separator}out var error");
}
```

with something like:
```CS
call.Append(Error.Render(function));
//OR
call.Append(Error.Render(method));
//OR
call.Append(Error.Render(constructor));
```

src/Generation/Generator/Renderer/Internal/Errors.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Generator.Model;
+
+namespace Generator.Renderer.Internal;
+
+internal static class Errors
```
badcel, line 8: Rename your `Errors` renderer to `Error`. The other renderers are named in plural form because their `Render` method accepts an `IEnumerable` of the corresponding class / interface. There are other renderer which are named singular, too.

src/Native/GirTestLib/girtest-error-tester.c
```diff
@@ -0,0 +1,99 @@
...
+ * @error: Return location for a potential error.
+ *
+ * Creates a new `GirTestErrorTester`, or produces an error depending on the boolean parameter.
+ *
+ * Returns: (transfer full) (nullable): The newly created `GirTestErrorTester`, on success.
+ */
+GirTestErrorTester*
+girtest_error_tester_new_can_fail (gboolean fail, GError **error)
+{
+    g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
```
badcel, line 40: In [GdkPixbuf](https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/blob/master/gdk-pixbuf/gdk-pixbuf-io.c#L1119) this returns `null` which is better I think as you are returning `NULL` in the error case, too?

` g_return_val_if_fail (error == NULL || *error == NULL, NULL);`

badcel (commented):
Looks pretty good already. Only 2 minor changes and one bigger one. But I think the bigger one improves the code a bit regarding readability and reuseablility.

src/Native/GirTestLib/girtest-error-tester.c
cameronwhite: Good catch! That was a copy-paste error and only worked because of implicit conversions - agreed it should return `NULL`.

src/Generation/GirModel/Constructor.cs
cameronwhite: That sounds like a good idea 👍 , it did feel like there was a bit too much duplicate code between the different cases

cameronwhite: Updated with those changes - I think it definitely works better, thanks for the suggestions!

src/Generation/Generator/Renderer/Internal/Error.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public static string Render(GirModel.Callable callable)
+    {
+        if (!callable.Throws)
+            return string.Empty;
+
+        // Add a separator if there are any previous parameters.
+        var separator = string.Empty;
+        if (callable.Parameters.Any() || callable.InstanceParameter is not null)
+            separator = ",";
```
badcel, line 18: The separator probably should contain the whitespace to avoid leaking a space in case there is no other parameter?

src/Generation/Generator/Renderer/Public/Class/ClassConstructors.cs
```diff
@@ -94,8 +94,14 @@ private static string RenderCallStatement(GirModel.Class cls, GirModel.Construct
         var call = new StringBuilder();
         call.Append($"var {variableName} = Internal.{cls.Name}.{Constructor.GetName(constructor)}(");
         call.Append(string.Join(", ", parameters.Select(x => x.GetCallName())));
+        if (constructor.Throws)
```
badcel, line 97: I think we can omit the `if`? In the internal namespace the renderer is called unconditionally, too. In this way the whole logic what to print is part of `Error` which is simpler.

src/Generation/Generator/Renderer/Public/Class/ClassConstructors.cs
```diff
@@ -94,8 +94,14 @@ private static string RenderCallStatement(GirModel.Class cls, GirModel.Construct
...
+            call.Append(Error.RenderParameter(constructor));
+
         call.Append(");" + Environment.NewLine);
 
+        if (constructor.Throws)
```
badcel, line 102: Remove `if`

src/Generation/Generator/Renderer/Public/Error.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public static string RenderParameter(GirModel.Callable callable)
+    {
+        if (!callable.Throws)
+            return string.Empty;
+
+        // Add a separator if there are any previous parameters.
+        var separator = string.Empty;
+        if (callable.Parameters.Any() || callable.InstanceParameter is not null)
+            separator = ",";
```
badcel, line 18: Move whitespace into separator.

src/Generation/Generator/Renderer/Public/FunctionRenderer.cs
```diff
@@ -79,8 +79,14 @@ private static string RenderCallStatement(GirModel.Function function, IReadOnlyL
 
         call.Append($"{Namespace.GetInternalName(function.Namespace)}.{parent}.{Function.GetName(function)}(");
         call.Append(string.Join(", ", parameters.Select(x => x.GetCallName())));
+        if (function.Throws)
```
badcel, line 82: Remove `if`

src/Generation/Generator/Renderer/Public/FunctionRenderer.cs
```diff
@@ -79,8 +79,14 @@ private static string RenderCallStatement(GirModel.Function function, IReadOnlyL
...
+            call.Append(Error.RenderParameter(function));
+
         call.Append(");\n");
 
+        if (function.Throws)
```
badcel, line 87: Remove `if` (add parameter `Callable` to the method and check inside wether to print or not).

src/Generation/Generator/Renderer/Public/MethodRenderer.cs
```diff
@@ -82,8 +82,14 @@ private static string RenderCallStatement(GirModel.Method method, IReadOnlyList<
         call.Append($"{Namespace.GetInternalName(method.Parent.Namespace)}.{method.Parent.Name}.{Method.GetInternalName(method)}(");
         call.Append("this.Handle" + (parameters.Any() ? ", " : string.Empty));
         call.Append(string.Join(", ", parameters.Select(x => x.GetCallName())));
+        if (method.Throws)
```
badcel, line 85: Remove `if`

src/Generation/Generator/Renderer/Public/MethodRenderer.cs
```diff
@@ -82,8 +82,14 @@ private static string RenderCallStatement(GirModel.Method method, IReadOnlyList<
...
+            call.Append(Error.RenderParameter(method));
+
         call.Append(");\n");
 
+        if (method.Throws)
```
badcel, line 90: Remove `if`

src/Generation/GirModel/Callable.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Collections.Generic;
+
+namespace GirModel;
+
+public interface Callable
+{
+    string Name { get; }
+    string CIdentifier { get; }
+    public bool Throws { get; }
```
badcel, line 9: Remove `public`

cameronwhite:
Thanks! Updated with the latest changes

Issue #791, opened 2023-02-05
Verify handling of string return parameters on Windows


cameronwhite:
Doing some preliminary testing on Windows I noticed some crashes related to methods returning a string, e.g. `GdkPixbuf.Internal.PixbufFormat.GetName()`

On Windows it looks like the default marshalling for string return values expects the string to have been allocated with `CoTaskMemAlloc()`, so there is a mismatch of allocators when it frees the native return value (https://github.com/dotnet/runtime/issues/13140)
I suspect we'll probably just have to do the marshalling manually and free using `g_free()` if there is an ownership transfer to make sure the same allocator is being used.

badcel:
We should add some GirTest code to verify string behavior on all 3 platforms:

- string as return value
- string as in parameter
- String as out parameter
- string as in/out parameter

For those cases we need to check which kind of ownership transfer makes sense and multiply the tests with the number of useful ownership transfers.

Each of the detected use cases must be checked for either UTF8 string and platform string. See https://github.com/gircore/gir.core/blob/main/src/Generation/GirLoader/Output/String.cs

cameronwhite:
For filename strings, I think we should have an `IntPtr` for the internal bindings and use `g_filename_from_utf8` to get a string in the correct encoding.
From https://docs.gtk.org/glib/character-set.html there is some complexity involved in determining which encoding is actually needed (e.g. the `G_FILENAME_ENCODING` environment variable can change this) so using the standard glib method seems like the safe approach.

badcel:
Closed as strings are implemented. String arrays are missing and are handled in #866.

PR #803, opened 2023-02-14
Improve handling of string parameters

cameronwhite:
- Add tests for filename and UTF8 string parameters / return values. By explicitly setting the `G_FILENAME_ENCODING` environment variable to an encoding other than UTF8, we can ensure that the tests fail if a filename string is incorrectly in UTF8 or UTF16 due to marshalling issues. On Windows this doesn't work since glib always uses UTF8, but at least some platforms will produce a failing test.
- Introduce `SafeHandle` types for UTF8 and filename-encoded native strings
- UTF8 string return types cannot use the default marshalling, since the wrong memory allocator can be used (bug https://github.com/gircore/gir.core/issues/791).
- Filename parameters and return values are manually marshalled since they need to be converted to and from the filename encoding using `g_filename_from_utf8()`
- UTF8 string parameters are also manually marshalled for better control over their lifetime. The default marshalling almost works for `in` parameters, but for the `girtest_string_tester_utf8_return_transfer_none()` test case the default marshalled string would be freed before the return value can be converted back to a managed string. For `out` parameters we'll likely need to manually marshal anyways for issue #791.
- Don't support callback return values that are a string with transfer=none. This doesn't work well since we need to return a newly-allocated UTF-8 string, which would then have no owner.
- Update manual bindings for these changes


cameronwhite:
This is getting a bit large :) so I'm thinking of implementing `inout` and `out` string parameters in a subsequent PR
These changes handle `in` parameters and returning strings from functions and callbacks

badcel (commented):
Thanks for the PR. The concept with the string handles is a great and future proof solution. In the generator I would like to keep the separation between the different string types.

src/Tests/Libs/GirTest-0.1.Tests/StringTest.cs
```diff
@@ -0,0 +1,62 @@
...
+    [TestMethod]
+    public void ReturnFilenameParameterShouldSucceed()
+    {
+        StringTester.FilenameReturnTransferFull(TestString).Should().Be(TestString);
+        StringTester.FilenameReturnTransferNone(TestString).Should().Be(TestString);
+
+        StringTester.FilenameReturnNullableTransferFull(TestString).Should().Be(TestString);
+        StringTester.FilenameReturnNullableTransferFull(null).Should().Be(null);
+
+        StringTester.FilenameReturnNullableTransferNone(TestString).Should().Be(TestString);
```
badcel, line 53: Shouldn't be the returned string different to the input string as it should get converted to regular UTF8?

cameronwhite: It should be identical here as long as the managed string's contents can be represented in the filename encoding (`ISO-8859-1` in these unit tests). 
- The input parameter is also a filename type, so the managed (UTF-16) string is converted to the filename encoding via `PlatformStringOwnedHandle.Create()` before being passed to the native function
- The return type is a `PlatformStringUnownedHandle` (filename encoding) which Is then converted back to a managed UTF-16 string

src/Libs/GObject-2.0/Public/Value.cs
```diff
@@ -206,7 +206,7 @@ public ParamSpec GetParam()
     public float GetFloat() => Internal.Value.GetFloat(Handle);
     public ulong GetFlags() => Internal.Value.GetFlags(Handle);
     public long GetEnum() => Internal.Value.GetEnum(Handle);
-    public string? GetString() => StringHelper.ToStringUtf8(Internal.Value.GetString(Handle));
+    public string GetString() => Internal.Value.GetString(Handle).ConvertToString();
```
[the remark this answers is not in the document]
cameronwhite: The native function doesn't return a nullable string so I think `string` is the correct return type here
https://docs.gtk.org/gobject/method.Value.get_string.html

src/Libs/GObject-2.0/Public/Type.cs
```diff
@@ -53,7 +53,7 @@ public Type(nuint value)
     // Print out the name of the GType
     public override string ToString()
```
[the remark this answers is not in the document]
cameronwhite: This function also shouldn't return a null string, so I'll add a check for that
https://docs.gtk.org/gobject/func.type_name.html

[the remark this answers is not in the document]
cameronwhite: Oh that's interesting, it's actually not marked as nullable in the annotations but it's documented as possibly returning null

src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/String.cs
```diff
@@ -21,12 +21,26 @@ public void Initialize(ParameterToNativeData parameter, IEnumerable<ParameterToN
...
+
         var prefix = parameter.Parameter.Direction == GirModel.Direction.Out
             ? "out "
             : string.Empty;
 
         var parameterName = Parameter.GetName(parameter.Parameter);
         parameter.SetSignatureName(parameterName);
-        parameter.SetCallName(prefix + parameterName);
+
+        string nativeVariableName = parameterName + "Native";
```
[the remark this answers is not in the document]
cameronwhite: Yeah that could be a good cleanup to do - `GetConvertedName` is also used in the other direction in `Internal.ParameterToManagedExpression` where it's actually the name of the managed variable being created from a native value

src/Generation/Generator/Renderer/Internal/ReturnTypeToNativeExpression/Converter/String.cs
```diff
@@ -5,10 +5,16 @@ namespace Generator.Renderer.Internal.ReturnTypeToNativeExpressions;
 internal class String : ReturnTypeConverter
 {
     public bool Supports(AnyType type)
-        => type.Is<GirModel.String>();
+        => type.Is<GirModel.Utf8String>();
```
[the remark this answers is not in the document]
cameronwhite: There weren't any cases of filename strings being returned from a callback so I don't think it needs to be implemented yet. This should be fine to rename to Utf8String.cs
If we do need to add support for filename strings, it would be in a separate file since different code is needed to generate a native filename string and transfer ownership to the native caller

src/Libs/GLib-2.0/Internal/PlatformStringHandle.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    public static PlatformStringNullHandle Instance = new();
+
+    private PlatformStringNullHandle() : base(true)
+    {
+    }
+
+    protected override bool ReleaseHandle()
+    {
+        throw new InvalidOperationException("It is not allowed to free a PlatformStringNullHandle");
```
badcel, line 35: use nameof: `$"It is not allowed to free a {nameof(PlatformStringNullHandle)}"`

src/Libs/GLib-2.0/Internal/PlatformStringHandle.cs
```diff
@@ -0,0 +1,69 @@
...
+
+public class PlatformStringUnownedHandle : PlatformStringHandle
+{
+    private PlatformStringUnownedHandle() : base(false)
+    {
+    }
+
+    protected override bool ReleaseHandle()
+    {
+        throw new InvalidOperationException("It is not allowed to free a PlatformStringUnownedHandle");
```
badcel, line 47: use nameof

src/Libs/GLib-2.0/Internal/Utf8StringHandle.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    public static Utf8StringNullHandle Instance = new();
+
+    private Utf8StringNullHandle() : base(true)
+    {
+    }
+
+    protected override bool ReleaseHandle()
+    {
+        throw new InvalidOperationException("It is not allowed to free a Utf8StringNullHandle");
```
badcel, line 31: use nameof

src/Libs/GLib-2.0/Internal/Utf8StringHandle.cs
```diff
@@ -0,0 +1,65 @@
...
+
+public class Utf8StringUnownedHandle : Utf8StringHandle
+{
+    private Utf8StringUnownedHandle() : base(false)
+    {
+    }
+
+    protected override bool ReleaseHandle()
+    {
+        throw new InvalidOperationException("It is not allowed to free a Utf8StringUnownedHandle");
```
badcel, line 43: use nameof

src/Libs/GObject-2.0/Public/Type.cs
```diff
@@ -51,9 +52,9 @@ public Type(nuint value)
     #endregion
 
     // Print out the name of the GType
-    public override string ToString()
+    public override string? ToString()
     {
-        return StringHelper.ToStringUtf8(Internal.Functions.TypeName(_value));
+        return Internal.Functions.TypeName(_value).ConvertToString();
```
badcel, line 57: The Typename can't be null I think? With the current code it would be correct to throw an exeption in case of a null.

I think we need some new classes to improve the handling of `null` strings:
- `NullablePlatformStringOwnedHandle` and `NonNullablePlatformStringOwnedHandle` 
-  `NullableUtf8StringOwnedHandle` and `NonNullableUtf8StringOwnedHandle`

Those classes could either return `null` or throw an exception in the `ConvertToString` method. The classes would need to be generated according to the nullable annotation.

src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/String.cs
```diff
@@ -21,12 +21,22 @@ public void Initialize(ParameterToNativeData parameter, IEnumerable<ParameterToN
...
-        parameter.SetCallName(prefix + parameterName);
+
+        string nativeVariableName = Parameter.GetConvertedName(parameter.Parameter);
+        var stringType = (GirModel.String) parameter.Parameter.AnyType.AsT0;
+
+        string createExpression = $"{Model.String.GetInternalOwnedHandleName(stringType)}.Create({parameterName})";
+        if (parameter.Parameter.Nullable)
+            createExpression = $"{parameterName} is null ? {Model.String.GetInternalNullHandleName(stringType)}.Instance : {createExpression}";
+
+        parameter.SetExpression($"{Model.String.GetInternalHandleName(stringType)} {nativeVariableName} = {createExpression};");
```
badcel, line 38: Split file into Platform / Utf8 version and use the corresponding utf8 / platform model to set the expression.

src/Generation/Generator/Renderer/Internal/ReturnTypeToNativeExpression/Converter/String.cs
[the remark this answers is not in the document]
badcel: Yes we can add it later :+1:

src/Libs/GObject-2.0/Public/Type.cs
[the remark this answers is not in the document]
badcel: I opened https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3290 to fix this upstream.

cameronwhite:
Updated with the changes - the separate `NonNullable` and `Nullable` types work pretty well and make it easier to avoid missing nullable checks
The only complexity was needing to convert between the nullable and non-nullable type or vice versa in the implementaiton of the handles

src/Libs/Gst-1.0/Public/Version.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace Gst;
+﻿using System;
```
badcel, line 1: Obsolete using.

badcel:
> The only complexity was needing to convert between the nullable and non-nullable type or vice versa in the implementaiton of the handles

I think I have a solution for this, which avoids the complexity and improves the usablity from an internal API point of view, too. I would suggest some renaming of the classes, too. Currently the handle class hierarchy looks like:

```
SafeHandle -> NullableUtf8StringHandle    -> NullableUtf8StringUnownedHandle
                                          -> NullableUtfStringOwnedHandle
           -> NonNullableUtf8StringHandle -> NonNullableUtf8StringUnownedHandle
           				  -> NonNullableUtf8StringOwnedHandle
```

My suggestion is to move it into such a structure:
```
SafeHandle -> OwnedUtf8StringHandle      -> NonNullableOwnedUtf8StringHandle
           -> UnownedUtf8StringHandle    -> NonNullableUnownedUtf8StringHandle
```

The `OwnedUtf8StringHandle` and `UnownedUtf8StringHandle` actually are `nullable` string handles. If some API accepts a "nullable owned utf8 string" it would only require an `OwnedUtf8StringHandle`. As `NonNullableOwnedUtf8StringHandle` derives from `OwnedUtf8StringHandle` the user is free to pass in either class.

On the other hand if an API requires a `NonNullableOwnedUtf8StringHandle` there is no way to pass in a `OwnedUtf8StringHandle` as it could contain null.

This is the complex code:
```CS
public static NullablePlatformStringOwnedHandle Create(string? s)
{
     var handle = new NullablePlatformStringOwnedHandle();
     if (s is not null)
     {
         var nonNullHandle = GLib.Internal.Functions.FilenameFromUtf8(GLib.Internal.NonNullableUtf8StringOwnedHandle.Create(s), -1, out _, out _, out var error);
         GLib.Error.ThrowOnError(error);

         // FilenameFromUtf8() returns a non-nullable handle, so transfer the result into our nullable handle.
         handle.SetHandle(nonNullHandle.DangerousGetHandle());
         nonNullHandle.SetHandleAsInvalid();
     }

     return handle;
 }
```

It could probably be rewritten to:
```CS
public static OwnedPlatformStringHandle Create(string? s)
{
    if(s is null)
        return OwnedPlatformStringHandle.CreateNullHandle(); 

    NonNullableOwnedPlatformStringHandle handle = GLib.Internal.Functions.FilenameFromUtf8(GLib.Internal.NonNullableUtf8StringOwnedHandle.Create(s), -1, out _, out _, out var error);
    GLib.Error.ThrowOnError(error);
    return handle;
 }
```

What do you think?

cameronwhite:
I think one oddity is that `in` parameters would now require having an OwnedHandle passed in, which is a bit weird since they just need a valid string and don't care about whether the handle owns its pointer?
In practice, the generated code is always calling `OwnedPlatformStringHandle.Create()` so it doesn't cause much of an issue.
But I think the `ConvertToString()` implementation for `UnownedPlatformStringHandle` would run into the problem of needing to "cast" itself to an OwnedHandle when calling `FilenameToUtf8()`?

badcel:
I see your point: From C perspective the owned / unowned information is irrelevant. But if PInvoke calls the C function there will always be only a pointer. So we do not pass this information for the sake of C but to explain the GObject requirements for the call. In this case the requirement is: Please provide a pointer to a string, which is owned by the caller as the string will not be freed by the function.

With the current design we could produce memory leaks as we could pass unowned strings to functions which do not transfer ownership (e.g. by manually binding some API in the wrong way as it requires checking the documentation to do the right thing).

With the new design we sattisfy both needs:
- Ensure that the nullablity requirements are met as it is not possible to pass a null pointer to a non nullable function
- Ensure that ownership requirements are met as it is not possible to pass an unowned string to a functions which does not take ownership and vice versa.

Do you agree or do I have some error in my thoughts?

badcel:
Hm you are right regarding your `UnownedPlatformStringHandle.ConvertToString` hint.

What about a small workaround (redeclaring the g_filename_to_utf8 function):

```CS
[DllImport(ImportResolver.Library, EntryPoint = "g_filename_to_utf8")]
private static extern GLib.Internal.NonNullableOwnedUtf8StringHandle FilenameToUtf8(IntPtr ptr, long len, out nuint bytesRead, out nuint bytesWritten, out GLib.Internal.ErrorOwnedHandle error);

public string? ConvertToString()
{
    if (IsInvalid)
        return null;

    NonNullableOwnedUtf8StringHandle result = FilenameToUtf8(DangerousGetHandle(), -1, out _, out _, out var error);
    GLib.Error.ThrowOnError(error);
    return result.ConvertToString();
}
```

badcel:
Hm I think it could work if there is some casting possible from "owned and freed by C#" to "unowned by C#, meaning it will not be freed anymore by C#". But I think this would be another pull request which must be evaluated carefully.

I think I would merge your PR in the first step do you agree?

cameronwhite:
Yeah I think using #811 to figure out any improvements sounds good :+1:  

```
If there is a function which does not transfer ownership it actually does not care if an owned or unowned pointer is provided right? As both pointers have an owner. In case of the "owned pointer" it's us, in case of the "unowned pointer" it's someone else. But C does not care if the function is called who is the owner
```
That's correct, the C function doesn't care whether we own the string ourselves or if we got the string from some other C function that returned us an UnownedHandle, so allowing either would be more consistent with the C side of things
I don't think there is any danger of memory leaks here with either proposed API

The manual binding of FilenameTo/FromUtf8 might be the better approach for avoiding any awkwardness in the internal implementation of the handles, and then we can just worry about what the more public API is?

badcel:
> The manual binding of FilenameTo/FromUtf8 might be the better approach for avoiding any awkwardness in the internal implementation of the handles, and then we can just worry about what the more public API is?

you mean you add private FilenameTo/FromUtf8 functions into your current handle implementations which does only need IntPtr parameter to avoid conversion? I think this would be a good idea and then we merge this PR.

cameronwhite:
Yep, I've added that in 👍

PR #804, opened 2023-02-14
Remove some unnecessary manual bindings

cameronwhite:
With #756 these manual bindings are no longer needed. Removing the `Filename` bindings is also useful for #791 which will introduce changes to string marshalling


badcel (approved):
Looks good to me. Thank you!

PR #807, opened 2023-02-17
Removed fallback return type renderer

### LINKED ISSUE #806: Unimplementable methods can be rendered in interfaces
While working on gtksourceview bindings, I've discovered that sometimes methods can be rendered in an interface which cannot be implemented. When this happens, the generated bindings fail to build, because the method in the interface is not implemented anywhere. Missing methods may be a problem, but I thought I'd file this issue as there is a broader issue here -- the method probably shouldn't to be rendered in the interface if it can't be implemented.

When I ran into this it happened because the return type of the method couldn't be converted from an internal to a public type (ie `ReturnTypeToManagedExpression.Render()` was throwing). This check was occurring when the method was being rendered in a class, but not when rendered in an interface.

hol430:
Fixes #806


src/Generation/Generator/Renderer/Public/ReturnType/From/Utf8StringReturnType.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Generator.Model;
+
+namespace Generator.Renderer.Public;
+
+internal class Utf8StringReturnType : RenderableScalarReturnType<GirModel.Utf8String>
```
badcel, line 5: - Please move this class into the namespace Generator.Renderer.Public.ReturnType
- Please remove the class suffix ReturnType so the name changes e.g. from BitfieldReturnType to Bitfield

src/Generation/Generator/Renderer/Public/ReturnType/From/Utf8StringReturnType.cs
```diff
@@ -0,0 +1,12 @@
...
+{
+    public override RenderableReturnType Create(GirModel.ReturnType returnType)
+    {
+        // new RenderableReturnType("string");
+        return new RenderableReturnType(Type.GetName(returnType.AnyType.AsT0));
```
badcel, line 10: Actually you can write like mentioned in your comment and remove the comment instead:

```CS
return new RenderableReturnType("string");
```

I think this is okay as it is a lot better to understand than the current cryptic expression.

But one thing is missing. The standard parameter supported the nullable annotation for return types:

`Type.GetName(type) + Nullable.Render((GirModel.Nullable) returnType)`
I suspect that it is needed here, too. So it should probably be similar to:

```CS
return new RenderableReturnType("string" + Nullable.Render(returnType));
```

src/Generation/Generator/Renderer/Public/ReturnType/From/StringArrayReturnType.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Generator.Model;
+
+namespace Generator.Renderer.Public;
+
+internal class StringArrayReturnType : RenderableArrayReturnType<GirModel.String>
+{
+    public override RenderableReturnType Create(GirModel.ReturnType returnType)
+    {
+        // Calling Supports() first guarantees that AsT1 will succeed.
```
badcel, line 9: If the interface is implemented here the comment should be obsolete. Please remove comment.

src/Generation/Generator/Renderer/Public/ReturnType/From/RenderableArrayReturnType.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Generator.Renderer.Public;
+
+internal abstract class RenderableArrayReturnType<T> : IRenderableReturnType where T : GirModel.Type
```
badcel, line 3: Please remove additional base class. Implement the `Supports` method in each class individually. It makes each concrete implementation easier to understand and to change.

src/Generation/Generator/Renderer/Public/ReturnType/From/Utf8StringReturnType.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Generator.Model;
```
badcel, line 1: - Please rename the class to `String` as it can handle both cases: `GirModel.Utf8String` and `GirModel.PlatformString` if it just supports `GirModel.String`.
- Support `GirModel.String` instead of `GirModel.PlatformString`
- Remove file "PlatformString" / "PlatformStringReturnType"

src/Generation/Generator/Renderer/Public/ReturnType/From/PlatformStringReturnType.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Generator.Model;
+
+namespace Generator.Renderer.Public;
```
badcel, line 3: Please remove this file. See file "Utf8StringReturnType" or "String" for explanation.

src/Generation/Generator/Renderer/Public/ReturnType/From/RecordArrayType.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Generator.Model;
+
+namespace Generator.Renderer.Public;
+
+internal class RecordArrayReturnType : RenderableArrayReturnType<GirModel.Record>
+{
+    public override RenderableReturnType Create(GirModel.ReturnType returnType)
+    {
+        // Calling Supports() first guarantees that AsT1 will succeed.
```
badcel, line 9: If the interface is implemented here the comment should be obsolete. Please remove comment.

src/Generation/Generator/Renderer/Public/ReturnType/From/VoidReturnType.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Generator.Model;
+
+namespace Generator.Renderer.Public;
+
+internal class VoidReturnType : RenderableScalarReturnType<GirModel.Void>
+{
+    public override RenderableReturnType Create(GirModel.ReturnType returnType)
+    {
+        return new RenderableReturnType(Type.GetName(returnType.AnyType.AsT0));
+        // equivalent to: return new RenderableReturnType("void");
```
badcel, line 10: Do like your comment says:

```CS
return new RenderableReturnType("void");
```

src/Generation/Generator/Renderer/Public/ReturnType/From/RecordArrayType.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Generator.Model;
+
+namespace Generator.Renderer.Public;
+
+internal class RecordArrayReturnType : RenderableArrayReturnType<GirModel.Record>
```
badcel, line 5: - Please move this class into the namespace `Generator.Renderer.Public.ReturnType`
- Please remove the class suffix `ReturnType` so the name changes e.g. from `BitfieldReturnType` to `Bitfield`

src/Generation/Generator/Renderer/Public/ReturnType/From/StringArrayReturnType.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Generator.Model;
+
+namespace Generator.Renderer.Public;
+
+internal class StringArrayReturnType : RenderableArrayReturnType<GirModel.String>
```
badcel, line 5: - Please move this class into the namespace `Generator.Renderer.Public.ReturnType`
- Please remove the class suffix `ReturnType` so the name changes e.g. from `BitfieldReturnType` to `Bitfield`

src/Generation/Generator/Renderer/Public/ReturnType/From/VoidReturnType.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Generator.Model;
+
+namespace Generator.Renderer.Public;
+
+internal class VoidReturnType : RenderableScalarReturnType<GirModel.Void>
```
badcel, line 5: - Please move this class into the namespace `Generator.Renderer.Public.ReturnType`
- Please remove the class suffix `ReturnType` so the name changes e.g. from `BitfieldReturnType` to `Bitfield`

badcel (changes requested):
Thank you for your PR. All in all this is already pretty good :+1: 

There are a lot of remarks as I want to have the code in a pretty consistent style. Most of them are duplicates either way so you can check one file after the other. Please forgive me my strictness but I think it is for the better longterm.

Regarding your git commits, we have to clean them up a bit to have a clean git history. Before doing something ensure your latest changes are pushed to github so you have a copy of your changes in case something goes wrong with the next steps (or even manually copy the code depending on how securly you can manage git). Please do the following:

- Rebase your branch onto current `main`. For this you will probably need to add a new `upstream` remote to your repos and rebase it onto gir.core/main ([example](https://www.atlassian.com/git/tutorials/git-forks-and-upstreams), remember to use rebase instead of merge)
- Currently you have 4 commits. As all your changes are actually one we need to make one commit from your 4. The easiest way is to use some git ui and do a "Soft" reset the the most recent commit from "gir.core/main". This will stash all your changes and the commits will be gone locally. Now create one new commit with a meaningfull message.
- If you are happy with your local changes "force push" your branch to github and overwrite your existing branch on github. The PR on gir.core will update automatically afterwards. Now your local repository and the one on github should be identical again and the 4 commits should be gone and be replaced by one.

src/Generation/Generator/Renderer/Public/ReturnType/From/BitfieldReturnType.cs
```diff
@@ -2,9 +2,9 @@
 
 namespace Generator.Renderer.Public;
 
-internal static class BitfieldReturnType
+internal class BitfieldReturnType : RenderableScalarReturnType<GirModel.Bitfield>
```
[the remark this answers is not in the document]
hol430: Regarding the namespace change - there is [already](https://github.com/gircore/gir.core/blob/main/src/Generation/Generator/Renderer/Public/ReturnType/ReturnType.cs) a record called `ReturnType` in the namespace `Generator.Renderer.Public`.

Actually, looking at where this `ReturnType` record is accessed from, I wonder if it belongs in this directory(/namespace) at all (should it be in the `Model` directory?).

src/Generation/Generator/Renderer/Public/ReturnType/From/IRenderableReturnType.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Generator.Renderer.Public;
+
+internal interface IRenderableReturnType
+{
+    RenderableReturnType Create(GirModel.ReturnType returnType);
+    bool Supports(GirModel.ReturnType returnType);
```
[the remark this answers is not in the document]
hol430: Yeah, I wasn't really sure what was best here. On one hand, having the extra info from the `ReturnType` object could be useful in some cases. On the other, there is perhaps a slight cognitive overhead associated with all of the `.AsT0` and `.AsT1` littered throughout the various implementations of the `Create()` method. And there's also the fact that mistakes or inconsistencies between the implementation of `Supports()` and `Create()` would lead to problems which get revealed at runtime, rather than compile-time.

But probably not super important either way.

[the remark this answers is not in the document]
badcel: Yeah there could be an inconsistency as both methods need to match each other.

If there is some bug reported during runtime I would suggest to write a unit test in this moment which makes this problem visible and afterwards fix the code so the problem should not be able to return.

Writing such tests beforehand is just to much work if the project is done in spare time (at least for me). Therefore it is okay for me to omit tests here as the classes are pretty simple in itself.

badcel (changes requested):
All in all this is nearly done, good work!

One class needs to be moved into another namespace. The rest are only stylistic changes.

Please rebase on current main and keep only one commit.

badcel (approved):
Looks good, thank you for your contribution.

PR #816, opened 2023-02-25
Update manual Cairo bindings for string parameters

cameronwhite:
This is largely just a refactoring to make the manual bindings consistent with the generated bindings after #803, using the new string handle classes

For string return values this adds some more type safety over IntPtr, and fixes a couple nullable warnings from the usage of `StringHelper.ToStringUtf8()`. `FontOptions.Variations` should have actually been a nullable string, while `ToyFontFace.Family` is correct to be a non-nullable string.


badcel:
Thank you!

PR #818, opened 2023-02-26
Support string parameters with direction=out

cameronwhite:
- The internal functions have a signature such as `out NonNullablePlatformStringUnownedHandle`, corresponding to the native signatures like `const gchar **`

- Introduce a "post-call" expression for parameters, which is used to take the resulting string handle and assign `handle.ConvertToString()` to the original `out string` parameter

- As with primitive value types, optional parameters are just generated as a normal `out` parameter that the caller can ignore with `out var _` if desired

- A direction of inout, or caller-allocates=1, is still not supported due to the issues noted in `ParameterToNativeExpression/Converter/Utf8String.cs`. These are low priority since inout string parameters only occur for deprecated functions, and `g_unichar_to_utf8` is the only function with caller-allocates=1

Bug: #791

### DIFF
--- a/src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/PlatformString.cs
+++ b/src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/PlatformString.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Text;
 using Generator.Model;
 
 namespace Generator.Renderer.Public.ParameterToNativeExpressions;
@@ -31,8 +32,11 @@ internal class PlatformString : ToNativeParameterConverter
             // Note: optional parameters are generated as regular out parameters, which the caller can ignore with 'out var _' if desired.
             parameter.SetCallName($"out var {nativeVariableName}");
 
-            // After the call, convert the resulting handle to a managed string.
-            parameter.SetPostCallExpression($"{parameterName} = {nativeVariableName}.ConvertToString();");
+            // After the call, convert the resulting handle to a managed string and free the native memory right away.
+            var expression = new StringBuilder();
+            expression.AppendLine($"{parameterName} = {nativeVariableName}.ConvertToString();");
+            expression.Append($"{nativeVariableName}.Dispose();");
+            parameter.SetPostCallExpression(expression.ToString());
         }
         else
         {
@@ -40,7 +44,7 @@ internal class PlatformString : ToNativeParameterConverter
                 ? Model.PlatformString.GetInternalNullableOwnedHandleName()
                 : Model.PlatformString.GetInternalNonNullableOwnedHandleName();
 
-            parameter.SetExpression($"var {nativeVariableName} = {ownedHandleTypeName}.Create({parameterName});");
+            parameter.SetExpression($"using var {nativeVariableName} = {ownedHandleTypeName}.Create({parameterName});");
             parameter.SetCallName(nativeVariableName);
         }
     }
--- a/src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/PlatformString.cs
+++ b/src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/PlatformString.cs
@@ -16,25 +16,32 @@ internal class PlatformString : ToNativeParameterConverter
         if (parameter.Parameter.CallerAllocates)
             throw new NotImplementedException($"{parameter.Parameter.AnyType}: String type with caller-allocates=1 not yet supported");
 
-        // TODO - the default marshalling for 'ref string' produces crashes for functions
-        // like pango_skip_space(), so custom marshalling may be required.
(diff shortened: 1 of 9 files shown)


src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/PlatformString.cs
```diff
@@ -16,25 +16,32 @@ public void Initialize(ParameterToNativeData parameter, IEnumerable<ParameterToN
...
+            // After the call, convert the resulting handle to a managed string.
+            parameter.SetPostCallExpression($"{parameterName} = {nativeVariableName}.ConvertToString();");
+        }
+        else
+        {
+            string ownedHandleTypeName = parameter.Parameter.Nullable
+                ? Model.PlatformString.GetInternalNullableOwnedHandleName()
+                : Model.PlatformString.GetInternalNonNullableOwnedHandleName();
+
+            parameter.SetExpression($"var {nativeVariableName} = {ownedHandleTypeName}.Create({parameterName});");
```
badcel, line 43: What about adding a `using` statement to free memory immediately?

```CS
parameter.SetExpression($"using var {nativeVariableName} = {ownedHandleTypeName}.Create({parameterName});");
```

src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/Utf8String.cs
```diff
@@ -16,25 +16,32 @@ public void Initialize(ParameterToNativeData parameter, IEnumerable<ParameterToN
...
+            // After the call, convert the resulting handle to a managed string.
+            parameter.SetPostCallExpression($"{parameterName} = {nativeVariableName}.ConvertToString();");
+        }
+        else
+        {
+            string ownedHandleTypeName = parameter.Parameter.Nullable
+                ? Model.Utf8String.GetInternalNullableOwnedHandleName()
+                : Model.Utf8String.GetInternalNonNullableOwnedHandleName();
+
+            parameter.SetExpression($"var {nativeVariableName} = {ownedHandleTypeName}.Create({parameterName});");
```
badcel, line 43: What about adding a `using` statement to free memory immediately?

```CS
parameter.SetExpression($"using var {nativeVariableName} = {ownedHandleTypeName}.Create({parameterName});");
```

src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/Utf8String.cs
```diff
@@ -16,25 +16,32 @@ public void Initialize(ParameterToNativeData parameter, IEnumerable<ParameterToN
...
-
-        parameter.SetExpression($"var {nativeVariableName} = {ownedHandleTypeName}.Create({parameterName});");
-        parameter.SetCallName(nativeVariableName);
+        if (parameter.Parameter.Direction == GirModel.Direction.Out)
+        {
+            // Note: optional parameters are generated as regular out parameters, which the caller can ignore with 'out var _' if desired.
+            parameter.SetCallName($"out var {nativeVariableName}");
+
+            // After the call, convert the resulting handle to a managed string.
+            parameter.SetPostCallExpression($"{parameterName} = {nativeVariableName}.ConvertToString();");
```
badcel, line 35: What about explicitly calling `Dispose` to free memory immediately?

```CS
parameter.SetPostCallExpression($"{parameterName} = {nativeVariableName}.ConvertToString();{Environment.NewLine}{nativeVariableName}.Dispose();");
```

src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/PlatformString.cs
```diff
@@ -16,25 +16,32 @@ public void Initialize(ParameterToNativeData parameter, IEnumerable<ParameterToN
...
-
-        parameter.SetExpression($"var {nativeVariableName} = {ownedHandleTypeName}.Create({parameterName});");
-        parameter.SetCallName(nativeVariableName);
+        if (parameter.Parameter.Direction == GirModel.Direction.Out)
+        {
+            // Note: optional parameters are generated as regular out parameters, which the caller can ignore with 'out var _' if desired.
+            parameter.SetCallName($"out var {nativeVariableName}");
+
+            // After the call, convert the resulting handle to a managed string.
+            parameter.SetPostCallExpression($"{parameterName} = {nativeVariableName}.ConvertToString();");
```
badcel, line 35: What about explicitly calling Dispose to free memory immediately?

```CS
parameter.SetPostCallExpression($"{parameterName} = {nativeVariableName}.ConvertToString();{Environment.NewLine}{nativeVariableName}.Dispose();");
```

badcel:
Looks good to me. I wonder if we should add explicit disposal of the unmanaged memory via calling `Dispose` (see commens)? If you agree you can add those changes in an additional commit as they are optional.

Issue #820, opened 2023-03-01
class Constants should contain only 'const' values


hhfiddelke:
In a first step the class Constants should contain only 'const' not 'static' values. 
Otherwise you can't use the values in a switch case statement. because VS expects constant values in a Case which static values aren't.

badcel:
Could you open separate issues for the different things if they are not related?

If you want to see your suggested changes sooner than later I would still suggest to open a PR yourself as I concentrate my work on the [milestones](https://github.com/gircore/gir.core/milestones) as those are already points I committed myself to. If I would do it I would take a look in the 0.7.0 release which is at least 3 month away, probably more.

cameronwhite:
I think the `uint` vs `int` issue is expected according to the gir files, e.g.
```
    <constant name="KEY_3" value="51" c:type="GDK_KEY_3">
      <source-position filename="gdk/gdkkeysyms.h" line="405"/>
      <type name="gint" c:type="gint"/>
    </constant>
```
so the constant should be generated as an `int`, but then functions like https://docs.gtk.org/gtk4/signal.EventControllerKey.key-pressed.html have a `uint` for the key value

On the C side of things this is probably less of an annoyance since more  implicit conversions / comparisons are allowed, and the constants are just `#define`s so their type isn't as explicit either (https://docs.gtk.org/gdk4/const.KEY_3.html)

PR #824, opened 2023-03-05
Render constants as 'const' rather than 'static' values

cameronwhite:
Also added some test coverage to GirTest for generating constants

Fixes: #820


badcel:
Thank you, great work as always 👍

Issue #622, opened 2022-05-08
Improve struct generation


badcel:
Differentiate struct types and render them as appropriate:

Todo
- [x] #747
- [x] Support boxed opaque records (#908)
- [x] Support unboxed opaque records (#942)
- [x] Support boxed record (#965)
  - [x] Value GetStringArray: ensure correct type before accessing pointer (#991)
  - [x] Add field access including tests (#994)
- [x] Support unboxed record (#1013)
- [x] Support foreign untyped records (#1029) 
- [x] Remove legacy record code (#1028)
- [x] Write unit tests for `GLib.VariantIter` (regular record) (#1033)
- [x] Ensure there are parameterless constructors for typed and untyped record types to be able to create new instances of them even if no constructor is defined on the C side. Add tests to ensure the constructors are present. Not needed for opaque record types as those have constructors defined. (#1002)
- [x] Render class property definitions for boxed properties in a class which are a a records (typed, opaque, foreign) (#1031)
- [x] create a new record type "foreign typed record". Those types can't have a managed handle, can't have a internal struct and can't have a parameterless constructor, have no methods, constructors, ... (they are manually implemented) but are registered in the type system (#1010)

-----
For non opaque records: Consider to somehow use `ref struct` to avoid memory allocations?

---------
Verify if we can differentiate between different types of structs to improve the overall struct handling.

E.g.: 
- Use a class in the public API for C structs which are opaque or class structs
- Use a struct in the public API for C structs which are only transporting data into / out of an object

Verify:  If we handle structs as structs could they be passed as "ref" to / from native code? How or those structs freed? Can those structs be defined as `public ref struct xxxx`?

Rust is implementing struct fields manually.

-------

Is it useful to differentiate between Gdk.ColorRef and Gdk.Color?

In this way ColorRef could have a method which converts its pointer in an actual struct.

If it is not useful it would be possible to use generated properties on the structs to access it fields without the need to introduce the actual struct to the user. The properties would hide the struct.

--------

 Verify:
- [In parameter modifier](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/in-parameter-modifier)
- [Out parameter modifier](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/out-parameter-modifier)
- [ref parameter modifier](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/ref)

------
[...]

PR #839, opened 2023-03-30
Slightly improved support for GPtrArray

hol430:

This is needed for the gtksourceview bindings (specifically, the completion API, which gets only partially rendered without this, which results in build errors).

I'm not totally certain of the internal -> public conversion so let me know if that looks sketchy. I largely copied it from the interface converter.

### DIFF
--- a/src/Generation/Generator/Renderer/Internal/Parameter/Converter/ClassGLibPtrArray.cs
+++ b/src/Generation/Generator/Renderer/Internal/Parameter/Converter/ClassGLibPtrArray.cs
@@ -10,27 +10,11 @@ public class ClassGLibPtrArray : ParameterConverter
     public RenderableParameter Convert(GirModel.Parameter parameter)
     {
         return new RenderableParameter(
-            Attribute: GetAttribute(parameter),
+            Attribute: string.Empty,
             Direction: string.Empty,
-            NullableTypeName: GetNullableTypeName(parameter),
+            NullableTypeName: Model.Type.Pointer,
             Name: Model.Parameter.GetName(parameter)
         );
-    }
-
-    private static string GetAttribute(GirModel.Parameter parameter)
-    {
-        return parameter.AnyTypeOrVarArgs.AsT0.AsT1.Length switch
-        {
-            { } length => MarshalAs.UnmanagedLpArray(sizeParamIndex: length),
-            _ => string.Empty,
-        };
-    }
-
-    private static string GetNullableTypeName(GirModel.Parameter parameter)
-    {
-        return parameter.AnyTypeOrVarArgs.AsT0.AsT1.Length is null
-            ? Model.Type.Pointer
-            : Model.Type.PointerArray;
     }
 }
 
--- a/src/Generation/Generator/Renderer/Internal/Parameter/Converter/PointerGLibPtrArray.cs
+++ b/src/Generation/Generator/Renderer/Internal/Parameter/Converter/PointerGLibPtrArray.cs
@@ -10,29 +10,11 @@ public class PointerGLibPtrArray : ParameterConverter
     public RenderableParameter Convert(GirModel.Parameter parameter)
     {
         return new RenderableParameter(
-            Attribute: GetAttribute(parameter),
+            Attribute: string.Empty,
             Direction: string.Empty,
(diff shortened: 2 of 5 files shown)


src/Generation/Generator/Renderer/Internal/ReturnType/Converter/InterfaceArray.cs
```diff
@@ -0,0 +1,14 @@
+namespace Generator.Renderer.Internal.ReturnType;
+
+internal class InterfaceArray : ReturnTypeConverter
```
badcel, line 3: As this class handles "GLibPtrArray" please name it `InterfaceGLibPtrArray` instead of `InterfaceArray`. We need to keep this name available for the regular array implementations.

src/Generation/Generator/Renderer/Public/ReturnType/Converter/InterfaceArray.cs
```diff
@@ -0,0 +1,14 @@
+using Generator.Model;
+
+namespace Generator.Renderer.Public.ReturnType;
+
+internal class InterfaceArray : ReturnTypeConverter
```
badcel, line 5: As this class handles "GLibPtrArray" please name it `InterfaceGLibPtrArray` instead of `InterfaceArray`. We need to keep this name available for the regular array implementations.

src/Generation/Generator/Renderer/Public/ReturnTypeToManagedExpression/Converter/InterfaceArray.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using Generator.Model;
+
+namespace Generator.Renderer.Public.ReturnTypeToManagedExpressions;
+
+internal class InterfaceArray : ReturnTypeConverter
```
badcel, line 6: As this class handles "GLibPtrArray" please name it `InterfaceGLibPtrArray` instead of `InterfaceArray`. We need to keep this name available for the regular array implementations.

badcel (changes requested):
Thank you for PR. In general this looks okayish to me. But this is not due to your code, but because I just discovered that there is potentially a pretty big bug (see #840).

Can you test if a method with an interface glib pointer array return type works even if some memory is leaked?

If not I think we should probably add a unit test to verify the behavior or go in the opposite direction not rendering any function returning a GLibPtrArray of interfaces. This could be done by keeping the files you added and just throw an exception in the "Convert" method. If the generator crashes after this change some "try/catch" blocks are missing: An exception in some converter should not stop the generator but just stop generation of the current element.

The build pipeline reported there is something wrong with your file encodings. Can you fix those, too?

badcel:
Hm, I think the code can't work. See the [docs](https://docs.gtk.org/glib/struct.PtrArray.html). The array is just a struct of a pointer and a length. Treating it as a regular array reads memory it probably shouldn't or can't read at all.

@cameronwhite: Do you agree?

Perhaps we should really go in the opposite direction and disable those functions. This should allow adding GtkSourceView and then we add support for GLib arrays as part of #840.

It can even be that there are more functions broken in a similar way if they use GLib arrays. I think I don't have a proper solution for them currently and we need to wait for #840.

cameronwhite:
If I'm reading these changes correctly, the internal return type currently would be `IntPtr[]`?
I agree that is incorrect and would probably result in a crash because it's not matching the C return type (which I think would be a `GPtrArray *`)

I see there are generated `Glib.Internal.PtrArray*Handle` classes which would be a correct return type. But I don't think those are quite usable yet since there isn't access to the struct members, and the owned handle doesn't know the correct free() function to use (related to #622)

hol430:
@badcel looks like totally preventing any GPtrArrays from being rendered is also problematic; there's a delegate in Gio which takes a parameter of type `GPtrArray*`; if we remove support for this parameter, the delegate doesn't get rendered but a callback which relies on the delegate still gets rendered, which results in broken bindings (same problem as we discussed the other day). What I've done here instead is to render GPtrArray in the internal functions as `IntPtr` which should actually work (previously it was `IntPtr`). Let me know if you have any better ideas.

badcel (approved):

badcel:
Looks good to me. Thank you :+1:

PR #841, opened 2023-03-31
Generate aliases

### LINKED ISSUE #757: Consider removal of GObject.Type
The gir defines that `GType` is an alias for `nuint`. Currently there is the struct `GObject.Type` which is a wrapper around a `nuint`. As the `GetGType` function is returning a `nuint` and all API is expecting a `nuint` consider removal of the custom `Type` struct? In the global Functions there are several helper Functions to work with the `nuint` values.

If the `GObject.Type` should be kept, it must be an integral part of the binding making `nuint` obsolete on the API surface. Which would need a lot of special casing in the generator. Currently it is probably not worth the effort.

Alternatively add proper support for aliases. If an alias is used for a primitive value type we can always generate a wrapper struct which is not used in the internal api but in the public api.

badcel:

Fixes #757


badcel:
@cameronwhite Do you have remarks regarding the alias support? You can see it's features in the tests [here](https://github.com/gircore/gir.core/blob/172642ca5823df9bd875e0b8a98b5160ffd8922a/src/Tests/Libs/GirTest-0.1.Tests/AliasTest.cs).

cameronwhite:
I think that looks fine to me.
I was wondering at first whether it should be more of a "strong typedef" that doesn't allow implicit conversions from the primitive type, if the user should mainly be using the alias type, but the implicit conversions are consistent with the C API. I don't have a strong opinion on this

badcel:
I made it implicit as constant values can't be of an alias type as creating an instance makes them not constant. So I fall back to their base type for rendering as constant.

The implicit cast allows to use the constant primitive value as an parameter to a function which requires the alias struct.

I wonder if it is to much "magic" as I normally don't like those implicit things happening.

cameronwhite:
That's a good point about constants, I can see that being necessary for pattern matching if there are any aliased types with constants that are useful to compare against

badcel:
There is currently only [GST_CLOCK_TIME_NONE](https://gstreamer.freedesktop.org/documentation/gstreamer/gstclock.html?gi-language=c#GST_CLOCK_TIME_NONE).

It seems to work in a switch statement but not a switch expression. Perhaps there is a way forward to solve this. Will test.

Issue #845, opened 2023-04-01
Support dealing with unhandled exceptions from signal callbacks


cameronwhite:
In GtkSharp, the bindings caught any unhandled exceptions from callbacks (e.g. https://github.com/GtkSharp/GtkSharp/blob/c6a2696d2288375c43486ba7ae2018ed027b16c5/Source/Libs/GLibSharp/SignalClosure.cs#L181 and https://github.com/GtkSharp/GtkSharp/blob/develop/Source/Libs/GLibSharp/ExceptionManager.cs) and raised an event that allowed the application to log or display errors
(Example from Pinta: https://github.com/PintaProject/Pinta/blob/5428fa04d3ca4562980e702b1bc568f9ec0b1bb7/Pinta/Main.cs#L137)

Otherwise, it seems like these unhandled exceptions will just terminate the program once they hit the native code boundary that invoked the delegate, so an outer try/catch block around `application.Run()` doesn't seem to work

Example program:
```csharp
using System;

void HandleClick(Gtk.Button sender, System.EventArgs args)
{
    throw new System.NotImplementedException();
}

var application = Gtk.Application.New("org.gir.core", Gio.ApplicationFlags.FlagsNone);
application.OnActivate += (sender, args) =>
{
    var window = Gtk.ApplicationWindow.New((Gtk.Application) sender);
    window.Title = "Gtk4 Window";
    window.SetDefaultSize(300, 300);

    var button = Gtk.Button.NewWithLabel("Press me");
    button.OnClicked += HandleClick;
    window.SetChild(button);

    window.Show();
};

try
{
    return application.Run();
}
catch (Exception e)
{
    System.Console.WriteLine("unhandled exception {0}", e);
}

return 1;

```

badcel:
This probably affects regular callbacks, too.

cameronwhite:
yep, I think it probably applies for any delegates that get passed into and then called by native code

badcel:
Fixed for signal callbacks in #861

cameronwhite:
One thing that I noticed in #862 is that return types (or out parameters) add some complexity for exception handling in callbacks. Since the exception is swallowed and then control returns to the native function, the native code needs to have a valid return value since it's proceeding as usual.
So, we probably need to set the return value to some default value if the managed callback has an unhandled exception

[...]

PR #856, opened 2023-04-19
Implement SynchronizationContext for GLib

### LINKED ISSUE #854: Support synchronization context to avoid manual code dispatch
### Discussed in https://github.com/gircore/gir.core/discussions/853

<div type='discussions-op-text'>

<sup>Originally posted by **ultimaweapon** April 18, 2023</sup>
I was tried to find the implementation of `SynchronizationContext` in this repository but not found so I'm not sure if it safe to use async on the event handler.</div>

ultimaweapon:
Closes #854

No unit tests yet. Need some feedback before adding a unit tests.

### DIFF
--- a/src/Libs/GLib-2.0/Internal/MainLoopSynchronizationContext.cs
+++ b/src/Libs/GLib-2.0/Internal/MainLoopSynchronizationContext.cs
@@ -1,0 +1,35 @@
+﻿using System;
+using System.Threading;
+
+namespace GLib.Internal;
+
+public sealed class MainLoopSynchronizationContext : SynchronizationContext
+{
+    public override SynchronizationContext CreateCopy()
+        => new MainLoopSynchronizationContext();
+
+    public override void Post(SendOrPostCallback d, object? state)
+        => ScheduleAction(() => d(state));
+
+    public override void Send(SendOrPostCallback d, object? state)
+        => throw new NotImplementedException();
+
+    private static void ScheduleAction(Action action)
+    {
+        var proxy = new SourceFuncNotifiedHandler(() =>
+        {
+            try
+            {
+                action();
+            }
+            catch (Exception ex)
+            {
+                UnhandledException.Raise(ex);
+            }
+
+            return false;
+        });
+
+        Functions.IdleAdd(Constants.PRIORITY_DEFAULT_IDLE, proxy.NativeCallback, IntPtr.Zero, proxy.DestroyNotify);
+    }
+}
--- a/src/Libs/GLib-2.0/Public/MainLoop.cs
+++ b/src/Libs/GLib-2.0/Public/MainLoop.cs
@@ -1,4 +1,6 @@
-﻿namespace GLib;
+﻿using System.Threading;
+
+namespace GLib;
 
 public sealed partial class MainLoop
 {
@@ -20,6 +22,22 @@ public sealed partial class MainLoop
     public bool IsRunning() => Internal.MainLoop.IsRunning(Handle);
 
     public void Run() => Internal.MainLoop.Run(Handle);
+
+    public void RunWithSynchronizationContext()
+    {
+        var original = SynchronizationContext.Current;
+
+        SynchronizationContext.SetSynchronizationContext(new Internal.MainLoopSynchronizationContext());
+
+        try
+        {
+            Run();
+        }
+        finally
+        {
+            SynchronizationContext.SetSynchronizationContext(original);
+        }
+    }
 
     public void Quit() => Internal.MainLoop.Quit(Handle);
 }
--- a/src/Libs/Gio-2.0/Public/Application.cs
+++ b/src/Libs/Gio-2.0/Public/Application.cs
@@ -1,4 +1,6 @@
-﻿namespace Gio;
+﻿using System.Threading;
+
+namespace Gio;
 
 public partial class Application
 {
@@ -10,5 +12,21 @@ public partial class Application
     public int Run()
     {
         return Internal.Application.Run(Handle, 0, new string[0]);
+    }
+
+    public int RunWithSynchronizationContext()
+    {
+        var original = SynchronizationContext.Current;
+
+        SynchronizationContext.SetSynchronizationContext(new GLib.Internal.MainLoopSynchronizationContext());
+
+        try
+        {
+            return Run();
+        }
+        finally
+        {
+            SynchronizationContext.SetSynchronizationContext(original);
+        }
     }
 }
--- a/src/Tests/Libs/GLib-2.0.Tests/SynchronizationContextTest.cs
+++ b/src/Tests/Libs/GLib-2.0.Tests/SynchronizationContextTest.cs
@@ -1,0 +1,68 @@
+﻿using System;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace GLib.Tests;
+
+[TestClass, TestCategory("IntegrationTest")]
+public class SynchronizationContextTest : Test
+{
+    [TestMethod]
+    public void AsyncMethodIsExecutedOnMainLoopThread()
+    {
+        var mainLoop = new MainLoop();
+        var context = mainLoop.GetContext();
+        var source = Functions.TimeoutSourceNew(1);
+        source.Attach(context);
+
+        int? resumeThread = null;
+        source.SetCallback(() =>
+        {
+            AsyncMethod(() =>
+            {
+                resumeThread = System.Threading.Thread.CurrentThread.ManagedThreadId;
+                mainLoop.Quit();
+            });
+            return false;
+        });
+
+        mainLoop.RunWithSynchronizationContext();
+
+        resumeThread.Should().Be(System.Threading.Thread.CurrentThread.ManagedThreadId);
+    }
+
+    [TestMethod]
+    public void ExceptionInAsyncMethodCanBeHandledViaUnhandledExceptionHandler()
+    {
+        var mainLoop = new MainLoop();
+        var context = mainLoop.GetContext();
+        var source = Functions.TimeoutSourceNew(1);
+        source.Attach(context);
+        source.SetCallback(() =>
+        {
+            AsyncMethod(() =>
+            {
+                throw new Exception();
+            });
+            return false;
+        });
+
+        var exceptionCaught = false;
+        UnhandledException.SetHandler(e =>
+        {
+            exceptionCaught = true;
+            mainLoop.Quit();
+        });
+
+        mainLoop.RunWithSynchronizationContext();
+
+        exceptionCaught.Should().BeTrue();
+    }
+
+    private static async void AsyncMethod(Action finish)
+    {
+        await Task.Delay(1);
+        finish();
+    }
+}


src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        if (exception is not null)
+            throw exception;
+    }
+
+    private static void ScheduleAction(Action action)
+    {
+        var data = GCHandle.Alloc(action);
+
+        Internal.Functions.IdleAdd(GLib.Constants.PRIORITY_DEFAULT_IDLE, IdleCallback, GCHandle.ToIntPtr(data), DestroyCallback);
```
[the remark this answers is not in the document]
cameronwhite: Aside from this, the `SynchronizationContext` implementation looks good to me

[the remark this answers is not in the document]
ultimaweapon: I'm not sure if `SourceFuncNotifiedHandler` is really safe from the GC. The moment that `gch.Free()` is called it might be possible for the GC to kick in immediately before we completely return to the GLib, which will cause the thunk for the delegate to be free too. It may seems like a very edge case but it is possible for the time slice of the thread to be used up before we completely return to the GLib, which will cause the OS to swap out our thread to run the other threads.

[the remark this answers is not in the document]
cameronwhite: That might be something to investigate for a separate PR and see if a test can reproduce the issue? (maybe from explicitly triggering a GC). 
If there is a fix needed it should be done in the generator (e.g. `CallbackNotifiedHandler.cs` so that all the generated callbacks get fixed

badcel:
I did not yet take a look at your code will do later.

To fix the build pipeline please run `dotnet format GirCore.sln`

src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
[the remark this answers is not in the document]
ultimaweapon: @badcel what do you think?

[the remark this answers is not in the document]
badcel: - I think calling code in `GLib` from `GLib.Internal` is not a problem as for example enumerations are generated exclusively in the public namespace and are used in the `Internal` namespace.
- I aggree with @cameronwhite. We should use the public code which is available. If there are edge cases we should not optimize here, but fix the problem in the generator. This leads to cleaner code here and fixes the problems in all other places.

Especially the code in the handlers can probably improved quite a lot. Since .NET core there were many improvements in the runtime to avoid allocations. E.g. I suppose that using [UnmanagedCallersOnlyAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.unmanagedcallersonlyattribute?view=net-6.0) could be a possible solution here. But like @cameronwhite said this is for another PR. I opened #857 to track this separately.

src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
```diff
@@ -0,0 +1,80 @@
+﻿using System;
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace GLib.Internal;
+
+public sealed class GLibSynchronizationContext : SynchronizationContext
+{
+    #region Fields
```
badcel, line 9: Please remove the regions. The contribution guidelines are not up to date. I remove them as I think if there is really a need for them it's an indicator that something is wrong with the code structure.

src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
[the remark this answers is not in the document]
ultimaweapon: `UnmanagedCallersOnlyAttribute` look promising. I'll update the code to use `SourceFuncNotifiedHandler` and fix the format once you have reviewed in a single commit.

src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    public override SynchronizationContext CreateCopy()
+        => new GLibSynchronizationContext();
+
+    public override void Post(SendOrPostCallback d, object? state)
+        => ScheduleAction(() => d(state));
+
+    public override void Send(SendOrPostCallback d, object? state)
+    {
+        using var executed = new ManualResetEvent(false);
```
badcel, line 47: Does it make sense to use [ManualResetEventSlim](https://learn.microsoft.com/de-de/dotnet/api/system.threading.manualreseteventslim?view=net-6.0)?

src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
```diff
@@ -0,0 +1,80 @@
...
+        Exception? exception = null;
+
+        ScheduleAction(() =>
+        {
+            try
+            {
+                d(state);
+            }
+            catch (Exception ex)
```
badcel, line 56: Can the catch block be removed completely? There is a [try-finally-statement](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/try-finally)?

badcel (commented):
Thank you for your contribution. This looks promising. Please see my comments.

badcel:
Please do not merge current `main` branch into this, but rebase your PR to avoid unnecessary commits.

src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
ultimaweapon: I was thinking to use `ManualResetEventSlim` in the first place but the document said:

> In .NET Core and .NET 5+, the default spin-waiting duration is short: on the order of 10s of microseconds, depending on platform and processor. If you expect wait times to be much longer than that, you can still use this class instead of ManualResetEvent (perhaps configured with less or no spin-waiting). **However, the performance benefit would likely be only marginal**.

So I decided to use `ManualResetEvent` instead because the callback will get executed only when idle and I'm not sure how long it will need to wait. Let me know if you still want to use `ManualResetEventSlim` so I can update this code.

src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
ultimaweapon: I was tried to mimic the behavior of [default](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Threading/SynchronizationContext.cs#L20) `SynchronizationContext` here by re-throw the exception that was thrown from the callback. The implementation on https://github.com/GtkSharp/GtkSharp also do the same. Let me know if you don't want this behavior.

src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
badcel: I think we can start with `ManualResetEventSlim`. Just from a gut feeling: I don't expect "much longer wait times" as the main loop has to render the ui as fast as possible to keep the UI responsive.

To ensure the behaviour i think we should have some unit tests.

src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
badcel: If a unit test confirms that omitting the catch block still works I would prefer to remove it.

By catching the exception, storing it, and throwing it again, the exception is not really "re-thrown" as it looses its stack details (see [CA2200](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2200)).

ultimaweapon: For stack trace problem we can use `ExceptionDispatchInfo` to capture and re-throw later without losing the stack trace, I just forgot this guy at the time I wrote this code.

> If a unit test confirms that omitting the catch block still works I would prefer to remove it.

Could you elaborate more on the unit test you are expected? By removing the catch block here the exception that was arise in the callback will be going to the GLib event loop instead of going back to the caller of `SynchronizationContext.Send` (the same behavior on the default implementation). Another problem with letting the exception going back to GLib is likely to cause the same issue as #845 but main problem here is we should maintain the same behavior as the default `SynchronizationContext` or not.

badcel: Actually I don't know if omitting the catch block works. This is why I asked you to confirm it's behaviour with a unit test.

I'm not very deeply into multi threading.  In case you are more deeply into this topic  perhaps you can give some more explanation in regards to your comment:

>By removing the catch block here the exception that was arise in the callback will be going to the GLib event loop instead of going back to the caller of SynchronizationContext.Send

We create some managed code, which gets executed in the glib main loop from native C code. If now this managed code throws some exception, what is expected to happen and why as C does not know anything about exceptions. Either the program runs and does "something" or it terminates itself with some signal. At least this is my experience so far with C.

ultimaweapon: `SynchronizationContext.Send` can be called from any thread. That it does is send the specified callback to execute and wait until the callback is executed. Where to execute the callback is depend on the implementation of `SynchronizationContext`. In our case it is GLib event loop. Unlike `SynchronizationContext.Post`, it must wait for the callback to execute completely before return to the caller while `Post` queue the callback and return immediately.

The default behavior of `SynchronizationContext.Send` is just execute the callback immediately without doing anything. That mean any exception from the callback will go to the caller directly. So it is make senses for us to reproduce the same behavior instead of sending exception somewhere else.

> We create some managed code, which gets executed in the glib main loop from native C code. If now this managed code throws some exception, what is expected to happen and why as C does not know anything about exceptions?

The problem here is C don't know how to handle the managed exception, which mean it is undefined behavior.

src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    #endregion
+
+    #region Methods
+
+    public override SynchronizationContext CreateCopy()
+        => new GLibSynchronizationContext();
+
+    public override void Post(SendOrPostCallback d, object? state)
+        => ScheduleAction(() => d(state));
```
badcel, line 43: Shouldn't here be a try/catch/ block to avoid crashes/undefined behavior?

ultimaweapon: Yeah I did not catch the exception here because we don't have any mechanism to do something with it and swallow is not good too because it will hide the problem from programmers.

ultimaweapon:
I'll comeback later to add unit tests so it is not ready for review yet.

badcel:
Thanks for updating the PR. Please:
- fix build errors
- run `dotnet format GirCore.sln` to fix remaining formatting errors
- squash your commits into one

I will do some in depth testing later.

ultimaweapon:
Strange. It was built successfully on my machine.

ultimaweapon:
I found the reason now. I did not run `GenerateLibs.fsx` after rebased. Now I have the same error.

badcel:
As this is a fairly new topic for me please be patient with me, while I try to familiarize with the topic and your work, as I'm going to ask questions and having comments.

- In general I think there is a unit tests missing for the case an exception is thrown.
- Do you know how to test the `Send` method? The current test only covers `Post`.

I'm not sure if the unit test you wrote is correct. I did not take a look into the internals but just thought that your implementation fixes the use case you described in your unit test. So I went ahead and removed the line `SynchronizationContext.SetSynchronizationContext(new Internal.GLibSynchronizationContext());` from `Module.Initialize` expecting your test to fail. Actually it did not fail on my system (Linux, fedora 38). Is the test failing for you if you disable the new `SynchronizationContext`? Which operating System are you using?

In regard to code style I would expect at least a comment if you call an `async` method without awaiting it. Additionally in C# there is a coding convention to suffix async methods with the word `Async`. So perhaps renaming `AsyncMethod` to `DelayActionAsync` (as it describes its behavior) is more appropriate? It would be nice if you could add a comment why you do not await the async method just before the method call.

ultimaweapon:
> Do you know how to test the Send method? The current test only covers Post.

I also don't have any idea to test this without invoking it directly. Let me know if you want this test.

> In general I think there is a unit tests missing for the case an exception is thrown.

Same answer as the above.

> So I went ahead and removed the line SynchronizationContext.SetSynchronizationContext(new Internal.GLibSynchronizationContext()); from Module.Initialize expecting your test to fail. Actually it did not fail on my system (Linux, fedora 38). Is the test failing for you if you disable the new SynchronizationContext? Which operating System are you using?

I'm using Arch Linux. Here is what happens on my machine when I removed the `GLibSynchronizationContext`:

```
----- Test Execution Summary -----

GLib.Tests.SynchronizationContextTest.WithGLibSynchronizationContextAsyncShouldResumeOnMainLoop:
    Outcome: Failed
    Error Message:
    Expected resumeThread to be 4, but found 7.
    Stack Trace:
       at FluentAssertions.Execution.LateBoundTestFramework.Throw(String message)
   at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message)
   at FluentAssertions.Execution.DefaultAssertionStrategy.HandleFailure(String message)
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
   at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args)
   at FluentAssertions.Numeric.NumericAssertions`2.Be(T expected, String because, Object[] becauseArgs)
   at GLib.Tests.SynchronizationContextTest.WithGLibSynchronizationContextAsyncShouldResumeOnMainLoop() in /home/ultimaweapon/Workspace/personal/gir.core/src/Tests/Libs/GLib-2.0.Tests/SynchronizationContextTest.cs:line 26
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)

    
Total tests: 1. Passed: 0. Failed: 1. Skipped: 0
```

cameronwhite:
It also fails reliably on my machine if I remove the synchronization context, though I don't know if it's guaranteed to always run on a different thread without it or if it might still be allowed to run on the main thread.

src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using System.Runtime.ExceptionServices;
+using System.Threading;
+
+namespace GLib.Internal;
+
+public sealed class GLibSynchronizationContext : SynchronizationContext
```
badcel, line 7: Please name the file like the class it contains. So this file should probably be named "GLibSynchronizationContext".

badcel:
I think last time I screwed somehow up. Today on my system disabling the `GLibSynchronizationContext` failed the test, too. I think the code did not get recompiled properly on my system after disabling the class.

Can you add the inverse test, too? This would be the code:

```CS
    [TestMethod]
    public void WithoutGLibSynchronizationContextAsyncShouldNotResumeOnMainLoop()
    {
        SynchronizationContext.SetSynchronizationContext(new SynchronizationContext());

        var mainLoop = new MainLoop();
        int? resumeThread = null;

        AsyncMethod(() =>
        {
            resumeThread = System.Threading.Thread.CurrentThread.ManagedThreadId;
            mainLoop.Quit();
        });

        mainLoop.Run();

        resumeThread.Should().NotBe(System.Threading.Thread.CurrentThread.ManagedThreadId);
    }
```

I wanted to go a bit into the topic of exceptions to see if we need to implement something similar to [Application.DispatcherUnhandledException](https://learn.microsoft.com/en-us/dotnet/api/system.windows.application.dispatcherunhandledexception?view=windowsdesktop-7.0). I tried to throw an exception like this:

```CS
        AsyncMethod(() =>
        {
            throw new Exception();
        });
```

But I was not able to catch it. You mentioned to "swallow" the Exception. If this could be done it should be possible to avoid crashing the runtime. If an exception occurs we could let the user gracefully handle the situation. How would you catch those exceptions?

ultimaweapon:
Yeah we need something like `Application.DispatcherUnhandledException` to handle the exception, which I think it is out of scope of this PR. I think it should be fixed by #845. Let me know what do you think.

For the inverse test there is a problem because `SynchronizationContext` is operate at the global level. That mean we need to disable parallel test execution on `GLib-2.0.Tests` otherwise it is possible for race conditions. IMO it is not worth because it will slow down a lot in the future when more tests are available and we already have a test that make sure the code is working as expected. Let me know if you want to proceed with the inverse test.

src/Libs/GLib-2.0/Internal/SynchronizationContext.cs
badcel: As the class integrates with the MainLoop how about calling it `MainLoopSynchronizationContext`?

badcel:
Yeah you are probably right we don't need to test things which are not working. It somehow defeats the purpose of tests.

In regard to exception handling I agree that there needs to be another PR but probably not for #845 but for some infrastructure classes which allow to signal an unhandled exception. I would like to land this PR first. Then your PR can use this code already and #845 can use it, too.

To be able to test this it would be nice if you could add a comment with the solution to catch an exception. If you want to, you can create this PR, too. Otherwise I would do it.

Regarding your statement that this operates on a "global" level it got me thinking. If there is no mainloop the new SynchronizationContext would probably not work?

So I wonder if the new SynchronizationContext should have a static method like "Ensure()" which sets it as the context for the current thread.

This method should probably not be called in "Module.Initialize" but perhaps in the static constructor of "Gio.Application" and "Glib.MainLoop". To enable it only if the user does something what actually runs a "MainLoop".

ultimaweapon:
> In regard to exception handling I agree that there needs to be another PR but probably not for https://github.com/gircore/gir.core/issues/845 but for some infrastructure classes which allow to signal an unhandled exception. I would like to land this PR first. Then your PR can use this code already and https://github.com/gircore/gir.core/issues/845 can use it, too.

Okay sounds good. Let me know when it is available so I can update my PR.

> To be able to test this it would be nice if you could add a comment with the solution to catch an exception. If you want to, you can create this PR, too. Otherwise I would do it.

I think I don't have enough capacity to work on this PR. AFAIK the only way to prevent the exceptions from being crossing the managed boundary is wrapping everything in the callback with try/catch something like this:

```cs
private static void ScheduleAction(Action action)
{
    var proxy = new SourceFuncNotifiedHandler(() =>
    {
        try
        {
            action();
        }
        catch (Exception ex)
        {
            // Do something with the exception.
        }

        return false;
    });

    Internal.Functions.IdleAdd(GLib.Constants.PRIORITY_DEFAULT_IDLE, proxy.NativeCallback, IntPtr.Zero, proxy.DestroyNotify);
}
```

> Regarding your statement that this operates on a "global" level it got me thinking. If there is no mainloop the new SynchronizationContext would probably not work?

Right because the callback will never be executed. I was wrong about the race condition on `SynchronizationContext` because it is per-thread but it seems like you already aware this.

I agree that we should change the way to install the `SynchronizationContext`. According to [this](https://stackoverflow.com/a/27234402/1829232) on Stack Overflow:

> In general, you shouldn't install a synchronization context on a thread you don't own. If you do place one on a thread pool thread, it should be removed before the thread is returned to the thread pool. More likely, if you're installing a synchronization context, you just shouldn't ever give the thread back (a custom synchronization context is commonly associated with a "main loop" for that thread).

I think the best place to install it is just before the call to `g_main_loop_run` and `g_application_run` then remove it when those functions is returned.

badcel:
Yeah this sounds best. But it will be hard to implement. We need to check or create some extension points somehow.

I already tried the try / catch block but for some reason it did not work. I hope it was not my ide failing me again. @cameronwhite if you have some capacity could you add some try / catch blocks down the call hierarchy and confirm that an exception is not caught?

ultimaweapon:
I just tested and found out `AppDomain.UnhandledException` has been raised before the crash, which mean the exception is not actually cross the managed boundary so this does not trigger the undefined behavior on the C side. The problem with `AppDomain.UnhandledException` is once it is triggered the application will be terminated so if we want some mechanism to report the exception and continue the execution it is not a viable choice.

> I already tried the try / catch block but for some reason it did not work.

I updated `ScheduleAction` to be the same as my above comment and it is working as expected on my machine.

> Yeah this sounds best. But it will be hard to implement. We need to check or create some extension points somehow.

Can we do it in the `Run` method on `src/Libs/GLib-2.0/Public/MainLoop.cs` something like this?

```cs
public void Run()
{
    var previousContext = SynchronizationContext.Current;

    SynchronizationContext.SetSynchronizationContext(new GLibSynchronizationContext());

    try
    {
        Internal.MainLoop.Run(Handle);
    }
    finally
    {
        SynchronizationContext.SetSynchronizationContext(previousContext);
    }
}
```

And did something similar to `src/Libs/Gio-2.0/Public/Application.cs`.

badcel:
> I updated ScheduleAction to be the same as my above comment and it is working as expected on my machine.
Thanks for testing and sharing. I will evaluate again. Perhaps there is some bug regarding compilation in Rider.

> Can we do it in the Run method on src/Libs/GLib-2.0/Public/MainLoop.cs something like this?
The problem is that the code is autogenerated and the generator does not know that special code should be executed. An easy solution would be to have a method "RunWithSynchronizationContext" and keep the original method alone.

badcel:
I tested again as we got support for `UnhandledExceptions` now.

Turns out that putting the try / catch block in the position works. The problem was in my unit test. As i did not quit the main loop in case of an exception the test was running indefinitely.

I updated the `Run` method to look like:
```CS
var previousContext = SynchronizationContext.Current;
SynchronizationContext.SetSynchronizationContext(new MainLoopSynchronizationContext());

try
{
    Internal.MainLoop.Run(Handle);
}
finally
{
    SynchronizationContext.SetSynchronizationContext(previousContext);
}
```

I updated the `MainLoopSynchronizationContext.ScheduleAction` to:
```CS
private static void ScheduleAction(Action action)
{
    var proxy = new SourceFuncNotifiedHandler(() =>
    {
        try
        {
            action();
        }
        catch(Exception ex)
        {
            UnhandledException.Raise(ex);
        }
        return false;
    });

    Internal.Functions.IdleAdd(GLib.Constants.PRIORITY_DEFAULT_IDLE, proxy.NativeCallback, IntPtr.Zero, proxy.DestroyNotify);
}
```

This test is green for me (I renamed `GLibSynchronizationContext` to `MainLoopSynchronizationContext`):

```CS
[TestMethod]
public void Test()
{
    var exceptionCaught = false;
    SynchronizationContext.SetSynchronizationContext(new MainLoopSynchronizationContext());
    var mainLoop = new MainLoop();
       
    GLib.UnhandledException.SetHandler(e =>
    {
        exceptionCaught = true;
        mainLoop.Quit();
    });

    AsyncMethod(() =>
    {
        throw new Exception();
    });

    mainLoop.Run();

    exceptionCaught.Should().BeTrue();
 }
```

There is one issue: As a workaround I set the synchronization context at the beginning. I tried to set the synchronization context only in `Run` as suggested by @ultimaweapon. But it looks like the AsyncMethod is not using the new `MainLoopSynchronizationContext` as the hidden async statemachine starts working immediately, meaning befor the call to `Run` thus missing out on the new `MainLoopSynchronizationContext` which will be set then.

From my point of view this is only a problem of the Unit-Test. Any ideas how to circumvent this problem?

If this is fixed I think this PR could be merged if all the discussed changes are applied.

badcel:
@ultimaweapon I think I worked it out. To get your PR merged please do the following:
- If #863 is merged, please  rebase your code on main as it contains some manual bindings which are needed for your PR.
- Please do not add additional commits to your PR, but amend any changes to your current commit.
- Please revert your changes in `Module.cs`
- Please rename `GLibSynchronizationContext` to `MainLoopSynchronizationContext`
- Please throw a `NotImplementedException` in `MainLoopSynchronizationContext.Send` and remove any other code in this metod as we currently don't know how to test it.
- Please add a new method (in addition to the existing `Run` method)`RunWithSynchronizationContext` in `MainLoop.cs` and `Gio.Application.cs` like you suggested [here](https://github.com/gircore/gir.core/pull/856#issuecomment-1529007805)
- Please add exception handling to the `MainLoopSynchronizationContext`:
```CS
private static void ScheduleAction(Action action)
{
    var proxy = new SourceFuncNotifiedHandler(() =>
    {
        try
        {
            action();
        }
        catch(Exception ex)
        {
            UnhandledException.Raise(ex);
        }
        return false;
    });

    Functions.IdleAdd(Constants.PRIORITY_DEFAULT_IDLE, proxy.NativeCallback, IntPtr.Zero, proxy.DestroyNotify);
}
```
- Please use the following code to test your canges:

```CS
using System;
using System.Threading.Tasks;
using FluentAssertions;
using Microsoft.VisualStudio.TestTools.UnitTesting;

namespace GLib.Tests;

[TestClass, TestCategory("IntegrationTest")]
public class SynchronizationContextTest : Test
{
    [TestMethod]
    public void AsyncMethodIsExecutedOnMainLoopThread()
    {
        var mainLoop = new MainLoop();
        var context = mainLoop.GetContext();
        var source = Functions.TimeoutSourceNew(1);
        source.Attach(context);

        int? resumeThread = null;
        source.SetCallback(() =>
        {
            AsyncMethod(() =>
            {
                resumeThread = System.Threading.Thread.CurrentThread.ManagedThreadId;
                mainLoop.Quit();
            });
            return false;
        });

        mainLoop.RunWithSynchronizationContext();

        resumeThread.Should().Be(System.Threading.Thread.CurrentThread.ManagedThreadId);
    }

    [TestMethod]
    public void ExceptionInAsyncMethodCanBeHandledViaUnhandledExceptionHandler()
    {
        var mainLoop = new MainLoop();
        var context = mainLoop.GetContext();
        var source = Functions.TimeoutSourceNew(1);
        source.Attach(context);
        source.SetCallback(() =>
        {
            AsyncMethod(() =>
            {
                throw new Exception();
            });
            return false;
        });

        var exceptionCaught = false;
        UnhandledException.SetHandler(e =>
        {
            exceptionCaught = true;
            mainLoop.Quit();
        });

        mainLoop.RunWithSynchronizationContext();

        exceptionCaught.Should().BeTrue();
    }
    
    private static async void AsyncMethod(Action finish)
    {
        await Task.Delay(1);
        finish();
    }
}

```

ultimaweapon:
Sorry for late response. I'm having some personal problems so I can't working on this until the end of next week. If you are in a hurry you can take over this PR.

ultimaweapon:
Done. Let me know if there is any problem.

badcel (approved):
Good work and thank you for your patience with me. In case you are missing more features more contributions are very welcome.

Issue #844, opened 2023-04-01
Do not print to Console


badcel:
If Gir.Core is used there are messages emitted if a type is registered.

The code is [here](https://github.com/gircore/gir.core/blob/388db75377610f5629a430c0ff66209b5d84773e/src/Libs/GObject-2.0/Internal/TypeRegistrar.cs#L59).

This should probably be changed to avoid console output. How could this be achieved?

- Prefered: Use `Debug.WriteLine` to print the message only during development of Gir.Core
- Rremove the Log as it is not very important
- Using microsoft abstractions logging interfaces is perhaps a bit heavy and adds a new dependency

PR #858, opened 2023-04-27
Avoid printing to the console when registering a GType.

cameronwhite:
This just switches to `Debug.WriteLine`, similar to other classes like `ObjectMapper`.
I also searched for other print statements and this seems to be the only one printing to the console, outside of the samples.

Fixes: #844


badcel:
Looks good. Thank you 🙏

PR #860, opened 2023-04-30
Add an event for unhandled exceptions from delegates invoked by native code

cameronwhite:
These unhandled exceptions would otherwise immediately terminate the program once they reach the native code boundary. This event lets applications do any logging / message dialogs etc and decide whether they want to terminate the program or attempt to continue.

Currently this only supports signal handlers for testing purposes, but would need to be extended to the other callback types.

Bug: #845

### DIFF
--- a/src/Libs/GLib-2.0/Public/UnhandledException.cs
+++ b/src/Libs/GLib-2.0/Public/UnhandledException.cs
@@ -1,0 +1,39 @@
+﻿using System;
+
+namespace GLib;
+
+/// <summary>
+/// Allows handlings exceptions which can't cross the native code boundary and
+/// would terminate the application.
+/// </summary>
+public static class UnhandledException
+{
+    /// <summary>
+    /// Invoked if an exception is raised which reaches the native code boundary.
+    /// This can be used by applications to handle these exceptions gracefully
+    /// without terminating the application.
+    /// </summary>
+    public static event EventHandler<UnhandledExceptionEventArgs>? Raised;
+
+    /// <summary>
+    /// Call this method to invoke the "Raised" event. If the event has no subscribers,
+    /// the application is terminated.
+    /// </summary>
+    /// <remarks>
+    /// This method is not intended to be called by application code, and should only
+    /// be called by code which catches an exception that would otherwise terminate
+    /// the application by unwinding to the native code boundary.
+    /// </remarks>
+    public static void Raise(Exception e)
+    {
+        if (Raised is null)
+        {
+            Console.Error.WriteLine($"{nameof(GLib.UnhandledException)} - unhandled exception: {e}");
+            Environment.Exit(1);
+        }
+        else
+        {
+            Raised.Invoke(null, new UnhandledExceptionEventArgs(e, false));
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Tests/Libs/GirTest-0.1.Tests/ExceptionManagerTest.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    public void SupportsSignalHandlers()
+    {
+        var tester = ReturningSignalTester.New();
+        tester.OnReturnBool += (_, _) =>
+        {
+            throw new NotImplementedException();
+        };
+
+        bool caught_exception = false;
```
badcel, line 18: Please use camel case and perhaps inverse name to `exceptionCaught`?

src/Libs/GLib-2.0/Public/ExceptionManager.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+
+namespace GLib;
+
+public static class ExceptionManager
```
badcel, line 5: I would prefer a little more descriptive class name like `UnhandledException`. The event could be named `Raised` and the method to raise could be named `Raise`.

Code would look like:

```CS
try
{
    _callback(returnValue, paramValues);
}
catch (Exception e)
{
    GLib.UnhandledException.Raise(e);
}
```

```CS
GLib.UnhandledException.Raised += (_, args) =>
{
    caught_exception = args.ExceptionObject is NotImplementedException;
};
```

As the class is part of the public API we should be generously document the code for a potential user of the library.

Possible summary for the "UnhandledException" class:
> Allows to handle exceptions which can't cross the native code boundary and would crash the runtime.

Event summary:
> Is invoked if an exception is raised which can't be transferred into managed code. This allows to handle those exceptions gracefully and avoid a runtime crash.

Method summary:
> Call this method to raise the "Raised" event. Exists the runtime if the event has no subscribers.

Method remark:
> This method is not intended for users to call directly but only for code which catches an exception which would otherwise crash the runtime as the caught exception can not cross the native code boundary.

badcel (commented):
The concept looks good to me. Please see my comments.

badcel:
I wonder if it should be an event at all? Is it useful to have more than one subscriber?

If not perhaps a simple callback should be used instead of an event?

Something similar to: `GLib.UnhandledException.SetHandler(..)` and `GLib.UnhandledException.Raise(..)`

cameronwhite:
Thanks! Updated with those changes.

That's a good question about it being an event .. this was mostly following GtkSharp's GLib.ExceptionManager, which I think in turn was probably following `AppDomain.UnhandledException`, so I guess there is some precedent there.
I suppose it might give a bit more flexibility for e.g. having one handler do logging, and a handler in another component adding some UI notification

In Pinta we only ever had one handler though, which pops open a message dialog and prints some info to the console

badcel:
I wonder if the generic handler approach enables this scenario:
- catch exception on the native side
- wrap exception in custom exception and add the closure instance.
- signal the unhandled exception 
- receive the event in the signal handler with the wrapped exception on the managed side
- throw original exception

cameronwhite:
Replaced by #861

PR #861, opened 2023-05-01
Allows to add an exception handler for unhandled exceptions

badcel:
These unhandled exceptions would otherwise immediately terminate the program once they reach the native code boundary. An exception handler lets applications do any logging / message dialogs etc and decide whether they want to terminate the program or attempt to continue.

@cameronwhite I took your initial code and updated it to not use an event. I think this is a simpler approach for now.

### DIFF
--- a/src/Tests/Libs/GirTest-0.1.Tests/UnhandledExceptionTest.cs
+++ b/src/Tests/Libs/GirTest-0.1.Tests/UnhandledExceptionTest.cs
@@ -1,0 +1,33 @@
+﻿using System;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace GirTest.Tests;
+
+[TestClass, TestCategory("BindingTest")]
+public class UnhandledExceptionTest : Test
+{
+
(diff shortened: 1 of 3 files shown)


src/Tests/Libs/GirTest-0.1.Tests/UnhandledExceptionTest.cs
```diff
@@ -0,0 +1,28 @@
...
+    public void SupportsSignalHandlers()
+    {
+        var tester = ReturningSignalTester.New();
+        tester.OnReturnBool += (_, _) =>
+        {
+            throw new NotImplementedException();
+        };
+
+        bool exceptionCaught = false;
+        GLib.UnhandledException.SetHandler(e =>
```
cameronwhite, line 20: Just a note for future changes - if tests can run in parallel we need to be careful about this when more test cases are added for other callback types
Maybe we should leave a comment in the code to mention this?

badcel: I added a comment.

PR #862, opened 2023-05-03
Add tests for callbacks.

cameronwhite:
This is good to have in general, but is also needed for #845 where we want to test exception handling for callbacks

Callbacks with `notified` scope are the only type that currently has full support for generating public bindings (see https://github.com/gircore/gir.core/blob/5e2b002aadc40dbf4eadfb86fdfcce03a0f8ef2e/src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/Callback.cs#L27)


badcel:
Looks good. Thank you :+1:

Issue #864, opened 2023-05-06
Windows 11 - Gdk-ERROR: OleInitialize failed


fsobolev:
A project using Gir.Core on Windows 11 22H2 crashes with the error:

```
Gdk-WARNING **: 16:20:58.129: COM runtime already initialized on the main thread with an incompatible apartment model
(process:21984): Gdk-ERROR **: 16:20:58.152: OleInitialize failed
```

Minimal reproducer:
```
public class Program
{
    private static int Main(string[] args)
    {
        var app = Gtk.Application.New("test", Gio.ApplicationFlags.FlagsNone);
        return app.Run(0, new string[0] {});
    }
}
```
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net7.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="GirCore.Gtk-4.0" Version="0.3.0" />
  </ItemGroup>

</Project>
```

Using net6.0 instead of 7.0, setting Windows 7/8 compatibility mode on the binary or launching the binary as admin has no effect.

Adding `Thread.CurrentThread.SetApartmentState(ApartmentState.STA);` causes crash with another error:
```
Unhandled exception. System.InvalidOperationException: Failed to set the specified COM apartment state. Current apartment state 'MTA'.
   at System.Threading.Thread.SetApartmentStateUnchecked(ApartmentState state, Boolean throwOnError)
   at System.Threading.Thread.SetApartmentState(ApartmentState state)
   at Program.Main(String[] args) in D:\GitHub\Test\Program.cs:line 7
```

GTK libs are installed using MSYS2. [C example from GTK docs](https://docs.gtk.org/gtk4/getting_started.html#basics) compiles and runs successfully on the same system.

fsobolev:
Adding `[STAThread]` attribute to Main method fixes the problem.

badcel:
Setting apartment state to unknown and STA afterwards works, too.

Consider doing this automatically during module initialization

PR #869, opened 2023-05-13
Set the main thread's apartment state when initializing GDK

cameronwhite:
This seems to be the logical place to do it since the startup error comes from GDK, and this also ensures it's done automatically without any extra steps to document for users.

This is wrapped in a check for whether we're running on Windows, to avoid warnings from the `SupportedOSPlatform` attribute on `SetApartmentState()`

I've tested on Windows with `Samples/Gtk-4.0/Window` and verified that this works

Fixes: #864


badcel (approved):
Looks good to me. Thank you very much :+1:.

PR #875, opened 2023-05-31
Sync'ed dependencies in csproj/gir/Module.cs

fogzot:
There are still some discrepancies between gir and csproj files, documented below. This changeset also normalize the format of csproj files moving metadata to the top and dependencies to the bottom.

Adw-1
- gir does not include GObject

cairo-1.0
- gir does not include GObject
- gir does not include GLib

freetype2-2.0
- gir does not include GObject
- gir does not include GLib
- Module.cs is missing

Gdk-4.0
- gir does not include GObject

GdkPixbuf-2.0
- gir does not include GObject
- gir includes GModule (that cannot be included by csproj)

Gst-1.0
- gir includes GModule (that cannot be included by csproj)

GstAudio-1.0
- gir includes GModule (that cannot be included by csproj)

GstBase-1.0
- gir includes GModule (that cannot be included by csproj)

GstPbutils-1.0
- gir does not include GObject

GstVideo-1.0
- gir does not include GObject

 I agree that my contribution may be licensed either under MIT or any version of LGPL license.

### DIFF
--- a/src/Libs/Adw-1/Adw-1.csproj
+++ b/src/Libs/Adw-1/Adw-1.csproj
@@ -1,11 +1,12 @@
 <Project Sdk="Microsoft.NET.Sdk">
-  <ItemGroup>
-    <ProjectReference Include="..\Gtk-4.0\Gtk-4.0.csproj" />
-  </ItemGroup>
-
   <PropertyGroup>
     <PackageId>GirCore.Adw-1</PackageId>
     <RootNamespace>Adw</RootNamespace>
     <Description>C# bindings for libadwaita.</Description>
   </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Gio-2.0\Gio-2.0.csproj" />
+    <ProjectReference Include="..\Gtk-4.0\Gtk-4.0.csproj" />
+  </ItemGroup>
 </Project>
--- a/src/Libs/Gdk-4.0/Gdk-4.0.csproj
+++ b/src/Libs/Gdk-4.0/Gdk-4.0.csproj
@@ -1,15 +1,15 @@
 <Project Sdk="Microsoft.NET.Sdk">
-  <ItemGroup>
-    <ProjectReference Include="..\cairo-1.0\cairo-1.0.csproj" />
-    <ProjectReference Include="..\GdkPixbuf-2.0\GdkPixbuf-2.0.csproj" />
-    <ProjectReference Include="..\Gio-2.0\Gio-2.0.csproj" />
-    <ProjectReference Include="..\GObject-2.0\GObject-2.0.csproj" />
-    <ProjectReference Include="..\Pango-1.0\Pango-1.0.csproj" />
-  </ItemGroup>
-
   <PropertyGroup>
     <PackageId>GirCore.Gdk-4.0</PackageId>
     <RootNamespace>Gdk</RootNamespace>
     <Description>C# bindings for Gdk.</Description>
   </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\cairo-1.0\cairo-1.0.csproj" />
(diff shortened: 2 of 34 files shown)


src/Libs/Adw-1/Adw-1.csproj
```diff
@@ -1,11 +1,13 @@
...
-
   <PropertyGroup>
     <PackageId>GirCore.Adw-1</PackageId>
     <RootNamespace>Adw</RootNamespace>
     <Description>C# bindings for libadwaita.</Description>
   </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Gio-2.0\Gio-2.0.csproj" />
+    <ProjectReference Include="..\GObject-2.0\GObject-2.0.csproj" />
```
badcel, line 10: Please do not add an explicit reference for GObject.

src/Libs/Gdk-4.0/Gdk-4.0.csproj
```diff
@@ -1,15 +1,16 @@
...
+    <PackageId>GirCore.Gdk-4.0</PackageId>
+    <RootNamespace>Gdk</RootNamespace>
+    <Description>C# bindings for Gdk.</Description>
+  </PropertyGroup>
+
   <ItemGroup>
     <ProjectReference Include="..\cairo-1.0\cairo-1.0.csproj" />
     <ProjectReference Include="..\GdkPixbuf-2.0\GdkPixbuf-2.0.csproj" />
     <ProjectReference Include="..\Gio-2.0\Gio-2.0.csproj" />
     <ProjectReference Include="..\GObject-2.0\GObject-2.0.csproj" />
```
badcel, line 12: Please remove explicit GObject initialization

src/Libs/Gdk-4.0/Public/Module.cs
```diff
@@ -13,14 +13,19 @@ public static void Initialize()
             return;
 
         GObject.Module.Initialize();
```
badcel, line 15: Please remove explicit GObject initialization

src/Libs/GdkPixbuf-2.0/GdkPixbuf-2.0.csproj
```diff
@@ -1,12 +1,12 @@
...
-
   <PropertyGroup>
     <PackageId>GirCore.GdkPixbuf-2.0</PackageId>
     <RootNamespace>GdkPixbuf</RootNamespace>
     <Description>C# bindings for GdkPixbuf.</Description>
   </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Gio-2.0\Gio-2.0.csproj" />
+    <ProjectReference Include="..\GObject-2.0\GObject-2.0.csproj" />
```
badcel, line 10: Please remove explicit GObject reference

src/Libs/GdkPixbuf-2.0/Public/Module.cs
```diff
@@ -10,6 +10,7 @@ public static void Initialize()
             return;
 
         GObject.Module.Initialize();
```
badcel, line 12: Please remove explicit GObject reference

badcel (changes requested):
Thank you! Very good work. There was some overhaul pretty necessary. Especially in `Gtk.Module` there was a lot of stuff which did not belong there.

In general I opted to remove explicit references to GObject if possible (in those cases it is referenced implicitly).  Otherwise every module would need to reference GObject explicitly. I think this keeps the dependencies more in line with the gir files. Please see my comments.

Regarding freetype please remove it's `GLib` dependency. As it does not define a "shared-library" in the `namespace` tag, please don't add a `Module.cs`. There would be the code to sett the DLL-Library name, which is not possible due to the missing "shared-library" attribute.

fogzot:
I applied all changes, as requested.

badcel (approved):
Looks good to me 👍. Thanks for your contribution.

PR #887, opened 2023-07-10
Added doc comments for all Module.cs files

fogzot:


badcel (commented):
Looks good to me. Only to remarks to check 👍.

ext/gir-files
badcel, line 1: Please don't update the submodule.

Issue #900, opened 2023-07-19
Missing methods that start async operations


fsobolev:
Methods such as:
* Gtk.FileDialog.Save
* Gtk.FileDialog.Open
* Gtk.FileDialog.SelectFolder
* Gtk.UriLauncher.Launch
* Gtk.ColorDialog.ChooseRgba

are missing. Although finishing methods exist (like Gtk.FileDialog.SaveFinish etc).

Gtk.FileDialog, Gtk.UriLauncher and Gtk.ColorDialog were added in GTK 4.10 and so weren't present in gir.core 0.3.0 at all.

cameronwhite:
The build log contains e.g. `Did not generate method 'FileDialog.Open': GirModel.AnyType: GirModel.Type: GirLoader.Output.Callback: Async scope not yet implemented` so there are some cases not yet handled in the generator

badcel:
Related to #893.

badcel:
I'm closing this in favor of #893 so the actual implementation for async is planned for 0.12 instead of 0.16.

PR #903, opened 2023-07-20
Add manual bindings for async functions

fsobolev:

Added:
- Gtk.FileDialog.OpenAsync
- Gtk.FileDialog.SaveAsync
- Gtk.FileDialog.SelectFolderAsync
- Gtk.FileLauncher.LaunchAsync
- Gtk.FileLauncher.OpenContainingFolderAsync
- Gtk.UriLauncher.LaunchAsync

~~I don't think that throwing an exception on error like it's done in [`WebView`](https://github.com/gircore/gir.core/blob/main/src/Libs/WebKit-6.0/Public/WebView.cs) is desired behavior, error is set even if a dialog was simply cancelled by the user, which is definitely not a critical error and I think just checking for `null`/`false` makes more sense:~~

Related: #900

### DIFF
--- a/src/Libs/Gtk-4.0/Public/FileDialog.cs
+++ b/src/Libs/Gtk-4.0/Public/FileDialog.cs
@@ -1,0 +1,104 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace Gtk;
+
+public partial class FileDialog
+{
+    //TODO: Async methods should be generated automatically (https://github.com/gircore/gir.core/issues/893)
+    public Task<Gio.File?> OpenAsync(Window parent)
+    {
+        var tcs = new TaskCompletionSource<Gio.File?>();
+
+        void Callback(IntPtr sourceObject, IntPtr res, IntPtr userData)
+        {
+            var fileValue = Internal.FileDialog.OpenFinish(sourceObject, res, out var error);
+            GLib.Error.ThrowOnError(error);
+
+            if (fileValue == IntPtr.Zero)
+            {
+                tcs.SetResult(null);
+            }
+            else
+            {
+                var value = new Gio.FileHelper(fileValue, true);
+                tcs.SetResult(value);
+            }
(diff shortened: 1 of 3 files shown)


fsobolev:
```
/home/runner/work/gir.core/gir.core/src/Libs/Gtk-4.0/Public/FileDialog.cs(1,1): error CHARSET: Fix file encoding. [/home/runner/work/gir.core/gir.core/src/Libs/Gtk-4.0/Gtk-4.0.csproj]
/home/runner/work/gir.core/gir.core/src/Libs/Gtk-4.0/Public/FileLauncher.cs(1,1): error CHARSET: Fix file encoding. [/home/runner/work/gir.core/gir.core/src/Libs/Gtk-4.0/Gtk-4.0.csproj]
/home/runner/work/gir.core/gir.core/src/Libs/Gtk-4.0/Public/UriLauncher.cs(1,1): error CHARSET: Fix file encoding. [/home/runner/work/gir.core/gir.core/src/Libs/Gtk-4.0/Gtk-4.0.csproj]
```

Ugh... What does that mean? 😓 Files are just ASCII text.

badcel:
You should be able to run `dotnet format GirCore.Libs.slnf` in folder `src` which should fix those things. If it does not work you can try to specify the `csproj` from GTK directly to only Format the GTK project.

badcel:
One more thing is necessary. If the `...Finish` method can return null (see documentation) then it is necessary to check the return value for `IntPtr.Zero` after the GLib.Error.ThrowOnError`:

Rough sample:
```CS
if(result == IntPtr.Zero)
{
tcs.SetResult(null);
}
else
{
//here original code to get and set the result in the task completion source.
}
```

fsobolev:
Added `Gtk.FileLauncher.OpenContainingFolderAsync`

badcel:
Looks good to me. Thanks for your contribution :rocket:

PR #910, opened 2023-07-27
Fix generic gobject types

### LINKED ISSUE #909: Cannot define a generic gobject type
Attempting to create/use a generic type which extends GObject.Object results in a TypeRegistrationException, due to the use of Type.FullName in the registration; for generic types, the FullName property contains lots of non-alphanumeric characters. On the native side we get an error like this:

```
GLib-GObject-CRITICAL **: type name 'GObjectTestsClassesGenericTypeRegistrationTestsGenericClass`2[[SystemInt32, SystemPrivateCoreLib, Version=7000, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[SystemInt32, SystemPrivateCoreLib, Version=7000, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]' contains invalid characters
```

hol430:

Fixes #909

### DIFF
--- a/src/Libs/GObject-2.0/Public/Object.Registration.cs
+++ b/src/Libs/GObject-2.0/Public/Object.Registration.cs
@@ -31,7 +31,14 @@ public partial class Object
     private static class FallbackRegistrationStrategy
     {
         private static string QualifyName(System.Type type)
-            => type.FullName?.Replace(".", "") ?? type.Name;
+            => type.ToString()
+                .Replace(".", string.Empty)
+                .Replace("+", string.Empty)
+                .Replace("`", string.Empty)
+                .Replace("[", "_")
+                .Replace("]", string.Empty)
+                .Replace(" ", string.Empty)
+                .Replace(",", "_");
 
         public static void RegisterSubclassRecursive(System.Type type)
         {
--- a/src/Tests/Libs/GObject-2.0.Tests/Classes/GenericTypeRegistrationTests.cs
+++ b/src/Tests/Libs/GObject-2.0.Tests/Classes/GenericTypeRegistrationTests.cs
@@ -1,0 +1,44 @@
+﻿using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace GObject.Tests.Classes;
+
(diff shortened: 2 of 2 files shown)


src/Libs/GObject-2.0/Public/Object.Registration.cs
```diff
@@ -31,7 +33,25 @@ private Type GetGTypeOrRegister(System.Type type)
     private static class FallbackRegistrationStrategy
     {
         private static string QualifyName(System.Type type)
-            => type.FullName?.Replace(".", "") ?? type.Name;
+        {
+            if (type.IsGenericType)
```
badcel, line 37: This looks a bit to complex for me. I think to fix this I think it should be possible to keep the idea of the original code (using simple `string.Replace`) but apply it to `Type.ToString()` instead of `Type.FullName`.

Something along the lines:
```CS
type.ToString()
    .Replace(".", string.Empty)
    .Replace("+", string.Empty)
    .Replace(",", "_")
    .Replace("`", string.Empty)
    .Replace("[", "_")
    .Replace("]", string.Empty);
```

src/Tests/Libs/GObject-2.0.Tests/Classes/GenericTypeRegistrationTests.cs
```diff
@@ -0,0 +1,46 @@
+using System;
+using System.Linq;
```
badcel, line 2: Obsolete using

src/Tests/Libs/GObject-2.0.Tests/Classes/GenericTypeRegistrationTests.cs
```diff
@@ -0,0 +1,46 @@
...
+using FluentAssertions;
```
badcel, line 3: Obsolete using

badcel (changes requested):
Thanks for the PR. Good catch that generics are not working. Please see my comments and fix the code format using `dotnet format GirCore.sln` in `src` folder.

badcel (approved):

badcel:
Thanks for your contribution :rocket:

PR #913, opened 2023-07-28
Enable unit tests on more platforms

badcel:


cameronwhite:
FYI for the macOS build we don't currently install any of the actual library dependencies:https://github.com/gircore/gir.core/blob/993443848c9d65d42036c3fd414131521169f37b/.github/workflows/ci.yml#L73

For this change I think `brew install cairo gdk-pixbuf` would be sufficient

badcel:
Yep same for windows.

PR #915, opened 2023-07-30
Implement `MainLoopSynchronizationContext.Send`

chairclr:

Tests still need to be written, if they are necessary.

### DIFF
--- a/src/Libs/GLib-2.0/Internal/MainLoopSynchronizationContext.cs
+++ b/src/Libs/GLib-2.0/Internal/MainLoopSynchronizationContext.cs
@@ -12,7 +12,17 @@ public sealed class MainLoopSynchronizationContext : SynchronizationContext
         => ScheduleAction(() => d(state));
 
     public override void Send(SendOrPostCallback d, object? state)
-        => throw new NotImplementedException();
+    {
+        using ManualResetEventSlim resetEvent = new ManualResetEventSlim(false);
+
+        ScheduleAction(() =>
+        {
+            d(state);
+            resetEvent.Set();
+        });
+
+        resetEvent.Wait();
+    }
 
     private static void ScheduleAction(Action action)
     {
--- a/src/Libs/GLib-2.0/Internal/MainLoopSynchronizationContext.cs
+++ b/src/Libs/GLib-2.0/Internal/MainLoopSynchronizationContext.cs
@@ -13,7 +13,7 @@ public sealed class MainLoopSynchronizationContext : SynchronizationContext
 
     public override void Send(SendOrPostCallback d, object? state)
     {
-        using ManualResetEventSlim resetEvent = new ManualResetEventSlim(false);
+        using var resetEvent = new ManualResetEventSlim(false);
 
         ScheduleAction(() =>
         {


badcel:
Thanks for your PR. We explicitly did not implement the `Send` method as we did not know how to invoke / test it (see: https://github.com/gircore/gir.core/pull/856#issuecomment-1526969582).

Do you have any insight on how to trigger the method or for which use cases it is used (perhaps even some examples / documentation)?

If this is clarified we can go ahead and add some tests to your PR and then merge it.

The tests should probably be part of: https://github.com/gircore/gir.core/blob/main/src/Tests/Libs/GLib-2.0.Tests/SynchronizationContextTest.cs

CEO-Chair:
Thanks for the swift response.

Our use case involved creating our own `ExecuteInMainContext` function, which directly called `SynchronizationContext.Post` for a blocking implementation. (see: [Launcher.cs](https://github.com/CEO-Chair/nightglow/blob/a163273a5ab8f06761e44cbbe0bd9fb3f1dc247c/src/Nightglow/Common/Launcher.cs#L54))

We acquired a reference to the GLib `MainLoopSynchronizationContext` through `SynchronizationContext.Current` inside of our `Gtk.Application.OnActivate` event. (see: [MainCommands.cs](https://github.com/CEO-Chair/nightglow/blob/a163273a5ab8f06761e44cbbe0bd9fb3f1dc247c/src/Nightglow/Commands/MainCommand.cs#L23))

I'm not exactly sure how to integrate this into the GLib tests either.

badcel:
After reading several implementations I think the correct thing to do would be to call [MainLoop.Invoke](https://docs.gtk.org/glib/method.MainContext.invoke.html) as it does the right thing as far as I understand. Your implementation is basically the last resort execution mentioned in the linked documentation which is fine but misses some optimizations mentioned there.

To test this I think it would be okay to manually invoke the `Send` method. 

@cameronwhite: any thoughts?

cameronwhite:
My understanding is that `g_main_context_invoke()` would be more optimal if this happens to be called while already on the main UI thread since it'll call the function immediately, but otherwise it's equivalent to `idle_add()`

badcel:
@ShortDevelopment I saw in your project that you implemented the [Send-Method](https://github.com/nearby-sharing/linux/blob/ef60aa5430b45e4780fe9ab2002621079f7095aa/src/GtkSynchronizationContext.cs#L15) on a synchronization context. Do you have some more insight why this method is important except calling it manually?

I wonder if this should be implemented in the way I mentioned in https://github.com/gircore/gir.core/pull/915#issuecomment-1657590175

ShortDevelopment:
@badcel I honestly did not expect someone to find this new repo that quickly 😅.
It's just a very early prototype to learn Gtk and this projection.

I'm using the "lowlevel" Tmds.DBus.Protocol library to interop with BLueZ DBus and there's a single "manual" call to `Send`:
https://github.com/tmds/Tmds.DBus/blob/1271f1a85fa2d31d4dcb3cf0221d8ce2b60cb834/src/Tmds.DBus.Protocol/DBusConnection.cs#L505-L520

This was just a quick hack to get it to work.
In a robust implementation there should be at least a shortcut (like you mentioned) to prevent deadlocks and optimize performance.

Also I just noticed I'm calling the delegate twice...

badcel:
Thanks for the quick reply.

It helps a lot to know that not some application developer but library developer would do the call manually like in the acenario you mentioned.

In this case I think there should be something done in regard to this PR.

----------
P.S.

I'm always looking for GirCore code from real world applications which could be a good starting template for newcomers to GTK. Especially your builder pattern is pretty clever and creates a nice and clean api to use.

You can find other projects using your own library pretty easy via [GitHub insights](https://github.com/gircore/gir.core/network/dependents).

badcel:
This is superseeded by #1222

PR #935, opened 2023-09-01
 Copy test image to output folder

### LINKED ISSUE #929: GdkPixbuf samples are not working
[GdkPixbuf samples](https://github.com/gircore/gir.core/tree/main/src/Samples/GdkPixbuf-2.0) are not working. The image file must be copied into the bin folder.

UrtsiSantsi:

Fixes #929  - "GdkPixbuf samples are not working"

### DIFF
--- a/src/Samples/GdkPixbuf-2.0/TestLoading/TestLoading.csproj
+++ b/src/Samples/GdkPixbuf-2.0/TestLoading/TestLoading.csproj
@@ -1,12 +1,16 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-    <PropertyGroup>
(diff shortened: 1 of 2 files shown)


badcel:
Hi, thanks for your contribution. Can you please stick to 2 spaces indentation in csproj files?

If you do the change please ensure to not create a 3rd commit. Meaning either you squash your commits into one big commit or keep two commits one for each file like it is now.

Afterwards it is good to go :rocket:

UrtsiSantsi:
TestLoading.csproj was using a mix of 2 and 4. And since it was the first file I edited, I made all use 4.
Now both of them should be using 2 spaces.

PR #936, opened 2023-09-01
Make Module class static

### LINKED ISSUE #876: Improve Module class
Almost every project has a `Module.cs` file which initializes the corresponding library. Verify if it is possible to make all the `Module` classes `static` (`public static class Module`). If so make them all static. If not check why and consider codechanges to allow it to be static.

UrtsiSantsi:
Fixes  #876


badcel:
Thank you. Looks good to me :+1:

PR #940, opened 2023-09-04
Improve "Libraries" section description

UrtsiSantsi:


badcel:
Thank you 👍

PR #953, opened 2023-09-15
Add more manual bindings for async functions

fsobolev:

Added:
* `Gdk.Clipboard.ReadTextAsync`
* `Gtk.AlertDialog.ChooseAsync`
* `Gtk.FileDialog.OpenMultipleAsync`
* `Gtk.FileDialog.SelectMultipleFoldersAsync`

Related: #900

### DIFF
--- a/src/Libs/Gtk-4.0/Public/FileDialog.cs
+++ b/src/Libs/Gtk-4.0/Public/FileDialog.cs
@@ -40,6 +40,38 @@ public partial class FileDialog
     }
 
     //TODO: Async methods should be generated automatically (https://github.com/gircore/gir.core/issues/893)
+    public Task<Gio.ListModel?> OpenMultipleAsync(Window parent)
+    {
+        var tcs = new TaskCompletionSource<Gio.ListModel?>();
+
+        var callbackHandler = new Gio.Internal.AsyncReadyCallbackAsyncHandler((sourceObject, res, data) =>
+        {
+            if (sourceObject is null)
+            {
+                tcs.SetException(new Exception("Missing source object"));
+                return;
+            }
+
+            var listValue = Internal.FileDialog.OpenMultipleFinish(sourceObject.Handle, res.Handle, out var error);
+
+            if (!error.IsInvalid)
+                tcs.SetException(new GLib.GException(error));
+            else
+                tcs.SetResult(GObject.Internal.ObjectWrapper.WrapNullableInterfaceHandle<Gio.ListModelHelper>(listValue, true));
+        });
+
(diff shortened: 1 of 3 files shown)


src/Libs/Gtk-4.0/Public/FileDialog.cs
```diff
@@ -39,6 +39,40 @@ public partial class FileDialog
...
+            }
+
+            var listValue = Internal.FileDialog.OpenMultipleFinish(sourceObject.Handle, res.Handle, out var error);
+
+            if (!error.IsInvalid)
+                tcs.SetException(new GLib.GException(error));
+            else if (listValue == IntPtr.Zero)
+                tcs.SetResult(null);
+            else
+                tcs.SetResult(new Gio.ListModelHelper(listValue, true));
```
badcel, line 62: You should not create a new `ListModel` on each call but go through `GObject.Internal.ObjectWrapper.WrapInterfaceHandle<Gio.ListModelHelper>` or `WrapNullableInterfaceHandle<Gio.ListModelHelper>` this ensures that a list model instance is reused if it is already known by the runtime.

In your specific case you should use the nullable variant as the pointer is nullable. In this case you can remove the `else if (listValue == IntPtr.Zero)` block as it will be handled for you automatically.

src/Libs/Gtk-4.0/Public/FileDialog.cs
```diff
@@ -107,4 +141,38 @@ public partial class FileDialog
...
+            }
+
+            var listValue = Internal.FileDialog.SelectMultipleFoldersFinish(sourceObject.Handle, res.Handle, out var error);
+
+            if (!error.IsInvalid)
+                tcs.SetException(new GLib.GException(error));
+            else if (listValue == IntPtr.Zero)
+                tcs.SetResult(null);
+            else
+                tcs.SetResult(new Gio.ListModelHelper(listValue, true));
```
badcel, line 165: You should not create a new `ListModel` on each call but go through `GObject.Internal.ObjectWrapper.WrapInterfaceHandle<Gio.ListModelHelper>` or `WrapNullableInterfaceHandle<Gio.ListModelHelper>` this ensures that a list model instance is reused if it is already known by the runtime.

In your specific case you should use the nullable variant as the pointer is nullable. In this case you can remove the `else if (listValue == IntPtr.Zero)` block as it will be handled for you automatically.

badcel (changes requested):
All in all this looks pretty good to me. Some minor things to fix, then it should be good to go.

badcel (approved):

badcel:
Thank you for your contribution :+1:

Issue #955, opened 2023-09-16
Improve String handling with size parameter


badcel:
Currently string arrays with a size parameter are handled automatically by the marshaller. This is probably not the right way to do it as the marshaller can't know how to handle / free the memory according to the transfer annotations (full / none / container).

carlosbet:
Yes, of course. This will cause the application to crash:

```CSharp
[DllImport("Gio", EntryPoint = "g_application_command_line_get_arguments")]
[Version("2.28")]
public static extern string[] GetArguments(nint cmdline, out int argc);
```
I do not have much experience in these matters. But I think it is impossible to marshal an array directly from unmanaged to managed code without knowing the size of the array first.

I think the trick is to set the native method to return the pointer directly. Then, thanks to the argc parameter, we create a managed array containing the pointers to the unmanaged strings. Finally, we marshal these strings into a string[] object.

This seems to work perfectly:

```CSharp
using System.Runtime.InteropServices;

[DllImport("libgio-2.0.so.0", EntryPoint = "g_application_command_line_get_arguments")]
static extern nint GetArguments(nint cmdline, out int argc);

static string[] GetArgs(nint cmdline, out int argc)
{
    var pointer = GetArguments(cmdline, out argc);
    
    var argv = new nint[argc]; // Managed IntPtr array
    var resultGetArguments = new string[argc];

    // Copies from an unmanaged memory pointer to a managed array.
    Marshal.Copy(pointer, argv, 0, argc);

    for (int i = 0; i < argc; i++)
    {
        resultGetArguments[i] = Marshal.PtrToStringAnsi(argv[i])!;
        Marshal.FreeCoTaskMem(argv[i]);
    }
    Marshal.FreeCoTaskMem(pointer);

    return resultGetArguments;
}

var application = Gtk.Application.New("org.gir.core", Gio.ApplicationFlags.HandlesCommandLine);
application.OnActivate += (sender, args) =>
{
    var window = Gtk.ApplicationWindow.New((Gtk.Application) sender);
    window.Title = "Gtk4 Window";
    window.SetDefaultSize(300, 300);
    window.Show();
};

application.OnCommandLine += (sender, signalArgs) =>
{
    //var a = signalArgs.CommandLine.GetArguments(out _);
    var a = GetArgs(signalArgs.CommandLine.Handle.DangerousGetHandle(), out _);

    foreach (string s in a)
        Console.WriteLine(s);

    return 0;
};
[...]

Issue #917, opened 2023-07-31
Do not render length parameter of string[] in public API


badcel:
Do not render length parameter of string[] in public API.

This happens currently I’m `Gio.Application.Run` which contains the number of array elements and the array itself as parameters.

Mark the length parameter via `IsArrayLengthParameter` in [ParameterToNativeData](https://github.com/gircore/gir.core/blob/main/src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/ParameterToNativeData.cs) and handle similar to [primitive value type arrays](https://github.com/gircore/gir.core/blob/a77ac8dcb58880d7c9ebeb745480fcf4ddfbfceb/src/Generation/Generator/Renderer/Public/ParameterToNativeExpression/Converter/PrimitiveValueTypeArray.cs#L50).

Ensure it is working for UTF and filenames via tests.

badcel:
This is already implemented.

PR #956, opened 2023-09-16
Add RunWithSynchronizationContext with arguments

### LINKED ISSUE #916: RunWithSynchronizationContext With Args
Can you provide a RunWithSynchronizationContext method that accepts args to pass to Application similar to regular Run? Needed so we can still use our dbus service...

fsobolev:

Closes #916

### DIFF
--- a/src/Libs/Gio-2.0/Public/Application.cs
+++ b/src/Libs/Gio-2.0/Public/Application.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading;
+﻿using System;
+using System.Threading;
 
 namespace Gio;
 
@@ -9,12 +10,7 @@ public partial class Application
         Module.Initialize();
     }
 
-    public int Run()
-    {
-        return Internal.Application.Run(Handle, 0, new string[0]);
-    }
-
-    public int RunWithSynchronizationContext()
+    public int RunWithSynchronizationContext(string[]? args)
     {
         var original = SynchronizationContext.Current;
 
@@ -22,7 +18,11 @@ public partial class Application
 
         try
         {
-            return Run();
+            if (args == null)
+            {
+                return Run(0, Array.Empty<string>());
+            }
+            return Run(args.Length, args);
         }
         finally
         {
--- a/src/Samples/Adw-1/Window/Program.cs
+++ b/src/Samples/Adw-1/Window/Program.cs
@@ -8,4 +8,4 @@ application.OnActivate += (sender, args) =>
     window.SetChild(label);
     window.Show();
 };
-return application.Run();
+return application.RunWithSynchronizationContext(null);
--- a/src/Samples/Gtk-4.0/AboutDialog/Program.cs
+++ b/src/Samples/Gtk-4.0/AboutDialog/Program.cs
@@ -6,4 +6,4 @@ application.OnActivate += (sender, args) =>
     dialog.Show();
 };
 
-return application.Run();
+return application.RunWithSynchronizationContext(null);
--- a/src/Samples/Gtk-4.0/Builder/Program.cs
+++ b/src/Samples/Gtk-4.0/Builder/Program.cs
@@ -7,4 +7,4 @@ application.OnActivate += (sender, args) =>
     dialog.OnResponse += (_, _) => application.Quit();
 };
 
-return application.Run();
+return application.RunWithSynchronizationContext(null);
--- a/src/Samples/Gtk-4.0/DrawingArea/DrawingArea.cs
+++ b/src/Samples/Gtk-4.0/DrawingArea/DrawingArea.cs
@@ -19,4 +19,4 @@ void Draw(Gtk.DrawingArea drawingArea, Cairo.Context cr, int width, int height)
     cr.MoveTo(40, 60);
     cr.ShowText("Powered by gir.core - GObject bindings for .NET");
 }
-return application.Run();
+return application.RunWithSynchronizationContext(null);
--- a/src/Samples/Gtk-4.0/Window/Program.cs
+++ b/src/Samples/Gtk-4.0/Window/Program.cs
@@ -6,4 +6,4 @@ application.OnActivate += (sender, args) =>
     window.SetDefaultSize(300, 300);
     window.Show();
 };
-return application.Run();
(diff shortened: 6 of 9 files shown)


src/Libs/Gio-2.0/Public/Application.cs
```diff
@@ -14,19 +14,21 @@ public int Run()
...
-            return Run();
+            return Run(argc, argv);
         }
         finally
         {
             SynchronizationContext.SetSynchronizationContext(original);
         }
     }
+
+    public int RunWithSynchronizationContext() => RunWithSynchronizationContext(0, new string[0]);
```
nlogozzo, line 33: `Array.Empty<string>()` would be better than `new string[0]`

nlogozzo (changes requested):

src/Libs/Gio-2.0/Public/Application.cs
fsobolev: @badcel should this be changed on line 14 as well?

badcel:
I think we should remove the parameterless method (`Run` and `RunWithSynchronizationContext`) as it is not available in the bindings, too. So let's keep the additions to a minimum.

Unfortunately the generated bindings are currently not 100% correct.

Meaning the encoding is probably not correct. I would ignore this for now as there are no length based platform arrays currently and it looks like it works more or less. (Should be fixed with #955.)

Additionally the length parameter is part of the public API. As it is declared to belong to the array it will vanish if #917 is merged. For this reason I would remove it already from the  `RunWithSynchronizationContext` to avoid breaking the API again.

The last point is that the array itself is nullable. For this reason I think we should mark it as nullable, too (see #955).

If it is null you can simply pass `0` and `Array.Empty<string>` to the current run implementation.

(See [documentation](https://docs.gtk.org/gio/method.Application.run.html)

After updating the code: Can you confirm that it works for you with `Array.Empty`, `null` and actual data as a parameter?

fsobolev:
Unfortunately there are some problems, but I guess it's not related to this PR?

| Arguments \ Flag | DefaultFlags | HandlesOpen | HandlesCommandLine |
| --- | --- | --- | --- |
| `null` | ✅ Calls `OnActivate` |  ✅ Calls `OnActivate` | ⚠️ <sup>[2]</sup> |
| `Array.Empty<string>()` | ✅ Calls `OnActivate`  | ✅ Calls `OnActivate` | ⚠️ <sup>[2]</sup> |
| `new string[] { ... }` | ✅ Closes with `This application can not open files.` | ⚠️ <sup>[1]</sup> | ⚠️ <sup>[2]</sup> |

<sup>[1]</sup> Calls `OnOpen` successfully, but trying to get the files (`e.Files`) causes crash:
```
UnhandledException - unhandled exception: System.InvalidCastException: Unable to cast object of type 'System.IntPtr' to type 'Gio.File[]'.
   at GObject.Value.Extract[T]() in /home/user/.var/distrobox/fedora/Projects/gir.core/src/Libs/GObject-2.0/Public/Value.cs:line 167
   at Gio.Application.OpenSignalArgs.get_Files() in /home/user/.var/distrobox/fedora/Projects/gir.core/src/Libs/Gio-2.0/Public/Application.Signals.Generated.cs:line 114
   at org.gircore.test.Program.<>c.<.ctor>b__3_3(Application sender, OpenSignalArgs e) in /home/user/.var/distrobox/fedora/Projects/GirTest/Program.cs:line 35
   at GObject.Signal`2.<>c__DisplayClass11_0.<Connect>b__0(Value returnValue, Value[] parameters) in /home/user/.var/distrobox/fedora/Projects/gir.core/src/Libs/GObject-2.0/Public/SignalTSenderTSignalArgs.cs:line 46
   at GObject.Closure.InternalCallback(IntPtr closure, IntPtr returnValuePtr, UInt32 nParamValues, ValueData[] paramValuesData, IntPtr invocationHint, IntPtr userData) in /home/user/.var/distrobox/fedora/Projects/gir.core/src/Libs/GObject-2.0/Public/Closure.cs:line 47
```

<sup>[2]</sup> Calls `OnCommandLine` succesfully, but trying to get arguments (`var argv = e.CommandLine.GetArguments(out var argc);`) causes crash:
```
UnhandledException - unhandled exception: System.Runtime.InteropServices.MarshalDirectiveException: Cannot marshal 'return value': Invalid managed/unmanaged type combination.
   at Gio.Internal.ApplicationCommandLine.GetArguments(IntPtr cmdline, Int32& argc)
   at Gio.ApplicationCommandLine.GetArguments(Int32& argc) in /home/user/.var/distrobox/fedora/Projects/gir.core/src/Libs/Gio-2.0/Public/ApplicationCommandLine.Methods.Generated.cs:line 30
   at org.gircore.test.Program.<>c.<.ctor>b__3_2(Application sender, CommandLineSignalArgs e) in /home/user/.var/distrobox/fedora/Projects/GirTest/Program.cs:line 29
   at GObject.ReturningSignal`3.<>c__DisplayClass11_0.<Connect>b__0(Value returnValue, Value[] parameters) in /home/user/.var/distrobox/fedora/Projects/gir.core/src/Libs/GObject-2.0/Public/ReturningSignalTSenderTSignalArgs.cs:line 50
   at GObject.Closure.InternalCallback(IntPtr closure, IntPtr returnValuePtr, UInt32 nParamValues, ValueData[] paramValuesData, IntPtr invocationHint, IntPtr userData) in /home/user/.var/distrobox/fedora/Projects/gir.core/src/Libs/GObject-2.0/Public/Closure.cs:line 47
```

badcel:
Thanks for testing things out. I will take a look at your findings later.

Can you switch the samples to use `RunWithSynchrinizarionContext(null)` as it probably should be the recommended way of doing it to avoid problems for new users.

badcel:
I could reproduce the problems with `HandleCommandLine` and added the defunct code as a sample to #955.

I could not reproduce the `HandlesOpen` issue but this is probably because I could not get the code to raise the event. From your exception it looks like an interface (`Gio.File`) should be extracted from signals args which is a `GObject.Value`. As extracting interfaces from a value is not yet supported I opened #958 for this. Could you provide some more help on how to get the sample working in issue #958?

As those points are not related to your PR it is good to go. Thank you :+1:

PR #993, opened 2024-01-06
Add Async FontDialog methods and sample

kashifsoofi:
Added manual binding for FontDialog Async methods.
- ChooseFaceAsync
- ChooseFamilyAsync
- ChooseFontAsync

Added a sample application to use `ChooseFontAsync` function.


badcel:
Thanks for your contribution. At a first glance the code looks good :+1: 

Some minor remarks:
- Is there a reason you use a regular `Button` over a [FontDialogButton](https://docs.gtk.org/gtk4/class.FontDialogButton.html) in your sample?
- Please don't add a new solution to the project, but add your sample project into the existing solution. The path inside the solution would be `Samples/Gtk-4.0/`

kashifsoofi:
> Thanks for your contribution. At a first glance the code looks good 👍
> 
> Some minor remarks:
> 
>     * Is there a reason you use a regular `Button` over a [FontDialogButton](https://docs.gtk.org/gtk4/class.FontDialogButton.html) in your sample?
> 
>     * Please don't add a new solution to the project, but add your sample project into the existing solution. The path inside the solution would be `Samples/Gtk-4.0/`
> 
>     * Please add your sample to https://github.com/gircore/gir.core/blob/main/src/GirCore.Libs.slnf, too.
> 
> 
> If you change your code please `Ammend` your code, to avoid creating additional commits.

No specific reason, I could already use the `FontDialogButton` not a real world scenario, I just wanted to show an alternate way of choosing font in tutorial.

badcel:
Can you add your sample to the [solution file](https://github.com/gircore/gir.core/blob/main/src/GirCore.sln) inside the `Samples/Gtk-4.0` virtual folder?

kashifsoofi:
> Can you add your sample to the [solution file](https://github.com/gircore/gir.core/blob/main/src/GirCore.sln) inside the `Samples/Gtk-4.0` virtual folder?

Done

badcel:
I took a detailed look into your PR and found an issue with `ChooseFontAsync`: According to the docs the finish method can return `Null`: https://docs.gtk.org/gtk4/method.FontDialog.choose_font_finish.html

Therefore I think the code must look like:

```CS
public Task<Pango.FontDescription?> ChooseFontAsync(Window parent, Pango.FontDescription? fontDescription)
{
    var tcs = new TaskCompletionSource<Pango.FontDescription?>();

    var callbackHandler = new Gio.Internal.AsyncReadyCallbackAsyncHandler((sourceObject, res, data) =>
    {
        if (sourceObject is null)
        {
            tcs.SetException(new Exception("Missing source object"));
            return;
        }

        var fontDescriptionOwnedHandle = Internal.FontDialog.ChooseFontFinish(sourceObject.Handle, res.Handle, out var error);

        if (!error.IsInvalid)
            tcs.SetException(new GLib.GException(error));
        else
        {
            if (fontDescriptionOwnedHandle.IsInvalid)
                tcs.SetResult(null);
            else
            {
                var result = new Pango.FontDescription(fontDescriptionOwnedHandle);
                tcs.SetResult(result);   
            }
        }
    });

    var initialValue = (Pango.Internal.FontDescriptionHandle?) fontDescription?.Handle ?? Pango.Internal.FontDescriptionUnownedHandle.NullHandle;
    Internal.FontDialog.ChooseFont(
        Handle,
        parent.Handle,
        initialValue,
        IntPtr.Zero,
        callbackHandler.NativeCallback,
        IntPtr.Zero
    );

    return tcs.Task;
}
```

I'm not sure if your sample can be considered complete. If a user presses cancel, it means that a "recoverable error" occurred. Which in C# is an exception. I already asked the gnome devs regarding this behaviour. They argument that "if the user is expected to select a font, not doing so is an error".

So if you press cancel in the font chooser your sample the app crashes. I'm not sure if or how you want to attribute to this behaviour or just add some comment?

One possible way to fix this would be:

```CS
buttonSelectFont.OnClicked += async (_, _) =>
{
    try
    {
        var fontDialog = Gtk.FontDialog.New();
        var selectedFont = await fontDialog.ChooseFontAsync(window, null);
        labelSelectedFont.SetLabel(selectedFont.ToString());
    }
    catch (Exception ex)
    {
        //Prints "Dismissed by user" in case of dialog cancellation
        Console.WriteLine(ex.Message);
    }
};
```

Do you have another proposal how to attribute to the possible exception?

If those points are fixed, it is good to go :+1:

kashifsoofi:
Thanks for the detailed review, I have updated as suggested.

badcel:
I think you missed the nullable annotations for the return type and the task completion source in `ChooseFontAsync`.

badcel:
Sorry to bother you even more. In the sample the nullable handling is missing, too: `labelSelectedFont.SetLabel(selectedFont?.ToString() ?? string.Empty);`

kashifsoofi:
No bother at all, done, Thanks

PR #1003, opened 2024-01-18
Enrich release handle exceptions

badcel:


badcel:
@cameronwhite: What do you thinkg about this change? It would not crash the application anymore if releasing a handle fails and write to stderror.

PR #1004, opened 2024-01-18
Release handle exception

badcel:

### DIFF
--- a/src/Libs/GObject-2.0/Internal/ObjectHandle.cs
+++ b/src/Libs/GObject-2.0/Internal/ObjectHandle.cs
@@ -17,7 +17,16 @@ public class ObjectHandle : SafeHandle
 
     protected sealed override bool ReleaseHandle()
     {
-        ObjectMapper.Unmap(handle);
-        return true;
+        try
+        {
+            ObjectMapper.Unmap(handle);
+            return true;
+        }
+        catch (Exception ex)
+        {
+            var typeName = Functions.TypeNameFromInstance(new TypeInstanceUnownedHandle(handle)).ConvertToString();
+            Console.Error.WriteLine($"Could not release instance {Handle} of type {typeName}. {ex}");
+            return false;
+        }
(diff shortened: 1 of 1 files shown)


badcel:
@cameronwhite: What do you thinkg about this change? It would not crash the application anymore if releasing a handle fails and write to stderror.

cameronwhite:
Adding a log with the type name looks like it would be helpful :+1: 
Is there a chance we'd hit a crash trying to do that though? If releasing the handle hit an AccessViolationException, we might get the same when trying to read the type name from the handle

badcel:
Hm, you are probably right.

cameronwhite:
Maybe instead we can print out the managed object's type which might be safer? The ObjectMapper / ToggleRef seems like it has a handle to the managed object

badcel:
Yeah, I would like to keep the try catch and logging in it's current place. It ensures that 'ReleaseHandle` never fails like the documentation is suggesting.

Accessing the object there is not directly possible. I don't have a good idea on how to refactor:

For example I wouldn't like to cache the type just to have it around in case of an error.

Currently there are like 3 layers: 
- ObjectHandle
- ObjectMapper
- ToggleRef

Perhaps this can be done differently so it gets natural to access the type 🤔

badcel:
Perhaps the ObjectMapper always holds a weak reference to the instance instead of the ToggleRef and the ObjectHandle contains the ToggleRef directly and calls the ObjectMapper.

badcel:
The [docs](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-accessviolationexception#accessviolationexception-and-trycatch-blocks) read like an `AccessViolationException` can't be caught by try/catch blocks.

So this change would potentially not help with https://github.com/PintaProject/Pinta/issues/674

badcel:
This is not the desired fix. The real bug is #1044.

Issue #397, opened 2021-08-07
Do not use reflection to create objects from IntPtr but SafeHandles


badcel:
Currently we use reflection to create an object from an IntPtr:

https://github.com/gircore/gir.core/blob/cc0732290999a390478602c51bafaf3650d8d006/Libs/GObject-2.0/Native/Classes/ObjectWrapper.cs#L70

https://github.com/gircore/gir.core/blob/2e2575061fca2cbdc7f359195e59ec7411ed998b/Libs/GObject-2.0/Native/Classes/BoxedWrapper.cs#L59

This is necessary because the TypeDictioanry only returns our instance type for a given GType.

If we update the Native Methods in a way that they return a concrete SafeHandle instead of an IntPtr and each SafeHandle provides a method like `CreateInstance` this method could create the concrete instance and effectively wrap itself into an object.

Currently only records use SafeHandle. For this to work classes would need to use SafeHandles, too.

One point which is not completely solved with this is that a `GValue` holding a boxed type can only return an `IntPtr` and not a SafeHandle
~~In this case we can create the concrete SafeHandle (probably via reflection) and again call `CreateInstance`~~
Another possibility is that the type dictionary does carry an action which accepts an `IntPtr` and creates the needed SafeHandle. In This way if one has the `GType` and an `IntPtr` one could get the correct action from the type dictioanry and with this the SafeHandle, which itself can create the instance. Thus avoiding reflection alltogether.

There is a special cases for GDK Event. In this case the given type does not correlate to one class, but to many different classes / events. If we can detect this scenario during generation time, we can generate all the needed code in the safe handle.
As this code is currently not generated, we need to allow the `CreateInstance` method of the SafeHandle to be customizeable.

This code should get obsolete with it:
https://github.com/gircore/gir.core/blob/395bf92f8d646dd7aff71bd79ba3829b089df5e1/src/Generation/Generator3/Generation/Record/PublicClass/PublicClassTemplate.cs#L36

https://github.com/gircore/gir.core/blob/395bf92f8d646dd7aff71bd79ba3829b089df5e1/src/Libs/GObject-2.0/Internal/Classes/BoxedWrapper.cs#L54

https://github.com/gircore/gir.core/blob/395bf92f8d646dd7aff71bd79ba3829b089df5e1/src/Libs/GObject-2.0/Internal/Classes/BoxedWrapper.cs#L44

Hint:
Boxed types have a copy/free function pair, which can be implemented via refcounting

[...]

Issue #919, opened 2023-08-02
Avoid reflection when wrapping unknown handles


badcel:
`GObject.Internal.ObjectWrapper.WrapHandle` needs to create a new instance of a class if the given handle is not known. It uses reflection to find the correct constructor.

With dotnet 7 it is possible to define static methods in interfaces. This allows all classes to implement a certain interface (e.g. `CreateableClass`) which defines a method like `Create(ptr, owned)`. The `WrapHandle` method can check if the given `T` implements this interface and then call the `Create` method which in turns calls the constructor. In this way reflection can be avoided. If the interface is not implemented throw an exception.

Create a fallback implementation for dotnet 6 which relies on reflection (similar to [GTypeProviderHelper](https://github.com/gircore/gir.core/blob/main/src/Libs/GObject-2.0/Internal/GTypeProviderHelper.cs)).

Verify if this can be applied to interfaces and records, too.

PR #1006, opened 2024-01-20
Avoid Reflection

badcel:

Fixes: #397, #919

**Before working on this there should probably be a PR which introduces a new way of registering classes and subclasses. Additionally there should be a source generator which uses this way to register the types and subclasses with the existing TypeDictionary. Subclasses can be registered in their static constructor as they can't be created by GObject before they were used in c#. If this is done there can be work done to remove reflection and the TypeDictionary. In this PR.**

Todo:
- [ ] Ensure that classes and subclass can be created with their own gtype.
- [ ] Allow users to Register their subclasses manually so all types are known to the typesystem
- [ ] Create SourceGenerator to automatically register all subclasses (optional)
- [ ] Find a way to remove the TypeDictionary and pass on the GType through constructor chain if a subclass is instantiated manually thus avoiding the lookup in the type dictionary
- [ ] Make sure that it is possible to get an instance of a subclass even if the metod returns a GObject.Object

### DIFF
--- a/src/Libs/GObject-2.0/Internal/BoxedWrapper.cs
+++ b/src/Libs/GObject-2.0/Internal/BoxedWrapper.cs
@@ -8,34 +8,7 @@ public class BoxedWrapper
 {
     public static object WrapHandle(IntPtr handle, bool ownsHandle, Type gtype)
     {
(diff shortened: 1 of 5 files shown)


badcel:
Replaced by #1118.

PR #1022, opened 2024-02-15
ListView sample with code only and ListItem template

kashifsoofi:


src/Samples/Gtk-4.0/ListView/TemplateListViewWindow.cs
```diff
@@ -0,0 +1,27 @@
...
+using GLib;
+using GObject;
+using Gtk;
+
+namespace ListViewSample;
+
+public class TemplateListViewWindow : Window
+{
+    public TemplateListViewWindow()
+        : base()
```
badcel, line 11: The call to `base()` can be removed as it is done implicitly

src/Samples/Gtk-4.0/ListView/ListView.csproj
```diff
@@ -0,0 +1,19 @@
...
+  </ItemGroup>
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Remove="**\*.ui" />
```
badcel, line 14: I think this can be removed `<None Remove="**\*.ui" />`

badcel (changes requested):
Thanks for your contribution. All in all this looks already pretty good to me. I'm no expert in `ListView` but as the sample works I think it is fine.

It would be nice if you could address my remarks, afterwards, we can land this sample :rocket:

kashifsoofi:
Done

PR #1024, opened 2024-02-21
GridView Sample with StringList and CustomObject ListStore

kashifsoofi:


badcel (changes requested):
Thank you for your contribution.

When I run the sample the first time the SVG files where not displayed. This happened because my working directory was inside the `bin/debug` folder which does not contain the `Resources` folder.

Perhaps adding something like this to the `csproj` file is a good idea as the path is a relative one:
```xml
<ItemGroup>
   <None Include="$(ProjectDir)Resources\**" CopyToOutputDirectory="PreserveNewest" />
</ItemGroup>
```

- Can you confirm that the used SVG files are allowed to be used?
- Can you add some link to the original SVG source in the source code as a comment so it is easy to find the original files?

kashifsoofi:
Project file updated, I have added README.md with links to the SVG files.
From [SVG Repo](https://www.svgrepo.com/)
> Download free SVG vectors and icons for commercial use.

PR #1068, opened 2024-05-30
Remove duplicate semicolon in return statements

adamreeve:

This is a minor tidy up to the generated code. Some of the expressions returned from `ReturnTypeConverter` implementations added a semicolon, but a semicolon is already added everywhere `ReturnTypeToManagedExpression.Render` is called.

This didn't cause any build failures but Rider would show red lines when viewing the generated code and complain about "unreachable code", referring to the empty statement before the second semicolon.


badcel:
Thank you :+1:

PR #1079, opened 2024-06-11
Add nullable annotation to nullable string array parameters in internal methods

### LINKED ISSUE #1077: Nullable string array parameter generates internal code with non-nullable parameter
I've added some test code to demonstrate this problem in this commit on my fork: https://github.com/adamreeve/gir.core/commit/a0b9be8919d271d729c877e68d56385c829c6f43

The other string array parameters in this test file have `array zero-terminated=1` annotations, but I'm generating bindings for code that uses `array length=PARAM`.

This generates an internal method like this:
```C#
[DllImport(ImportResolver.Library, EntryPoint = "girtest_string_array_tester_utf8_return_element_parameter_transfer_none_nullable_with_size")]
public static extern GLib.Internal.NullableUtf8StringOwnedHandle Utf8ReturnElementParameterTransferNoneNullableWithSize([MarshalAs(UnmanagedType.LPArray, SizeParamIndex=1)]string[] data, int dataSize, int position);
```

And a public method like this:
```c#
public static string? Utf8ReturnElementParameterTransferNoneNullableWithSize(string[]? data, int dataSize, int position)
{
    
    var resultUtf8ReturnElementParameterTransferNoneNullableWithSize = GirTest.Internal.StringArrayTester.Utf8ReturnElementParameterTransferNoneNullableWithSize(data, dataSize, position);

    
    return resultUtf8ReturnElementParameterTransferNoneNullableWithSize.ConvertToString();
}
```

The public method has a nullable parameter (`string[]? data`) but the internal method does not (`string[] data`). This causes build warnings (or errors if `TreatWarningsAsErrors` is true) at the line where the public method calls the internal method:
```
warning CS8604: Possible null reference argument for parameter 'data' in 'NullableUtf8StringOwnedHandle StringArrayTester.Utf8ReturnElementParameterTransferNoneNullableWithSize(string[] data, int dataSize, int position)'
```

Also, it would be nice if the length parameter wasn't included in the public method, but I can see there's already an issue for this (#917).

adamreeve:

This fixes #1077

I'm not super familiar with the code so let me know if there's a better way to handle this, or if you'd rather just fix it yourself :smile:

### DIFF
--- a/src/Generation/Generator/Renderer/Internal/Parameter/Converter/Utf8StringArray.cs
+++ b/src/Generation/Generator/Renderer/Internal/Parameter/Converter/Utf8StringArray.cs
@@ -25,11 +25,12 @@ internal class Utf8StringArray : ParameterConverter
     private static RenderableParameter SizeBasedArray(GirModel.Parameter parameter)
     {
         var length = parameter.AnyTypeOrVarArgs.AsT0.AsT1.Length ?? throw new Exception("Length must not be null");
+        var typeName = Model.ArrayType.GetName(parameter.AnyTypeOrVarArgs.AsT0.AsT1) + Nullable.Render(parameter);
 
         return new RenderableParameter(
             Attribute: MarshalAs.UnmanagedLpArray(sizeParamIndex: length),
             Direction: string.Empty,
-            NullableTypeName: Model.ArrayType.GetName(parameter.AnyTypeOrVarArgs.AsT0.AsT1),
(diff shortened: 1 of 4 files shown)


badcel:
The code looks fine. Can you Squash your last 2 commits into one, so you have 2 commits left? One for tests and one for the actual code?

adamreeve:
Yep sure, I've rebased this now.

PR #1086, opened 2024-06-26
Disable rendering code for repositories only required via includes

### LINKED ISSUE #1084: Skip code generation for repositories that were pulled in as includes
Continuing the discussion from https://github.com/gircore/gir.core/issues/1078#issuecomment-2187914806

Currently the GirTool generate command accepts a list of input repository files to generate code for. Any includes in these files are resolved and added to the list of output repositories by the `RepositoriesDependencyResolver` class. This is needed for generating code for interacting with dependencies, but sometimes users won't want to generate code for all dependencies but rather use them via package references, such as when using GirCore.GLib-2.0 for example.

I'd like to change the GirTool behaviour so that these includes that aren't from repositories explicitly listed in the `input` argument don't get code generated. They would still need to be loaded and resolved though.

adamreeve:

This fixes #1084

I couldn't see a good way to test this change, it would probably need a separate GitHub action job to generate code for a subset of the libraries which seems pretty heavy handed, but let me know if you think that's necessary or there's a better way.

### DIFF
--- a/src/Generation/GirTool/GenerateCommand.cs
+++ b/src/Generation/GirTool/GenerateCommand.cs
@@ -19,20 +19,21 @@ public partial class GenerateCommand : Command
         {
             Configuration.SetupLogLevel(logLevel);
 
-            var namespaces = GetNamespaces(searchPathLinux, searchPathMacos, searchPathWindows, disableAsync, input, invocationContext);
+            var (allNamespaces, generatedNamespaces) = GetNamespaces(
+                searchPathLinux, searchPathMacos, searchPathWindows, disableAsync, input, invocationContext);
 
             if (disableAsync)
             {
-                foreach (var @namespace in namespaces)
+                foreach (var @namespace in allNamespaces)
                     PlatformGenerator.Fixup(@namespace);
 
-                foreach (var @namespace in namespaces)
+                foreach (var @namespace in generatedNamespaces)
                     PlatformGenerator.Generate(@namespace, output);
             }
             else
             {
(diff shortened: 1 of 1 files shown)


src/GenerateLibs.fsx
```diff
@@ -16,6 +16,7 @@ let girFiles =
         "GdkPixbuf-2.0.gir"
         "Gio-2.0.gir"
         "GLib-2.0.gir"
+        "GModule-2.0.gir"
```
adamreeve, line 19: This was the only library that was previously getting code generated but wasn't listed explicitly here. I'm not sure whether this is really needed though as it doesn't look like this library is used as a project dependency by any of the other libraries or published as a NuGet package.

badcel: Please remove this line as there is no GModule-2.0 project / nuget. I don't know if / when we are going to need this project.

badcel:
In regard to your code changes in `GirTool` I found nothing to improve, great work :-).

Looking at the code after some pretty long time I don't like what I did there originally. It is a little bit to hard to grasp what is happening for my feeling. But it is like it is. Perhaps it will see some improvements over time.

Please remove the `GModule-2.0.gir` from the list of generated projects in `GenerateLibs.fsx`, it is currently not needed. (Just remove your last commit to avoid adding an additional commit.)

After this change the PR is good to go.

adamreeve:
OK thanks, I've removed the `GModule-2.0.gir` project now.

PR #1091, opened 2024-07-05
Support loading repositories from files embedded as resources in GirTool

### LINKED ISSUE #1089: Include gir files in GirTool as embedded resources
Related to #1078 

In order to make GirTool useful as a standalone tool, it would be helpful to have the gir files for the common GObject based libraries already supported by gir.core available as embedded resources, so that users don't need to check out the gir-files repository to use GirTool.

Users would then need to ensure that they reference the required GirCore NuGet packages corresponding to the dependencies needed by their library, and the package versions should match the version of GirTool used.

I don't think this needs a new command line argument to GirTool, but the logic for looking up a gir file would be to first check in the directories from the `--search-path-linux/macos/windows` arguments, and if a file isn't found there, look it up from the embedded files.

adamreeve:

Resolves #1089

One thing I'm not sure about is whether there should be a more explicit opt in to loading files from the embedded resources. I think the current approach is OK but let me know if you disagree.


src/Generation/GirTool/GenerateCommand.cs
```diff
@@ -137,15 +137,14 @@ private static (DeserializedInput, DeserializedInput, DeserializedInput) LoadRep
...
-            return DeserializedInput.Empty();
+        var repositoryResolver = GetRepositoryResolver(platformName, searchPath);
 
         var inputRepositories = input
-            .Select(x => Path.Join(searchPath, x))
-            .Where(File.Exists)
-            .Select(x => new FileInfo(x).OpenRead().DeserializeGirInputModel())
+            .Select(fileName => repositoryResolver.ResolveRepository(fileName))
+            .Where(inputRepository => inputRepository != null)
+            .Select(inputRepository => inputRepository!)
```
badcel, line 147:
```suggestion
```

This is obsolete.

src/Generation/GirTool/GenerateCommand.cs
```diff
@@ -137,15 +137,14 @@ private static (DeserializedInput, DeserializedInput, DeserializedInput) LoadRep
...
-        if (searchPath is null)
-            return DeserializedInput.Empty();
+        var repositoryResolver = GetRepositoryResolver(platformName, searchPath);
 
         var inputRepositories = input
-            .Select(x => Path.Join(searchPath, x))
-            .Where(File.Exists)
-            .Select(x => new FileInfo(x).OpenRead().DeserializeGirInputModel())
+            .Select(fileName => repositoryResolver.ResolveRepository(fileName))
+            .Where(inputRepository => inputRepository != null)
```
badcel, line 146:
```suggestion
            ..OfType<GirLoader.Input.Repository>()
```

This automatically discards all `null` values.

badcel (changes requested):
I think my comments are all little nit picks so all in all this is pretty good work, thank you :+1:.

In regard to your question: I think it is not a good idea to automatically fall back to the embedded resolver if there is no search path supplied: A current feature of `GirTool` is that it allows to create bindings for only a subset of the available platforms.

For example if no search path for windows is supplied the code is generated like always but all the methods are annotated in a way that makes clear that it is not supported under windows. I think it makes sense to keep this feature alive.

The resulting question is how to solve this. One idea would be to have some `RepositoryResolverFactory` class which is responsible for delivering some `IRepositoryResolver` via a `Create` method. So for each platform there can be a repository resolver, but it shouldn't be a problem if there is none, as the code would be annotated, accordingly. (I would like to see a separate class to create `IRepositoryResolver` as the `GenerateCommand.cs` file is already pretty crammed.)

Perhaps you can come up with some nice idea how to solve this elegantly?

src/Generation/GirLoader/Helper/ChainedRepositoryResolver.cs
```diff
@@ -0,0 +1,31 @@
...
+/// <summary>
+/// Tries multiple repository resolvers, returning the first result found
+/// </summary>
+public class ChainedRepositoryResolver : IRepositoryResolver
+{
+    private readonly IRepositoryResolver[] _resolvers;
+
+    public ChainedRepositoryResolver(IEnumerable<IRepositoryResolver> resolvers)
+    {
+        _resolvers = resolvers.ToArray();
```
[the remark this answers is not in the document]
adamreeve: In general I don't like storing an `IEnumerable` member due to enumerating possibly being expensive, and the rarer problems like results being able to change between enumerations. But accepting an `IEnumerable` parameter provides maximum flexibility to calling code.

But like you say this is all internal code rather than being part of an API for external use, so that flexibility is unnecessary. This could probably just be an `IRepositoryResolver[]` parameter to be honest, there isn't really a need to accept a more general interface. But the extra overhead of using something like `IReadOnlyCollection` is going to be negligible and that better fits the intended behaviour of this member without tying it to a specific type, so I will switch to that.

And yes you're right that this will cause a copy of the array. Copying an array of two elements once on application startup didn't seem like a big deal, but yes it's better to avoid the copy if it's not needed.

src/Tests/Generation/GirLoader.Tests/DisposableTempDirectory.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    private readonly string _directoryPath;
+
+    public DisposableTempDirectory()
+    {
+        _directoryPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+        Directory.CreateDirectory(_directoryPath);
+    }
+
+    public string DirectoryPath => _directoryPath;
```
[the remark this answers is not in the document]
adamreeve: I had called this `DirectoryPath` rather than `Path` to avoid clashing with the `System.IO.Path` namespace, but I've renamed this and updated the other uses of `Path` to qualify them with their namespace.

src/Generation/GirTool/GenerateCommand.cs
adamreeve: This was needed to avoid nullability errors later as the compiler thinks the type of `inputRepositories` is `List<IRepository?>` rather than `List<IRepository>` without this line. But your suggestion to use `OfType` means this won't be needed.

src/Generation/GirTool/GenerateCommand.cs
adamreeve: I'm not sure I 100% agree about this being easier to understand, it's less obvious to me that the intent of this is to filter out null values, but I do think this is nicer on balance due to not needing to extra select to tell the compiler that the values are non-null.

adamreeve:
Thanks for the review @badcel, I think I've addressed all of your comments now.

I've left the changes as separate commits to help with reviewing, but let me know if you want me to rebase everything down to one commit (and rebase on the new main head).

badcel:
Hey, thanks for the update. :rocket: It looks pretty 
good I found no issues while reviewing the changes.

Currently I don't have the time to do a test on my hardware. So there merge has to wait until then (probably in 2-3 days).

Can you squash your commits and rebase on the current main branch so everything is ready to be merged as soon as I have the possibility to test on my own hardware?

Thank you and sorry for the delay.

adamreeve:
No problem, I've squashed that now thanks.

badcel (approved):
Thank you, this is good to go!

PR #1097, opened 2024-07-18
Add new modifier to class methods that hide base class methods

### LINKED ISSUE #1094: ToString methods hide object.ToString
In the Arrow GObject library there are some `to_string()` methods on derivable types that result in warnings like:
```
X.ToString()' hides inherited member 'object.ToString()'. To make the current member override that implementation, add the override keyword. Otherwise add the new keyword.
```

This is only a warning rather than an error, but we build with `TreatWarningsAsErrors` enabled, and it would be nice to fix this.

I can see there's a `RecordEqualsMethodCollidesWithGeneratedCode` class that fixes a similar problem with `Equals` methods in Records by renaming them to `NativeEquals`, so a similar fix would be to rename these to `NativeToString`. It might make sense to add an `override` modifier instead though as it seems like it would generally make sense to use the native `to_string` method to implement `object.ToString`.

I'm happy to implement the fix for this after we agree on how to do it.

adamreeve:

Resolves #1094


src/Generation/Generator/Model/Method.cs
```diff
@@ -84,4 +85,80 @@ public static bool IsValidFreeFunction(GirModel.Method method)
...
+    }
+
+    /// <summary>
+    /// Whether this method hides a method from System.Object
+    /// </summary>
+    private static bool HidesObjectMethod(GirModel.Method method, string publicName)
+    {
+        if (method.Parameters.Any())
+        {
+            // We do not currently support detecting overrides of object methods that accept parameters
```
adamreeve, line 143: It looks like it would be a bit more complicated to check the parameter types against the ones from reflection of the `object` type, and given I've only seen this be an issue for `ToString`, which doesn't take any parameters, it doesn't seem worth the effort to implement this yet.

badcel: This is fine for now 👍

badcel:
I'm not sure if the unit tests in this case are of the type I want to have: They are tightly coupled to the implementation.

For generated code I would prefer to have tests that are independent from the generator implementation.

I don't know if this works out but my preferred solution would be to have some new `GObject` class which has a `to_string` method.

The warning which is emitted for the missing `new` / `override` operator must be converted to an error so the build would fail if it is missing. As the generator is applying the operator the build would not fail. The test would then call the method to verify the expected string is returned.

Additionally the instance can be casted to `System.Object` to call `System.Object.ToString` method and verify the original implementation is still accessible.

What do you think?

adamreeve:
Thanks for the feedback, yes that seems like a better approach. I wasn't that happy with how the tests were implemented either but didn't think of changing the test project to treat the warning as an error.

src/Generation/Generator/Renderer/Public/InterfaceImplementation/InterfaceImplementationFunctions.cs
```diff
@@ -22,7 +22,8 @@ namespace {Namespace.GetPublicName(@interface.Namespace)};
 
 public partial class {Interface.GetImplementationName(@interface)} : GObject.GTypeProvider
 {{
-    {FunctionRenderer.Render(@interface.TypeFunction)}
+    // The implementation class also inherits from GObject.Object so GetType needs the new modifier
```
adamreeve, line 25: After adding the `WarningsAsErrors` property to the test project I found that the `GetType` methods for `InterfaceHelper` classes were also causing an error. This fix feels a bit clunky but I couldn't see a better way to handle this automatically.

badcel: Hey sorry for the delay I'm pretty occupied currently.

I can't give a detailed explanation because I have none, it's still early in the morning here. But I don't know when I will find time otherwise so please bear with me.

From taking a look i hope it should be fixable without a new parameter. In [HidesMethod](https://github.com/gircore/gir.core/blob/8b6a4587e5ddb9b3a5ec49e83698336dde72400c/src/Generation/Generator/Model/Function.cs#L36) you see that the `Parent` is checked for `GirModel.Class`. This implies that it could be something different. So see the type of the Parent property which is [ComplexType](https://github.com/gircore/gir.core/blob/8b6a4587e5ddb9b3a5ec49e83698336dde72400c/src/Generation/GirModel/ComplexType.cs#L3).

An interface is a ComplexType, too. So you could do some special casing for interfaces (like for classes) in the method.

I think I like this a lot better because **all** the logic which decides remains inside the model and is not dependent on who is calling the method this providing a "magic" parameter to fix things.

badcel (commented):
I checked the code and except the new parameter to force the `new` modifier it looks fine to me. Let's see what your thoughts on my comment are.

src/Generation/Generator/Renderer/Public/InterfaceImplementation/InterfaceImplementationFunctions.cs
adamreeve: No problem. I tried this approach but I got stuck as I think I'd need to get a handle to the `GirModel.Class` for `GObject.Object` in order to reuse the `HidesMethod` logic, and I don't think that would be possible without plumbing it in through a lot of different layers from the Loader, but maybe I'm missing something? Eg: 
```c#
public static bool HidesFunction(GirModel.Function function)
{
    switch (function.Parent)
    {
        case GirModel.Class cls:
            return HidesFunction(cls.Parent, function);
        case GirModel.Interface:
        {
            // Rendering a function for an interface helper class,
            // which inherits from GObject.Object
            GirModel.Class gobjectObjectClass = ??? // How to get this?
            return HidesFunction(gobjectObjectClass, function);
        }
        default:
            return false;
    }
}
```

badcel: Yeah this is not possible.

Just as a sidenote: Currently interfaces never implement other interfaces (#1101) as the model is not supporting it. I don't know if this is relevant for you? Interfaces could have conflicting methods / functions, too.

Back to topic: Interfaces are like an independent chain of "things" not related to classes in itself. But classes are the things implementing the interfaces. So there can be conflicts in classes implementing interfaces like you mentioned.

As interfaces have their own type the function can clash with the type function of classes. For methods this solved via a [fixer](https://github.com/gircore/gir.core/blob/8b6a4587e5ddb9b3a5ec49e83698336dde72400c/src/Generation/Generator/Fixer/Class/InterfaceMethodsCollidingWithClassMethodsFixer.cs#L5) to remember to implement this methods explicitly.

I think starting with dotnet 7 we can have static methods in interfaces. If it is possible to implement them explicitly every `GetType` function of interfaces could be marked to be implemented explicitly. (We would drop dotnet 6/7 support for this what would happen sooner or later in either case.)

The more quick and dirty workaround is similar to your initial approach: you hardcoded the "force parameter" to true for type functions in interfaces. So you could just hardcode in the function renderer: "If this function is inside an interface and it is the GetType function render the "new" modifier.

I probably would not like to see this if statement inside the actual rendering part of the method but more like in your last comment. So there is some clever implementation needed. (I dont have an idea currently. For me it is okay if the solution is a workaround / hardcoded but the  structure should be clean.)

Alternatively you just retain the new parameter but not as part of the "public" API but just set it internally if the if statement which checks for interfaces with a GetType function detects those and just sets the parameter.

If you are interested in verifying and implementing the explicit implementation way I would prefer this. If not it is okay too. In this case I would open an issue for the explicit interface function implementation as I think it is ultimately the rigjt thing to do in case it works.

adamreeve: Sorry for the slow response, I've changed this now to use a workaround in the `HidesFunction` function that checks whether the function is the type function, which I think is what you had in mind for the quick and dirty workaround.

I haven't come across an interface implementing another interface in the Arrow library, there's a `File` interface and a `WriteableFile` interface but `WritableFile` doesn't implement `File` for example, so I don't think this is a problem for me.

One scenario that is still causing errors is where a subclass and parent class both define methods that implement an interface, eg:
```c#
public interface File {
  bool Close();
}

public class OutputStream : File {
  bool Close() {}
}

public class BufferOutputStream : OutputStream {
  bool Close() {}  // "hides inherited member" error
}
```

I think I can extend `Method.HidesMethod` to detect this situation too, or possibly just disable converting this warning to an error for my project to work around that issue for now :smile:

badcel: Hey,

no problem your work is very much appreciated.

If you are fine with the current code state, I'm too.

I had no specific solution in mind. The one you choose is even simpler than I thought possible. So great work 👍.

I will run a manual test on my system before merging but I don't expect anything to come up.

adamreeve: Cool OK yeah I'm happy with this as is, I might follow up with a separate PR for the related issue with subclasses. Thanks

badcel (approved):

PR #1115, opened 2024-08-18
ImageSurface: Support memory pressure

badcel:

### DIFF
--- a/src/Libs/cairo-1.0/Internal/SurfaceHandle.cs
+++ b/src/Libs/cairo-1.0/Internal/SurfaceHandle.cs
@@ -1,9 +1,30 @@
-﻿namespace Cairo.Internal;
+﻿using System;
+
+namespace Cairo.Internal;
 
 public partial class SurfaceOwnedHandle
 {
+    private long _memoryPressure;
+
+    internal void AddMemoryPressure(long memoryPressure)
+    {
+        if (memoryPressure > 0)
+        {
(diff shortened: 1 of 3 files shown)


badcel:
@cameronwhite Can you take a look here? I'm pretty sure that the code to add memory pressure in `Surface` is incorrect / to low. Can we add some kind of logic to calculate the size in bytes of all supported surfaces?

cameronwhite:
Yeah I think the `width * height` is incorrect since that's the width in pixels, but that doesn't take into account the pixel format (number of bytes per pixel)
I don't think there is actually an API to get a memory usage estimate other than for ImageSurface (which does directly correspond to a memory buffer with a specific pixel format). E.g. some backends like SVG probably don't have a reasonable way to produce this

A general function could check the surface's type (https://www.cairographics.org/manual/cairo-cairo-surface-t.html#cairo-surface-get-type) and then only add memory pressure for ImageSurface? (That's also the only surface type we create in Pinta)
Using width * height could also be a fallback estimate, but I don't know if over-estimating the memory pressure is a bad thing for the GC

badcel:
@cameronwhite I updated the code to only apply memory pressure if we know the size of the image surface and skip the memory pressure in other cases. I think it is better to start as conservative as possible and we can add additional use cases if the need comes up.

Do you think the code is good to go?

cameronwhite (approved):
Looks good to me, thanks!

PR #1116, opened 2024-08-29
Support class based safe handles

badcel:

**This is only a concept PR. See #1118 for the work in progress PR**


badcel:
Replaced by #1118.

Issue #1125, opened 2024-09-23
Strange drag and drop crash bug


ravener:
I'm dealing with a weird crash happening with the use of `Gtk.DropTarget`

Strangely enough this only seems to happen inside Flatpak runtime. On both GNOME SDK 46 and 47. But I cannot reproduce it on my own machine locally (Debian 12, GNOME 43).

The problem seems related to garbage collection.

I've narrowed it down to a simple reproducible window that shows this:

```cs

namespace MyApp
{
    public class BugWindow : Adw.ApplicationWindow
    {
        private readonly Gtk.Button _button;
        private readonly Gtk.Button _button2;
        private readonly Gtk.DropTarget _drop;
        private readonly Gtk.DropTarget _drop2;
        public BugWindow(Gtk.Application app)
        {
            Application = app;

            var box = Gtk.Box.New(Gtk.Orientation.Vertical, 6);
            box.Append(Adw.HeaderBar.New());
            _button = Gtk.Button.NewWithLabel("File 1");
            _button2 = Gtk.Button.NewWithLabel("File 2");
            _button.OnClicked += OnFileSelect;
            _button2.OnClicked += OnFileSelect;
            box.Append(_button);
            box.Append(_button2);

            SetContent(box);

            _drop = Gtk.DropTarget.New(Gio.FileHelper.GetGType(), Gdk.DragAction.Copy);
            _drop.OnDrop += OnDrop;
            _button.AddController(_drop);

            _drop2 = Gtk.DropTarget.New(Gio.FileHelper.GetGType(), Gdk.DragAction.Copy);
            _drop2.OnDrop += OnDrop;
            _button2.AddController(_drop2);
        }

        private bool OnDrop(Gtk.DropTarget drop, Gtk.DropTarget.DropSignalArgs e)
        {
            System.GC.Collect(); // Force gc for stress testing.
            Console.WriteLine("dropped!");
            var file = new Gio.FileHelper(e.Value.GetObject()!.Handle, false);
            var path = file.GetPath();
            Console.WriteLine(path);

            if (File.Exists(path))
            {
                Console.WriteLine("File exists.");
                if (drop == _drop)
                {
                    _button.SetLabel(file.GetBasename()!);
                    Console.WriteLine("Set label");
                }
                else if (drop == _drop2)
                {
                    _button2.SetLabel(file.GetBasename()!);
                    Console.WriteLine("Set label 2");
                }

                return true;
            }

            return false;
        }

[...]

PR #1128, opened 2024-09-29
Object: Fix ToggleNotify called after callback is disposed

badcel:
Fixes: #1125 

### DIFF
--- a/src/Libs/GObject-2.0/Internal/ObjectMapper.ToggleRef.cs
+++ b/src/Libs/GObject-2.0/Internal/ObjectMapper.ToggleRef.cs
@@ -5,10 +5,9 @@ namespace GObject.Internal;
 
 public partial class ObjectMapper
 {
-    private class ToggleRef : IDisposable
+    private class ToggleRef
     {
         private object _reference;
-        private readonly ToggleNotify _callback;
         private readonly IntPtr _handle;
 
         public object? Object
@@ -40,17 +39,9 @@ public partial class ObjectMapper
         public ToggleRef(IntPtr handle, object obj, bool ownedRef)
         {
             _reference = obj;
-            _callback = ToggleReference;
             _handle = handle;
 
             OwnReference(ownedRef);
-            RegisterToggleRef();
-        }
-
-        private void RegisterToggleRef()
-        {
-            Internal.Object.AddToggleRef(_handle, _callback, IntPtr.Zero);
-            Internal.Object.Unref(_handle);
         }
 
         private void OwnReference(bool ownedRef)
@@ -72,7 +63,7 @@ public partial class ObjectMapper
             }
         }
 
-        private void ToggleReference(IntPtr data, IntPtr @object, bool isLastRef)
+        internal void ToggleReference(bool isLastRef)
         {
             if (!isLastRef && _reference is WeakReference weakRef)
             {
@@ -85,16 +76,6 @@ public partial class ObjectMapper
             {
                 _reference = new WeakReference(_reference);
             }
-        }
-
-        public void Dispose()
-        {
-            var sourceFunc = new GLib.Internal.SourceFuncAsyncHandler(() =>
-            {
-                Internal.Object.RemoveToggleRef(_handle, _callback, IntPtr.Zero);
-                return false;
-            });
-            GLib.Internal.MainContext.Invoke(GLib.Internal.MainContextUnownedHandle.NullHandle, sourceFunc.NativeCallback, IntPtr.Zero);
         }
     }
 }
--- a/src/Libs/GObject-2.0/Internal/ObjectMapper.ToggleRegistration.cs
+++ b/src/Libs/GObject-2.0/Internal/ObjectMapper.ToggleRegistration.cs
@@ -1,0 +1,32 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace GObject.Internal;
+
+public partial class ObjectMapper
+{
+    private static unsafe class ToggleRegistration
+    {
+        internal static void AddToggleRef(IntPtr handle)
+        {
+            AddToggleRef(handle, &ToggleNotify, IntPtr.Zero);
+            Internal.Object.Unref(handle);
+        }
+
+        internal static void RemoveToggleRef(IntPtr handle)
+        {
+            var sourceFunc = new GLib.Internal.SourceFuncAsyncHandler(() =>
+            {
+                RemoveToggleRef(handle, &ToggleNotify, IntPtr.Zero);
+                return false;
+            });
+            GLib.Internal.MainContext.Invoke(GLib.Internal.MainContextUnownedHandle.NullHandle, sourceFunc.NativeCallback, IntPtr.Zero);
+        }
+
+        [DllImport(ImportResolver.Library, EntryPoint = "g_object_add_toggle_ref")]
+        private static extern void AddToggleRef(IntPtr @object, delegate* unmanaged<IntPtr, IntPtr, int, void> toggleNotify, IntPtr data);
+
+        [DllImport(ImportResolver.Library, EntryPoint = "g_object_remove_toggle_ref")]
+        private static extern void RemoveToggleRef(IntPtr @object, delegate* unmanaged<IntPtr, IntPtr, int, void> toggleNotify, IntPtr data);
+    }
+}
--- a/src/Libs/GObject-2.0/Internal/ObjectMapper.cs
+++ b/src/Libs/GObject-2.0/Internal/ObjectMapper.cs
@@ -2,24 +2,38 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
 using GLib;
 
 namespace GObject.Internal;
 
 public static partial class ObjectMapper
 {
+    private static readonly object Lock = new();
     private static readonly Dictionary<IntPtr, ToggleRef> WrapperObjects = new();
 
-    public static int ObjectCount => WrapperObjects.Count;
+    public static int ObjectCount
+    {
+        get
+        {
+            lock (Lock)
+            {
+                return WrapperObjects.Count;
+            }
+        }
+    }
 
     public static bool TryGetObject<T>(IntPtr handle, [NotNullWhen(true)] out T? obj) where T : class, IHandle
     {
-        if (WrapperObjects.TryGetValue(handle, out ToggleRef? weakRef))
+        lock (Lock)
         {
-            if (weakRef.Object is not null)
+            if (WrapperObjects.TryGetValue(handle, out ToggleRef? weakRef))
             {
-                obj = (T) weakRef.Object;
-                return true;
+                if (weakRef.Object is not null)
+                {
+                    obj = (T) weakRef.Object;
+                    return true;
+                }
             }
         }
 
@@ -29,9 +43,10 @@ public static partial class ObjectMapper
 
     public static void Map(IntPtr handle, object obj, bool ownedRef)
     {
-        lock (WrapperObjects)
+        lock (Lock)
         {
             WrapperObjects[handle] = new ToggleRef(handle, obj, ownedRef);
+            ToggleRegistration.AddToggleRef(handle);
         }
 
         Debug.WriteLine($"Handle {handle}: Mapped object of type '{obj.GetType()}' as owned ref '{ownedRef}'.");
@@ -39,12 +54,24 @@ public static partial class ObjectMapper
 
     public static void Unmap(IntPtr handle)
     {
-        lock (WrapperObjects)
+        lock (Lock)
         {
-            if (WrapperObjects.Remove(handle, out var toggleRef))
-                toggleRef.Dispose();
+            if (WrapperObjects.Remove(handle))
+                ToggleRegistration.RemoveToggleRef(handle);
         }
 
         Debug.WriteLine($"Handle {handle}: Unmapped object.");
+    }
+
+    [UnmanagedCallersOnly]
+    private static void ToggleNotify(IntPtr data, IntPtr @object, int isLastRef)
+    {
+        lock (Lock)
+        {
+            if (WrapperObjects.TryGetValue(@object, out var toggleRef))
+                toggleRef.ToggleReference(isLastRef != 0);
+            else
+                Debug.WriteLine($"Handle {@object}: Could not toggle to {isLastRef} as there is no toggle reference.");
+        }
     }
 }


badcel:
@ravener can you confirm that this PR fixes the bug #1125 for you?

@cameronwhite This fix updates code introduced due to a reported bug by a pinta user. Any thoughts on this? 

I think it should now work more reliable than before. I'm not sure if the dictionary access is "safe enough" though or if there is perhaps some concurrent dictionary needed. As those versions would probably be slower I think I will keep it like it is for now if there are no objections.

cameronwhite:
I'll need to build Pinta against this change and test it out, but from reading through the code I had a couple questions:

- It looks like this removes the behaviour of scheduling the `RemoveToggleRef()` call to occur on the main thread, which I think we needed for UI-related GObjects that can't be destroyed from the GC thread?

- I don't think the access to `WrapperObjects` seems safe, same as in `TryGetObject` too. In other methods like `Map()` and `Unmap()` we lock around access to the dictionary, so that implies that read-only accesses should be locked too so that they can't happen while a modification of the dictionary is in progress

badcel:
@cameronwhite regarding your threadding remark i think it should ne Fine. The [docs](https://docs.gtk.org/gobject/method.Object.remove_toggle_ref.html) say that a call to `ToggleNotify` can happen after removing the toggle reference. The managed toggle notify callback

badcel:
@cameronwhite thanks for your answer. I misremembered the original pinta bug. I think both of your points are valid. I will update this PR accordingly.

ravener:
I'd love to test it, but I'm not sure how, is there an easy way to pull a PR and have it as a dependency? Not so used to working with dotnet sorry.

badcel:
@ravener you would basically do:

```sh
$ git clone --recursive https://github.com/gircore/gir.core.git
$ cd gir.core/src
$ git checkout fix-1125
$ dotnet fsi GenerateLibs.fsx
$ dotnet build GirCore.Libs.slnf
```

This is mostly identical to the [build instructions in the readme](https://github.com/gircore/gir.core?tab=readme-ov-file#build) except that you checkout the branch of this PR instead of the `main` branch.

Afterwards you have to use the build binaries in your project. The binaries are for example located in the `Libs/Adw-1/bin` folder.

badcel:
@ravener 

I updated the `Samples/Adw-1/Window/Program.cs` to the following code to test your issue. The code is not nicely formatted but still reproduced your issue for me and is a little more stripped down than yours.

It revealed that the problem was the file helper instance. You can simply copy and paste the code into the same file and should be able to test your issue without the need to do more than just coping over the code.

```CSharp
using System;
using System.IO;
using Gtk;

var application = Adw.Application.New("org.gir.core", Gio.ApplicationFlags.FlagsNone);
application.OnActivate += (sender, args) =>
{
    var w = new BugWindow((Application)sender);
    w.Show();
};
return application.RunWithSynchronizationContext(null);

public class BugWindow : Adw.ApplicationWindow
    {
        private readonly Gtk.Button _button;
        private readonly Gtk.DropTarget _drop;
        private int count;
        public BugWindow(Gtk.Application app)
        {
            Application = app;

            var box = Gtk.Box.New(Gtk.Orientation.Vertical, 6);
            box.Append(Adw.HeaderBar.New());
            _button = Gtk.Button.NewWithLabel("File 1");
            _button.OnClicked += OnFileSelect;
            box.Append(_button);

            SetContent(box);

            _drop = Gtk.DropTarget.New(Gio.FileHelper.GetGType(), Gdk.DragAction.Copy);
            _drop.OnDrop += OnDrop;
            _button.AddController(_drop);
        }

        private bool OnDrop(Gtk.DropTarget drop, Gtk.DropTarget.DropSignalArgs e)
        {
            System.GC.Collect(); // Force gc for stress testing.
            Console.WriteLine("dropped!");
            Console.WriteLine("FilePointer: " + e.Value.GetObject()!.Handle);
            var file = new Gio.FileHelper(e.Value.GetObject()!.Handle, false);
            //var path = file.GetPath();
            //Console.WriteLine(path);

            //if (!File.Exists(path))
            //    return false;

            Console.WriteLine("File exists.");
            if (drop != _drop)
                return false;

            _button.SetLabel("Bla" + count++);
            //_button.SetLabel(file.GetBasename()!);
            Console.WriteLine("Set label");

            return true;

        }

        private void OnFileSelect(Gtk.Button sender, EventArgs args)
        {
            Console.WriteLine("here we are.");
            var chooser = Gtk.FileChooserNative.New("Open stuff", this, Gtk.FileChooserAction.Open, null, null);
            chooser.Show();
        }
    }
```

Please also confirm that the issure is present if you run the test code on the main branch. Checkout the code via

```
git checkout main
```

After Checkout you do not need to generate the project again as no generated files are altered. Just start the Window project again and see that the error is there to double check that my sample code creates the same issue like your sample code on your machine.

ravener:
Unfortunately I'm having issues building locally

`dotnet fsi GenerateLibs.fsx` gives me a large output of over 6000 lines of warnings and failing with some error, had to upload a file since it was too big

[err.txt](https://github.com/user-attachments/files/17249748/err.txt)

If I just continue to change the adwaita sample and run `dotnet run` anyways (keyword 'anyways' so this error could just be a result of the fsi command not finishing)

```
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Public/Uri.Generated.cs(483,15): warning CS0114: 'Uri.ToString()' hides inherited member 'object.ToString()'. To make the current member override that implementation, add the override keyword. Otherwise add the new keyword. [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/AsyncQueueHandle.cs(11,42): error CS0759: No defining declaration found for implementing declaration of partial method 'AsyncQueueHandle.OwnedCopy()' [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/AsyncQueueHandle.cs(16,44): error CS0759: No defining declaration found for implementing declaration of partial method 'AsyncQueueHandle.UnownedCopy()' [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/AsyncQueueHandle.Generated.cs(22,34): error CS0111: Type 'AsyncQueueHandle' already defines a member called 'OwnedCopy' with the same parameter types [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/AsyncQueueHandle.Generated.cs(27,36): error CS0111: Type 'AsyncQueueHandle' already defines a member called 'UnownedCopy' with the same parameter types [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/AsyncQueueHandle.cs(27,49): error CS0759: No defining declaration found for implementing declaration of partial method 'AsyncQueueOwnedHandle.FromUnowned(IntPtr)' [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/AsyncQueueHandle.cs(32,37): error CS0759: No defining declaration found for implementing declaration of partial method 'AsyncQueueOwnedHandle.ReleaseHandle()' [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/AsyncQueueHandle.Generated.cs(71,45): error CS0111: Type 'AsyncQueueOwnedHandle' already defines a member called 'FromUnowned' with the same parameter types [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/AsyncQueueHandle.Generated.cs(78,29): error CS0111: Type 'AsyncQueueOwnedHandle' already defines a member called 'ReleaseHandle' with the same parameter types [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Public/Variant.Generated.cs(631,25): warning CS0108: 'Variant.GetType()' hides inherited member 'object.GetType()'. Use the new keyword if hiding was intended. [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Public/Variant.Generated.cs(895,22): error CS0111: Type 'Variant' already defines a member called 'Dispose' with the same parameter types [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/VariantHandle.cs(11,39): error CS0759: No defining declaration found for implementing declaration of partial method 'VariantHandle.OwnedCopy()' [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/VariantHandle.cs(16,41): error CS0759: No defining declaration found for implementing declaration of partial method 'VariantHandle.UnownedCopy()' [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/VariantHandle.Generated.cs(22,31): error CS0111: Type 'VariantHandle' already defines a member called 'OwnedCopy' with the same parameter types [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/VariantHandle.Generated.cs(27,33): error CS0111: Type 'VariantHandle' already defines a member called 'UnownedCopy' with the same parameter types [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/VariantHandle.cs(27,46): error CS0759: No defining declaration found for implementing declaration of partial method 'VariantOwnedHandle.FromUnowned(IntPtr)' [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/VariantHandle.cs(32,37): error CS0759: No defining declaration found for implementing declaration of partial method 'VariantOwnedHandle.ReleaseHandle()' [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/VariantHandle.Generated.cs(71,42): error CS0111: Type 'VariantOwnedHandle' already defines a member called 'FromUnowned' with the same parameter types [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/Internal/VariantHandle.Generated.cs(78,29): error CS0111: Type 'VariantOwnedHandle' already defines a member called 'ReleaseHandle' with the same parameter types [/home/ravener/repos/cs/gir.core/src/Libs/GLib-2.0/GLib-2.0.csproj::TargetFramework=net6.0]
```

I'm on Debian 12, maybe outdated system libraries are a potential reason?

Recall that the original bug was not reproducible on my debian 12 machine anyway, it only happened inside flatpak runtime, so newer gnome runtimes probably introduced it. I'm still not comfortable with flatpak build tools or I would've tried to build this inside flatpak in case the build step requires newer system libraries (can you confirm this? or do you expect it to build on older distros regardless?)

ravener:
I can at least do the second part, but without building from source, continuing with the latest nuget version like before, I can confirm your sample is still the same issue as my original code. And again, works fine locally on Debian 12 but crashes on flatpak

badcel:
@ravener if you have dotnet 8 installed I would expect it to generate code on any system. To build you would need dotnet 6,7 and 8.

The tool simply reads the Gir xml files which are part of the repository and outputs some textfiles which happen to be code: nothing system dependent here.

I would even expect the sample to work on your Debian machine as the used adwaita API in the sample program should be "old enough".

Yesterday I explicitly tested the commands I gave you in a fresh folder and it worked for me. I need to do some more investigation.

On the other hand I expect the problem to be solved with this as you confirmed the problem occurs with my demo code on your side, too.

badcel:
If you want to keep Debian but want to create GTK apps which you distribute as a flatpak I would recommend using tools like [distrobox](https://distrobox.it/) and [toolbx](https://containertoolbx.org/).

With those you can get development environments which are a lot closer to the flatpak GNOME runtimes than Debian if you choose fedora for example.

The GirCore nuget packages are build against the flatpak runtime.

If you want to create apps for Debian you would probably be better served if you generate the binding binaries locally and use the Debian Gir files. Then the generated C# API matches the C api available in Debian and not in the flatpak runtime.

ravener:
Well I did try to build it locally, doesn't that target debian then? For my usual development, I have a nice VSCode + Flatpak workflow setup that allows me to build the app against the Flatpak runtime which works well so far.

Anyway I'm not going to waste your time, I'll just wait for prebuilt binaries before I retest this since building proved to be harder than I thought, sorry for being unhelpful.

badcel:
You are not unhelpful.

If you build the project according to the build instructions the binaries target the flatpak runtime because the [gir files](https://github.com/gircore/gir-files/tree/c2dab7dee23d5b5aa29bfb43c9191412cef65f59) are part of the repository. The system Gir files are not used by default.

In the beginning of the project  I tried different sources of Gir files. Different problems occurred depending on which Gir files were used. As flatpaks are distro agnostic I settled with the flatpak runtime Gir files.

Perhaps there could be some command line option to make it easier to use system Gir files.

ravener:
> I would be very interested in your workflow to build against the flatpak runtime. Do you have some repository which shows how you do it?

This is my first app so far, but since I'm close to finishing it and was gonna open source it soon anyway, I decided to make the repo now at https://github.com/ravener/Oto

I was kind of inspired by [Denaro](https://github.com/NickvisionApps/Denaro) but they used Cake scripts which I didn't feel like adding to my project so I opted to write a simple shell script that just copies the relevant files.

Then I rely on [flatpak-vscode](https://marketplace.visualstudio.com/items?itemName=bilelmoussaoui.flatpak-vscode) extension, which automatically picks up the flatpak manifest and adds a 'Build & Run' button (shortcut CTRL + Alt + B) but you may as well manually use the flatpak builder tool if you are familiar with it, I never touched that myself yet, the extension is just a neat layer on top of it making vscode as easy as using something like gnome builder.

PR #1139, opened 2024-11-26
Initial hello world tutorials

anthonyirwin82:

I have created a basic hello world tutorial to help people getting started.


badcel:
Thanks I will take a look in the coming days.

docs/docs/tutorial/gtk/index.md
```diff
@@ -0,0 +1,10 @@
+# Gtk Tutorial
+
+When learning GUI frameworks like GTK, QT and wxWidgets it is common to teach the basics using code to make the learning experience easier. In a real life GTK application you will probably want to create a ui file using either [GTK Builder XML](https://docs.gtk.org/gtk4/class.Builder.html) or [Blueprint](https://jwestman.pages.gitlab.gnome.org/blueprint-compiler/) to build the UI.
+
+On the internet you will often see Glade mentioned for building GUIs in a user friendly wysiwyg editor. Glade does not work with Gtk 4 however there is a new project [Camblache](https://gitlab.gnome.org/jpu/cambalache) which will build GUIs in a similar way to the old Glade project that was used in GTK 2/3.
```
badcel, line 5:
```suggestion
On the internet you will often see Glade mentioned for building GUIs in a user friendly wysiwyg editor. Glade does not work with Gtk 4 however there is a new project [Cambalache](https://gitlab.gnome.org/jpu/cambalache) which will build GUIs in a similar way to the old Glade project that was used in GTK 2/3.
```

docs/docs/tutorial/gtk/index.md
```diff
@@ -0,0 +1,10 @@
...
+
+If you are wanting to use Blueprint to build your UI then you can use [Workbench](https://apps.gnome.org/en-GB/Workbench/) to see a preview as you write the blueprint.
```
badcel, line 7:
```suggestion
If you want to use Blueprint to build your UI then you can use [Workbench](https://apps.gnome.org/en-GB/Workbench/) to see a preview as you write the blueprint.
```

badcel (commented):
Hi,

thanks for your contribution.

I added some small grammatical improvements. In general I think there should not be a GTK and a libadwaita tutorial mostly doing the same.

I would suggest starting with a GTK tutorial and then transforming the GTK App into a libadwaita app in a next tutorial.

In this way the tutorial evolves the sample more and more.

Can you remove the libadwaita part of the tutorial? If the GTK part is merged we can see how to go on.

I wonder if it is a good idea to include the tutorial code in the solution so it gets build by the CI so we can be sure the code is valid?

anthonyirwin82:
It makes sense to concentrate on Gtk first as that is what Libadwaita is built on top of and some people only want to make Gtk apps and people who want Libadwaita it's easy to transfer over once you are up and running with Gtk.

I have removed the Libadwaita tutorial section and updated the Gtk tutorial to have full project files so it can be setup with CI to test the tutorial code works.

I have started on a second follow up tutorial using Gtk.Box to add a label and having the button click change the label text as a second tutorial demonstrating using the Box layout widget with multiple widgets on the screen.

I then plan to look at using things like the following:

- Gtk.HeaderBar with a Gtk.MenuButton with AboutDialog option to click.
- Traditional Gtk.Menubar
- Gtk.ToolBar
- using other widgets

I tried to use Gtk.AlertDialog but could not get it to work with GirCore. I have used the gtkmm book tutorial for alert dialog as well as either vala or pygobject I can't remember which. I cannot seem to get a GtkAlertDialog to work in GirCore, I got Gtk.MessageDialog to work with GirCore, but that is deprecated from Gtk 4.10 so it makes sense to use Gtk.AlertDialog.

Another thing I think that would be worth covering is how easy it is for the GirCore tooling to create new C# api for gtk. What happens when gtk 5 comes out for example is the GObject introspection going to make it easy to upgrade and how much manual work needs to be done for things that don't use GObject.

I looked at github issues from GtkSharp and how that project died and left a few upset people stuck on a framework no longer being updated/maintained and there may be people hesitant to use GirCore wondering if they will get stuck with something that is not updating like GtkSharp project did. (I am kind of wondering that myself)

Anyway, just some of my plans for contributing as I learn the basics of GirCore and Gtk and some of my thoughts on what I think new potential users would want to know.

docs/docs/tutorial/gtk/src/HelloWorld/Program.cs
```diff
@@ -0,0 +1,28 @@
...
+    window.Application = (Gtk.Application) sender;
+    
+    // Show the window on the screen.
+    // This makes the window visible to the user.
+    window.Show();
+};
+
+// Start the application's event loop and process user interactions.
+// RunWithSynchronizationContext ensures proper thread synchronization for GTK.
+// The "null" parameter means no custom synchronization context is provided.
```
badcel, line 27: The `null` parameter takes the arguments from the commandline. As there are no arguments supported in our sample the parameter is not filled and thus null.

badcel (commented):
Thanks in general this looks already very good! Please don't add more content to this PR but lets focus on fixing remaining issues so we can get it merged.

I left one small comment regarding the `RunWithSynchronizationContext` method call which needs to be fixed.

Please remove the `Tutorials.sln` file. Instead create inside the `GirCore.sln` solution a new solution folder in the following place: "SolutionItems/docs/docs/tutorial/gtk".

Inside this virtual folder please add all the markdown and image files so the location in the solution folder hierarchy matches the file system hierarchy. Additionally  Please add your new  `csproj` file in the corresponding solution folder so the tutorial project is part of the primary solution.

If this is done please report back for me to check. If the check is successful you need to squash your commits into one big commit as I don't want to have so many commits for the initial tutorial inside the git history.

If this is done it should be ready to be merged.

badcel:
@anthonyirwin82  Can you check what is wrong with the solution file?

anthonyirwin82:
Not sure what happened with the solution file. I did do some manual editing, and the solution file opened in both rider and visual studio without errors.

I have taken the raw solution file from the main branch and made the changes again using only the tooling provided in the ide so that the ide is making the changes to the solution file.

Hopefully this resolves the issue.

badcel (approved):

badcel:
Thanks for your contribution :+1:

PR #1143, opened 2024-12-14
Box Layout Tutorial

anthonyirwin82:

This is a tutorial teaching users how to use the Gtk.Box widget to display multiple widgets at the same time.

It covers:
- Gtk.Box with vertical orientation
- Gtk.Box with horizontal orientation
- Creating both vertical and horizontal spacers to push widgets to the bottom of the window or the end (right) of the window and remain there when the window is resized
- Creating a Gtk.Button with both an icon and a label


badcel (changes requested):
Hey,
sorry for the delayed review. You did a pretty amazing job here :+1: 

I just found two simple typos. If those are fixed the tutorial is good to go :rocket: 

More PRs are welcome :-)

Issue #1198, opened 2025-02-24
Methods to open dialogs should accept nullable parent window


CommonGuy:
The methods to open dialogs, such as `FileDialog#OpenAsync` currently require a `Window` reference. See https://github.com/gircore/gir.core/blob/main/src/Libs/Gtk-4.0/Public/FileDialog.cs. 

The GTK function themselves allow NULL values: https://docs.gtk.org/gtk4/method.FileDialog.open.html

badcel:
Good catch. This could happen because the api is currently implemented manually: https://github.com/gircore/gir.core/blob/main/src/Libs/Gtk-4.0/Public/FileDialog.cs

Feel free to open a PR.

PR #1199, opened 2025-02-24
Allow NULL parent window in async dialog functions

CommonGuy:

Fixes https://github.com/gircore/gir.core/issues/1198


badcel:
Thank you for your contribution :rocket:

PR #1222, opened 2025-03-26
MainLoopSynchronizationContext: Implement Send

badcel:

### DIFF
--- a/src/Libs/GLib-2.0/Internal/MainLoopSynchronizationContext.cs
+++ b/src/Libs/GLib-2.0/Internal/MainLoopSynchronizationContext.cs
@@ -9,18 +9,12 @@ public sealed class MainLoopSynchronizationContext : SynchronizationContext
         => new MainLoopSynchronizationContext();
 
     public override void Post(SendOrPostCallback d, object? state)
-        => ScheduleAction(() => d(state));
-
-    public override void Send(SendOrPostCallback d, object? state)
-        => throw new NotImplementedException();
-
-    private static void ScheduleAction(Action action)
     {
         var proxy = new SourceFuncNotifiedHandler(() =>
         {
             try
             {
-                action();
+                d(state);
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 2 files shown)


badcel:
@shortdevelopment do you have an idea how to test this? I wonder how to create a scenario that shows how  a call will be dispatched to the main thread if it is on secondary thread.

ShortDevelopment:
Maybe something simple like this combined with checking `ManagedThreadId`?

```cs
int counter = 0;
int? a = null, b = null, c = null, d = null;

MainLoopSynchronizationContext ctx = ...;
Thread thread = new(() =>
{
    a = Interlocked.Increment(ref counter);
    ctx.Send(state =>
    {
        b = Interlocked.Increment(ref counter);
        Thread.Sleep(300);
        c = Interlocked.Increment(ref counter);
    }, state: null);
    d = Interlocked.Increment(ref counter);
});
thread.Start();
thread.Join();

a.Should().Be(1);
b.Should().Be(2);
c.Should().Be(3);
d.Should().Be(4);
```

badcel:
The question is more like: how can I verify that the thread executing code before the call to `Send` is not equal to the main thread and the code inside "Send" is equal to the main thread.

badcel:
@ShortDevelopment I think I got it :rocket:

badcel:
@ShortDevelopment Can you confirm that the following code is working for your project?

```CS
 public override void Send(SendOrPostCallback d, object? state)
     {
         var proxy = new SourceFuncAsyncHandler(() =>
         {
             try
             {
                 d(state);
             }
             catch (Exception ex)
             {
                 UnhandledException.Raise(ex);
             }

             return false;
         });

         MainContext.Invoke(MainContextUnownedHandle.NullHandle, proxy.NativeCallback, IntPtr.Zero);
}
```

If yes I would merge this PR as it contains exactly this code and it would be part of 0.7.0. Otherwise the behaviour would only be confirmed by a unit test and not by a real world example. I really would like to avoid to publish a hotfix after the release in case it is broken for you.

PR #1235, opened 2025-04-18
Add and use in tests platform attribute

UrtsiSantsi:

### DIFF
--- a/src/Tests/Shared/PlatformConditionAttribute.cs
+++ b/src/Tests/Shared/PlatformConditionAttribute.cs
@@ -1,0 +1,82 @@
+using System;
+using System.Runtime.InteropServices;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+[Flags]
+public enum Platform
+{
+    Windows32 = 1 << 0,
+    Windows64 = 1 << 1,
+    Windows = Windows32 | Windows64,
+    Linux32 = 1 << 2,
+    Linux64 = 1 << 3,
+    Linux = Linux32 | Linux64,
+    OSX32 = 1 << 4,
+    OSX64 = 1 << 5,
+    OSX = OSX32 | OSX64,
+    Unix32 = Linux32 | OSX32,
+    Unix64 = Linux64 | OSX64,
+    Unix = Unix32 | Unix64,
+}
+
+/// <summary>
+/// This attribute is used to execute a test class or a test
+/// method only if the platform matches, with an optional message.
+/// </summary>
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, Inherited = false)]
+public sealed class PlatformConditionAttribute : ConditionBaseAttribute
+{
+    private readonly Platform _platform;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="PlatformConditionAttribute" /> class.
+    /// </summary>
+    /// <param name="platform">The platform that this test supports.</param>
+    public PlatformConditionAttribute(Platform platform) : base(ConditionMode.Include)
+    {
+        _platform = platform;
+        IgnoreMessage = $"Test is supported only on {platform}";
+    }
+
+    public override bool ShouldRun
+    {
+        get
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                if (Environment.Is64BitOperatingSystem)
+                {
+                    return _platform.HasFlag(Platform.Windows64);
+                }
+
+                return _platform.HasFlag(Platform.Windows32);
+            }
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                if (Environment.Is64BitOperatingSystem)
+                {
+                    return _platform.HasFlag(Platform.Linux64);
+                }
+
+                return _platform.HasFlag(Platform.Linux32);
+            }
+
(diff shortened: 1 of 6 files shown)


src/Tests/Shared/PlatformConditionAttribute.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// <summary>
+    ///     Initializes a new instance of the <see cref="PlatformConditionAttribute" /> class.
+    /// </summary>
+    /// <param name="mode">An enumeration used with to control whether the condition is reversed.</param>
+    /// <param name="platform">The platform that this test supports.</param>
+    public PlatformConditionAttribute(ConditionMode mode, Platform platform)
+        : base(mode)
+    {
+        _platform = platform;
+        IgnoreMessage = $"Test is supported only on {platform}";
```
badcel, line 40: Message does not match the exclude case. Let's just remove this constructor for now but keep the IgnoreMessage.

The constructor to support exclude platforms can be added if needed later.

Did you check how the message is displayed? Is the enum converted into a human readable string?

src/Tests/Shared/PlatformConditionAttribute.cs
```diff
@@ -0,0 +1,93 @@
...
+                return (_platform & Platform.OSX32) != 0;
+            }
+
+            return false;
+        }
+    }
+
+    public override string? IgnoreMessage { get; }
+
+    public override string GroupName => "PlatformCondition";
```
badcel, line 92: Use `nameof` instead a plain string.

src/Tests/Shared/PlatformConditionAttribute.cs
```diff
@@ -0,0 +1,93 @@
...
+                }
+
+                return (_platform & Platform.OSX32) != 0;
+            }
+
+            return false;
+        }
+    }
+
+    public override string? IgnoreMessage { get; }
```
badcel, line 90: I think it is possible to define the ignore message as not nullable but I'm not 100% sure. Can you try?

src/Tests/Shared/PlatformConditionAttribute.cs
```diff
@@ -0,0 +1,93 @@
...
+
+    public override bool ShouldRun
+    {
+        get
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                if (Environment.Is64BitOperatingSystem)
+                {
+                    return (_platform & Platform.Windows64) != 0;
```
badcel, line 60: Please use the [HasFlag](https://learn.microsoft.com/de-de/dotnet/api/system.enum.hasflag?view=net-9.0) method.

badcel (changes requested):
Pretty good work thank you. 🚀 

I have some minor remarks which should improve mainly the maintenance aspect of the code.

After those changes it should be ready to go.

Please amend your changes to avoid creating new commits.

Please confirm that you are fine to release the code either under MIT or LGPL license in the body of your PR text.

src/Tests/Shared/PlatformConditionAttribute.cs
badcel: Yes. The [docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.conditionbaseattribute.groupname?view=mstest-net-3.8#remarks) recommend using `nameof`.

Additionally it makes sense to do so. The property is used to decide how attributes are chained together. So if the group matches the class name different classes are or connected which is fine. If some other attribute should be "and" connected one can use the same class name via `nameof` to keep the group name in sync automatically even if the code the attribute contains is fundamentally changing this requiring a class rename.

The chance of having clashing names is pretty low on General but having it match the class name should lower it even more as it is not some descriptive name anymore.

src/Tests/Shared/PlatformConditionAttribute.cs
```diff
@@ -0,0 +1,82 @@
...
+    OSX32 = 1 << 4,
+    OSX64 = 1 << 5,
+    OSX = OSX32 | OSX64,
+    Unix32 = Linux32 | OSX32,
+    Unix64 = Linux64 | OSX64,
+    Unix = Unix32 | Unix64,
+}
+
+/// <summary>
+///     This attribute is used to execute a test class or a test
```
badcel, line 23: Please only use a single empty space character inside a doc comment like shown in the [docs](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments).

badcel (changes requested):

src/Tests/Shared/PlatformConditionAttribute.cs
badcel: Yeah, I think Rider tries to implicitly adopt to some coding styles used in a project. But sometimes it just suggests things that do not match at all.

Thanks for fixing it this fast 👍

badcel (approved):
Thanks for your work 👍

I need to check it on the PC one more if the solution file is working like intended (I did the review on mobile).

badcel (changes requested):
I checked the solution in Rider. I think it is not necessary to change the `GirCore.sln` file. The new attribute is part of the Test projects through the `Directory.Build.props` file already. This would match how it is done in the  "Generation" folder where a "Shared" folder exists in the file system, too.

badcel:
Thanks a lot for your contribution it improves the readability of the tests quite a bit 🚀.

PR #1238, opened 2025-04-24
Closure: Bind closure to object lifetime

badcel:

### DIFF
--- a/src/Libs/GObject-2.0/Public/Closure.cs
+++ b/src/Libs/GObject-2.0/Public/Closure.cs
@@ -6,12 +6,12 @@ namespace GObject;
 
 public delegate void ClosureCallback(Value? returnValue, Value[] paramValues);
 
-public partial class Closure : IDisposable
+public partial class Closure
 {
     private readonly ClosureCallback _callback;
     private readonly Internal.ClosureMarshal _closureMarshal; //Needed to keep delegate alive
 
-    internal Closure(ClosureCallback callback)
+    internal Closure(ClosureCallback callback, Internal.ObjectHandle handle)
     {
         _callback = callback;
         //The initial state is floating (meaning there is already a ref which is unowned).
@@ -19,7 +19,7 @@ public partial class Closure : IDisposable
         //Afterward sink is called, which might decrement the reference count again by 1 if the instance
         //is not yet sunk. See: https://docs.gtk.org/gobject/method.Closure.sink.html
         Handle = Internal.Closure
-            .NewSimple((uint) Marshal.SizeOf<Internal.ClosureData>(), IntPtr.Zero)
+            .NewObject((uint) Marshal.SizeOf<Internal.ClosureData>(), handle.DangerousGetHandle())
             .OwnedCopy();
 
         Debug.WriteLine($"Instantiating Closure: Address {Handle.DangerousGetHandle()}.");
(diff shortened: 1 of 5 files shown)


badcel:
@UrtsiSantsi: I tried to reproduce https://github.com/PintaProject/Pinta/issues/1340 but failed.

In case you can still reproduce: Can you confirm or or deny that this PR is fixing the issue?

You would need to build Pinta against GirCore with this branch `fix-1234` checked out.

badcel:
@cameronwhite: Any thoughts on this?

Using [g_closure_new_object](https://docs.gtk.org/gobject/ctor.Closure.new_object.html) results in calls to [g_object_watch_closure](https://docs.gtk.org/gobject/method.Object.watch_closure.html) wich increases the ref count of an object during closure invocation thus preventing ref count becomming 0 and which calls [g_closure_invalidate](https://docs.gtk.org/gobject/method.Closure.invalidate.html) which ignores any future calls to this closure after object finalization.

cameronwhite:
I've never been able to reproduce this reliably, so hopefully others can test it out!

I'm not very familiar with the signals implementation so I'm still trying to figure out how it all works.
If the error was `A callback was made on a garbage collected delegate of type 'GObject-2.0!GObject.Internal.ClosureMarshal::Invoke`, this means that the delegate is being invoked from native code after the managed Closure object has already been garbage collected. Wouldn't that mean that the native sender object is still alive and emitting signals, so `g_object_watch_closure()` wouldn't have disabled the closure yet?

badcel:
Let's go into detail here a little more:
The implementation between a callback parameter and signals are not related. Parameter callbacks manage their lifetime themself depending on their scope (`call`, `notified`, `async`, `forever`).

Signals on the other hand are invoked on the native side via a Closure. The [GirCore Closure](https://github.com/gircore/gir.core/blob/main/src/Libs/GObject-2.0/Public/Closure.cs) class is the managed counter part which receives those calls.

The lifetime of the managed closure instance is bound to the lifetime of the managed object which should receive the signal: So every signal that's connected on an instance has a managed closure instance that is [saved](https://github.com/gircore/gir.core/blob/36def08e4e253b2c32020c7eea761a48c5215107/src/Libs/GObject-2.0/Public/Object.Signals.cs#L9) inside the managed object as a dictionary entry.

There are 3 possibilities to remove a closure from this dictionary:
1. [Disconnecting](https://github.com/gircore/gir.core/blob/36def08e4e253b2c32020c7eea761a48c5215107/src/Libs/GObject-2.0/Public/Object.Signals.cs#L22) a signal 
2. [Manual disposal](https://github.com/gircore/gir.core/blob/36def08e4e253b2c32020c7eea761a48c5215107/src/Libs/GObject-2.0/Public/Object.cs#L22) of the instance
3. The instance got garbage collected thus removing the dictionary completely including the closure instances.

I suspect that there can be some race condition between the last ref of an object and an additional closure invocation.

So the new code should improve scenario 3.

But after writing this text I would suspect that it makes sense to [invalidate](https://docs.gtk.org/gobject/method.Closure.invalidate.html) the closure manually in scenario 1 and 2 because any further calls into the closure will be invalid if the native world keeps a ref of the closure around for some reason but it is already gone on managed side.

What do you think?

cameronwhite:
Thanks for the info!
One thing that comes to mind is the ToggleRef changes where we release the native object on the main UI thread rather than from the GC thread. I think that could be a scenario where the managed object is already gone while the native object is still alive?

badcel:
Yeah I think this change and the proposed changes could improve the situation.

But perhaps I need to do the rewrite either way. So if the callback is received for a native instance I could check the instance cache to verify if there is still a managed instance around and if there is an instance invoke the signal.

cameronwhite:
Making sure the closures are invalidated sounds like a good approach to me - checking the instance cache is definitely safer, but might add some small overhead to every signal invocation?

badcel:
Yeah I don't know how the dotnet runtime handles this internally as I pass an address to native code but the method is invoked on the correct instance automatically. So there must be something happening internally to execute the callback on the correct instance.

Doing it manually will result in a dictionary lookup for each received signal.

If I can avoid the rewrite for now it would be fine for me. This depends on @UrtsiSantsi feedback.

In general I would like to apply the fixes on by one each time having a new feedback round to get a feeling where the problem is located instead of doing it all in one sweep.

PR #1251, opened 2025-05-14
Use Combinatorial.MSTest

Youssef1313:


badcel:
Thanks for the PR. What do you think are the big benefits of using the library?

At a first glance it does not add much value, and at the same time introduces an additional dependency.

Currently I'm not convinced merging this.

Youssef1313:
For your case, it just simplifies few `[DataRow(true)] [DataRow(false)]`. Whether or not to accept the PR is definitely up to you. The library is very similar in nature to (and is mostly forked from) [xunit.combinatorial](https://github.com/AArnott/Xunit.Combinatorial), which is a very popular library already and is widely used by xunit users.

Issue #1243, opened 2025-05-07
Remove Parabolic from App list


badcel:
https://github.com/gircore/gir.core/blob/main/docs/docs/apps.md

PR #1253, opened 2025-05-16
Remove Parabolic from App list

Lemon73-Computing:

Close: #1243


badcel:
Thank you!

Issue #1053, opened 2024-04-30
dependabot multi directory configuration


badcel:
See: https://github.blog/changelog/2024-04-29-dependabot-multi-directory-configuration-public-beta-now-available/

PR #1254, opened 2025-05-16
CI: Introduce dependabot multi directory config

Lemon73-Computing:

Close: #1053

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,13 +5,10 @@ updates:
     schedule:
       interval: "daily"
   - package-ecosystem: "nuget"
-    directory: "/src/Extensions/Integration"
-    schedule:
-      interval: "daily"
-      
-  # This traces the dependencies through the dependency chain:
-  # Generator3, GirModel, GirLoader
-  - package-ecosystem: "nuget"
-    directory: "/src/Generation/GirTool"
+    directory:
+      - "/src/Extensions/Integration"
+        # This traces below the dependencies through the dependency chain:
+        # Generator3, GirModel, GirLoader
+      - "/src/Generation/GirTool"
     schedule:
       interval: "daily"


.github/dependabot.yml
```diff
@@ -5,13 +5,10 @@ updates:
...
   - package-ecosystem: "nuget"
-    directory: "/src/Extensions/Integration"
-    schedule:
-      interval: "daily"
-      
-  # This traces the dependencies through the dependency chain:
-  # Generator3, GirModel, GirLoader
-  - package-ecosystem: "nuget"
-    directory: "/src/Generation/GirTool"
+    directory:
```
badcel, line 8: I think you used the wrong keyword. See https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#directories-or-directory--

badcel (commented):
Thanks for the PR. Please see my comment.

Lemon73-Computing:
@badcel
Thanks for review. Now, I fixed it!

badcel:
I think the folders are not matching the code?

I don't have knowledge how this works in detail and would need to read the docs so I can't tell if the change is correct.

It would be great if you do the research on this topic. It would be great to get PRs for outdated nuget packages for any affected project including tests.

Lemon73-Computing:
> I think the folders are not matching the code?

Oh, I couldn't notice it. I will change `src/Extensions/Integration` to `src/Extensions/GObiject-2.0.Integration`.

badcel:
Do you know if the folders "find" projects inside? Or must every project be listed one by one? Because GirCore has lots of csproj files.

Lemon73-Computing:
> Do you know if the folders "find" projects inside? Or must every project be listed one by one? Because GirCore has lots of csproj files.

I have no experience with that, but might be able to do it by using the wildcard character `*`. I will try it.

> [!note]
> The `directories` key supports globbing and the wildcard character `*`. These features are not supported by the `directory` key.
> 
> (https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#directories-or-directory--)

Lemon73-Computing:
I just realised that if dependabot target `.sln` it seems to work.

badcel:
You did not use either the "sln" file nor an globing pattern with "*". Do you think this will work as it is?

Lemon73-Computing:
I tried it on my Fork and it worked well.

(Dependabot does not seem to need to specify `.sln`, as it specifies a directory)

(I found that it works by specifying a directory that contains `.sln`, even if `*` is not used, so I specify it that way. If this doesn't work, I might want to use `- "/src/*/*"`, `- "/src/*/*/*"`)

Lemon73-Computing:
@badcel OK, now it can merge into `main` branch!

badcel:
Thank you

Issue #1258, opened 2025-05-23
Verify: Long values in properties


badcel:
Using `long` values in properties is probably broken as GObject.Value uses `long` instead of `CLong`.

cameronwhite:
Some findings from looking into this a bit:
- A `gint64` property in C should translate into `long` in C#. Like #1259 this doesn't work because GObject.Value is using `SetLong()` and similar methods rather than `SetInt64()`
- If you do have a `glong` property on the C side, this also translates into `long` on the C# side which is a problem. If the C# side uses GObject.Value.SetInt64(), this fails on the C side when using `gobject_value_get_long()` since `gint64` and `glong` are unique types
- I expect a `gulong` property would have the same issue.

badcel:
The translation is probably wrong:
- [Gint64](https://docs.gtk.org/glib/types.html#gint64) is correct to be translated to dotnet `long` as both are 64 bit in all cases.
- [glong](https://docs.gtk.org/glib/types.html#glong) should probably translate to dotnet `CLong` like it is already done for parameters and return types.

Issue #1257, opened 2025-05-23
UInt64 properties produce errors at runtime


cameronwhite:
To reproduce:
```csharp
Gdk.MemoryTextureBuilder builder = new ();
builder.Stride = 256ul;
```

produces the following error:
```
System.NotSupportedException: Type System.UInt64 is not supported as a value type
      Stack Trace:
         at GObject.Value.Set(Object value) in /Users/cameron/code/gir.core/src/Libs/GObject-2.0/Public/Value.cs:line 200
         at GObject.Property`2.Set(K obj, T value) in /Users/cameron/code/gir.core/src/Libs/GObject-2.0/Public/Property.cs:line 61
         at Gdk.MemoryTextureBuilder.set_Stride(UInt64 value) in /Users/cameron/code/gir.core/src/Libs/Gdk-4.0/Public/MemoryTextureBuilder.Properties.Generated.cs:line 68
```

A workaround is to use the corresponding methods instead, e.g. `builder.SetStride ((nuint) value)`

PR #1259, opened 2025-05-23
Add missing support for ulong in GObject.Value to fix property accessors

cameronwhite:
As bug #1258 notes, we also should set up tests for `long` values as well

Bug: #1257

### DIFF
--- a/src/Tests/Libs/GirTest-0.1.Tests/PropertyTest.cs
+++ b/src/Tests/Libs/GirTest-0.1.Tests/PropertyTest.cs
@@ -100,6 +100,18 @@ public class PropertyTest : Test
         obj.BooleanValue.Should().Be(b);
     }
 
+    [DataTestMethod]
+    [DataRow(42ul)]
+    [DataRow(ulong.MinValue)]
+    [DataRow(ulong.MaxValue)]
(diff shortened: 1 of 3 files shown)


src/Tests/Libs/GirTest-0.1.Tests/PropertyTest.cs
```diff
@@ -100,6 +100,16 @@ public void TestBooleanProperty(bool b)
         obj.BooleanValue.Should().Be(b);
     }
 
+    [DataTestMethod]
+    [DataRow(42ul)]
```
badcel, line 104: Can you add [Uint64.MaxValue](https://learn.microsoft.com/en-us/dotnet/api/system.uint64.maxvalue?view=net-9.0) and Uint64.MinValue to the tests?

badcel (dismissed):
Thanks, looks good already can you add some more data to the tests? See comment.

src/Tests/Libs/GirTest-0.1.Tests/PropertyTest.cs
cameronwhite: Sure, I've added that in now 👍

PR #1263, opened 2025-05-30
Rsvg-2.0.gir

czirok:

Rsvg 2.0


src/Libs/Rsvg-2.0/Public/Module.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// Calling this method is necessary to correctly initialize the bindings
+    /// and should be done before using anything else in the <see cref="Rsvg" />
+    /// namespace.
+    /// </para>
+    /// <para>
+    /// Calling this method will also initialize the modules this module
+    /// depends on:
+    /// </para>
+    /// <list type="table">
+    /// <item><description><see cref="GObject.Module" /></description></item>
```
badcel, line 21: The dependencies are not completely listed inside the doc comments.

badcel (changes requested):
Thanks for the PR. Looks almost perfect 🚀  The only thing I found is the incomplete doc comment.

Could you amend the changes to keep only one commit for this PR?

Then I think this PR is good to go 👍.

Issue #1215, opened 2025-03-14
Migrate to slnx solution file format


badcel:
https://devblogs.microsoft.com/dotnet/introducing-slnx-support-dotnet-cli/

PR #1292, opened 2025-08-06
#1215: Migrate to slnx solution file format

kashifsoofi:


badcel:
Is slnx already supported oficially by dotnet?

kashifsoofi:
> Is slnx already supported oficially by dotnet?

Yes it seems to be the case, since sdk version 9.0.200, build is successful as well

badcel:
Hm it seem to work for dotnet 8, too.

I will need to test this with Rider and read a little bit in the Microsoft docs before merging.

badcel:
It reads like it is not generally available. So I would wait a little more before merging this.

badcel:
Just checked it one more time in Rider and it works like expected :-).

Thanks again for your contribution and ongoing support :rocket:

Issue #1042, opened 2024-04-12
Switch to new MSTest runner


badcel:
https://devblogs.microsoft.com/dotnet/introducing-mstest-sdk/

PR #1299, opened 2025-08-08
#1042: Use MSTest.Sdk

kashifsoofi:


badcel:
Thank you 👍

Is there a way to specify the Version of the sdk in a file? Like mentioned in the docs via `global.json`.

Updating the sdk Version would otherwise be a little cumbersome.

It should be possible to remove some other properties: https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-sdk

I think the most recent version is 10.1 according to [nuget](https://www.nuget.org/packages/MSTest.Sdk).

badcel:
Can you mark the PR as a draft as long as it still work in progress?

PR #1306, opened 2025-08-22
GLib.Bytes: Add GetRegionSpan method

badcel:


AeonLucid:
> @AeonLucid: This is what I would start with. If you want to get all data this is still on the users side.
> 
> 
> 
> If we want to provide a method to return all data it should be based on [get_data](https://docs.gtk.org/glib/method.Bytes.get_data.html) which is not yet generated for some reason.

I haven't tested it but looks good to me!

PR #1321, opened 2025-09-20
Add Pango CSS scale factor constants

ousnius:
See header https://gitlab.gnome.org/GNOME/pango/-/blob/main/pango/pango-font.h#L261

* Pango.Constants.XX_SMALL
* Pango.Constants.X_SMALL
* Pango.Constants.SMALL
* Pango.Constants.MEDIUM
* Pango.Constants.LARGE
* Pango.Constants.X_LARGE
* Pango.Constants.XX_LARGE

Other wrappers like GtkSharp or Vala have also introduced these as custom code.
e.g. https://gitlab.gnome.org/GNOME/vala/-/blob/main/vapi/metadata/Pango-1.0-custom.vala

### DIFF
--- a/src/Libs/Pango-1.0/Public/Constants.cs
+++ b/src/Libs/Pango-1.0/Public/Constants.cs
@@ -1,0 +1,39 @@
+﻿namespace Pango;
+
+public partial class Constants
+{
+    /// <summary>
+    /// The CSS scale factor for three shrinking steps (1 / (1.2 * 1.2 * 1.2)).
+    /// </summary>
+    public const double XX_SMALL = 0.5787037037037;
+
+    /// <summary>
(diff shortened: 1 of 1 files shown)


badcel:
I wonder why the constants are not generated?

badcel:
Can you please put the calculation into the code, instead of the result? It will still be a constant expression and the compiler will use full precission for calculations.

ousnius:
> Can you please put the calculation into the code, instead of the result? It will still be a constant expression and the compiler will use full precission for calculations.

0.5787037037037 -> 0.5787037037037037
0.6944444444444 -> 0.6944444444444444
0.8333333333333 -> 0.8333333333333334

badcel:
Can you squash and it's good to go 🚀.

Issue #1191, opened 2025-02-15
can't get text iterators from a TextBuffer


medovina:
Gir.Core doesn't seem to expose any methods that would return a Gtk.TextIter from a Gtk.TextBuffer.  This seriously limits the set of operations that a program may perform with a TextBuffer.  For example:

- [gtk_text_buffer_get_selection_bounds()](https://docs.gtk.org/gtk4/method.TextBuffer.get_selection_bounds.html) has no binding, so I can't retrieve TextIter objects at the start and end of the current selection.

- [gtk_text_buffer_get_iter_at_line()](https://docs.gtk.org/gtk4/method.TextBuffer.get_iter_at_line.html) has no binding, so I can't get an iterator at a given line number.

- [gtk_text_buffer_get_iter_at_mark()](https://docs.gtk.org/gtk4/method.TextBuffer.get_iter_at_mark.html) has no binding, so I can't convert a mark (such as the current cursor position) to a TextIter.

- [gtk_text_buffer_get_start_iter()](https://docs.gtk.org/gtk4/method.TextBuffer.get_start_iter.html) and [gtk_text_buffer_get_end_iter()](https://docs.gtk.org/gtk4/method.TextBuffer.get_end_iter.html) have no binding, so I can't get iterators at the start/end of the buffer.

badcel:
If you open issues which directly reference some c api please insert links to the documentation. This makes life for me easier.

medovina:
OK, no problem.  I just edited my posting above and added links to it.

Issue #1120, opened 2024-09-12
record parameter with direction != in not yet supported. 


BehroozAmoozad:
I'm getting a bunch of these warnings when generating code. Is there any way for me or you to fix the issue? Even manually for the couple functions that I need will be really helpful. I don't know where to start and what to modify. point me in the right direction and I'll do it.

What brought me here is that I can't initialize ListStores and a whole lot of treemodel code that I converted from gtk 3 stopped working which I removed thinking it's no longer supported in gtk 4.

```
[16:24:54 WRN] Did not generate method 'TreeModel.Get': Variadic parameters are not yet supported
[16:24:54 DBG] Did not generate method 'TreeModel.GetIter': GirModel.AnyType: GirModel.Type: Gtk.TreeIter: record parameter 'iter' with direction != in not yet supported
[16:24:54 DBG] Did not generate method 'TreeModel.GetIterFirst': GirModel.AnyType: GirModel.Type: Gtk.TreeIter: record parameter 'iter' with direction != in not yet supported
[16:24:54 DBG] Did not generate method 'TreeModel.GetIterFromString': GirModel.AnyType: GirModel.Type: Gtk.TreeIter: record parameter 'iter' with direction != in not yet supported
[16:24:54 DBG] Did not generate method 'TreeModel.GetValue': GirModel.AnyType: GirModel.Type: GObject.Value: record parameter 'value' with direction != in not yet supported
[16:24:54 DBG] Did not generate method 'TreeModel.IterChildren': GirModel.AnyType: GirModel.Type: Gtk.TreeIter: record parameter 'iter' with direction != in not yet supported
[16:24:54 DBG] Did not generate method 'TreeModel.IterNthChild': GirModel.AnyType: GirModel.Type: Gtk.TreeIter: record parameter 'iter' with direction != in not yet supported
[16:24:54 DBG] Did not generate method 'TreeModel.IterParent': GirModel.AnyType: GirModel.Type: Gtk.TreeIter: record parameter 'iter' with direction != in not yet supported

```

BehroozAmoozad:
I did something. but now the internal function is not generated. I have no idea where that one is.  

```
        if (parameter.Parameter.Direction == GirModel.Direction.In)
        {
            var record = (GirModel.Record) parameter.Parameter.AnyTypeOrVarArgs.AsT0.AsT0;
            var typeHandle = Model.TypedRecord.GetFullyQuallifiedHandle(record);
[...]

Issue #932, opened 2023-08-30
Support Out parameters for Records and Classes


badcel:
Support Out parameters for Records and Classes.

Implementation for both should be similar. This could require heavy refactorings as out parameters inverse the direction of flow similar to return values. Out-Parameters in a callback would double inverse the direction of flow.

One idea would be to have a completely new set of converters for `out` datatypes in addition to the currently available converters for "ToNative" / "ToManaged".

Keep in mind that there are already converters for primitive value types, which support out parameters. Those should probably be split up into two converters to align with the new converters for classes / records.

PR #1324, opened 2025-09-21
Generate methods with typed records as out parameter (non-nullable, no transfer)

ousnius:
* Only non-nullable out parameters without ownership transfer for now
* Nullable out-parameters don't make much sense in C# in the traditional way
* Added new native test
* Fixed a few typos

See #1120, #1191 and #932 to see what part of them is solved by this.

### DIFF
--- a/src/Native/GirTestLib/girtest-typed-record-tester.c
+++ b/src/Native/GirTestLib/girtest-typed-record-tester.c
@@ -99,7 +99,7 @@ girtest_typed_record_tester_nullable_mirror(GirTestTypedRecordTester *data, gboo
 }
 
 /**
- * girtrest_typed_record_tester_unref:
+ * girtest_typed_record_tester_unref:
  * @self: (transfer full): a `GirTestTypedRecordTester`
  *
  * Decrements the reference count on `data` and frees the
@@ -222,6 +222,18 @@ int girtest_typed_record_tester_get_ref_count_sum_nullable(GirTestTypedRecordTes
         return -1;
 
     return girtest_typed_record_tester_get_ref_count_sum(data, size);
+}
+
+/**
+ * girtest_typed_record_tester_out_no_ownership_transfer_non_nullable_caller_allocates:
+ * @data: (not nullable) (out caller-allocates) (transfer none): a `GirTestTypedRecordTester`
+ **/
+void
+girtest_typed_record_tester_out_no_ownership_transfer_non_nullable_caller_allocates (GirTestTypedRecordTester *data)
+{
+    g_return_if_fail (data != NULL);
+
+    data->custom_string = g_strdup("OutParameter");
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/Native/GirTestLib/girtest-typed-record-tester.c
```diff
@@ -224,6 +224,18 @@ int girtest_typed_record_tester_get_ref_count_sum_nullable(GirTestTypedRecordTes
     return girtest_typed_record_tester_get_ref_count_sum(data, size);
 }
 
+/**
+ * girtest_typed_record_tester_out_no_ownership_transfer:
+ * @data: (out) (transfer none): a `GirTestTypedRecordTester`
```
badcel, line 229: A litte more explicit:

```suggestion
 * @data: (not nullable) (out caller-allocates) (transfer none): a `GirTestTypedRecordTester`
```

src/Native/GirTestLib/girtest-typed-record-tester.c
```diff
@@ -224,6 +224,18 @@ int girtest_typed_record_tester_get_ref_count_sum_nullable(GirTestTypedRecordTes
     return girtest_typed_record_tester_get_ref_count_sum(data, size);
 }
 
+/**
+ * girtest_typed_record_tester_out_no_ownership_transfer:
```
badcel, line 228: Perhaps name the method: `girtest_typed_record_tester_out_no_ownership_transfer_non_nullable_caller_allocates`?

src/Tests/Libs/GirTest-0.1.Tests/TypedRecordTest.cs
```diff
@@ -127,6 +127,13 @@ public void SupportsParameterNullableTransferNone()
         result.Should().Be(1);
     }
 
+    [TestMethod]
+    public void SupportsOutParameterTransferNone()
```
badcel, line 131: Adapt name similar to C method name with nullable / caller allocates info.

src/Native/GirTestLib/girtest-typed-record-tester.c
```diff
@@ -224,6 +224,18 @@ int girtest_typed_record_tester_get_ref_count_sum_nullable(GirTestTypedRecordTes
...
+ * @data: (out) (transfer none): a `GirTestTypedRecordTester`
+ **/
+void
+girtest_typed_record_tester_out_no_ownership_transfer (GirTestTypedRecordTester *data)
+{
+    g_return_if_fail (data != NULL);
+
+    data->custom_string = g_strdup("Hello");
```
badcel, line 236: Please change the string to a value which is different from the default.

badcel (commented):
This is fantastic work! Great job it's exactly like I would have done it.

My comments are only minor things which are not absolutely needed, but will probably help later on during development.

PR #1342, opened 2025-11-14
Add WebCamControl

Daniel15:
Simple app but I had to figure out quite a bit of stuff for it, so the code might be useful to other people too.


badcel:
Looks good thanks. Just 2 remarks:
- Please add an alt text for the image (compare pinta image).
- Please squash your commits into one.

If this is fixed it's good to go 👍

Daniel15:
@badcel This should be good to go now! Sorry for the delay.

badcel:
No problem. I think I will update the homepage more often in the future. At least with every preview version, too.

PR #1345, opened 2025-11-17
For TaskCompletionSource use Try-methods to avoid invalid operation

gfoidl:

So far the `SetResult`, etc. methods are used and this can result in
```
System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed.
   at System.Threading.Tasks.TaskCompletionSource`1.SetResult(TResult result)
   at Gtk.FontDialog.<>c__DisplayClass1_0.<ChooseFaceAsync>b__0(Object sourceObject, AsyncResult res, IntPtr data)
   at Gio.Internal.AsyncReadyCallbackAsyncHandler.<.ctor>b__3_0(IntPtr sourceObject, IntPtr res, IntPtr data)
   at Gio.Internal.Application.Run(IntPtr application, Int32 argc, String[] argv)
   at Gio.Application.Run(Int32 argc, String[] argv)
   at Gio.Application.RunWithSynchronizationContext(String[] args)
   at Program.<Main>$(String[] args)
```
when it is tried to call the Set-method more than once. Here with a FontDialog that is "Cancel"led by user.
Looking at the implementation I don't see a bug here, so maybe the native side invokes the callback more than once.

Anyway, by using the `TrySet`-methods on TCS this won't happen.

### DIFF
--- a/src/Libs/Gdk-4.0/Public/Clipboard.cs
+++ b/src/Libs/Gdk-4.0/Public/Clipboard.cs
@@ -14,16 +14,16 @@ public partial class Clipboard
         {
             if (sourceObject is null)
             {
-                tcs.SetException(new Exception("Missing source object"));
+                tcs.TrySetException(new Exception("Missing source object"));
                 return;
             }
 
             var readValue = Internal.Clipboard.ReadTextFinish(sourceObject.Handle.DangerousGetHandle(), res.Handle.DangerousGetHandle(), out var error);
 
             if (!error.IsInvalid)
-                tcs.SetException(new GLib.GException(error));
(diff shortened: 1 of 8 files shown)


badcel:
Is this happening with the font dialog sample for you?

https://github.com/gircore/gir.core/tree/main/src/Samples/Gtk-4.0/FontDialog

badcel:
This is probably the correct fix: #1350

badcel:
Yeah I think it is not needed currently.

Later on the code will be auto generated and bugs like this should not happen anymore.

gfoidl:
Closing, as https://github.com/gircore/gir.core/pull/1350 is the proper fix for the bug.

PR #1359, opened 2025-11-24
Add GDK Win32

badcel:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,7 +120,7 @@ jobs:
     
     steps:
       - name: Install dependencies
-        run: sudo dnf -y upgrade && sudo dnf -y install meson gobject-introspection-devel git gcc glib2-devel gdk-pixbuf2 gdk-pixbuf2-modules cairo-gobject libicu gh
+        run: sudo dnf -y upgrade && sudo dnf -y install meson gobject-introspection-devel git gcc glib2-devel gdk-pixbuf2 cairo-gobject libicu gh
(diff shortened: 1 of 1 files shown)


src/Libs/GdkWin32-4.0/Win32/Fence.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
+
+namespace Win32;
+
+[UnsupportedOSPlatform("Linux")]
+[UnsupportedOSPlatform("OSX")]
```
gfoidl:
```suggestion
[SupportedOSPlatform("windows")]
```
* is more direct
* browser, and all other would need to be set as unspported otherwise too

gfoidl (commented):
Same on other structs too.

src/Libs/GdkWin32-4.0/Win32/Fence.cs
gfoidl: When the supported OS are linux, windows, and MacOS, then for
> with all 3 supported operating systems.
you don't need any annotation at all.
Then it's a matter of convention: whether annotate unsupported ones (deny-list like) or annotate when a type / member is specific for one OS (where the analyzer automatically flags all other OSes as unsupported).

> web assembly feel free to try. If you got it running let's talk about missing annotations
OK 🤣 (fun aside, I guess it could work by cross-compiling GObject to wasm, but for what use case?)

Issue #1180, opened 2025-02-09
allow ColumnView sorting via CustomSorter with CompareDataFunc


medovina:
I'd like to implement a Gtk.ColumnView with sortable columns.

One way to do this is to create a Gtk.CustomSorter object.  To build one, I need to provide a CompareDataFunc, which is defined like this:

```
public delegate int CompareDataFunc(nint a, nint b)
```

If the arguments a and b had type `object`, I could cast them to the appropriate type and perform my own comparison.  But they have type `nint`.  Shouldn't `nint` be `object` here?  Or is there some way (e.g. using unsafe code) that I can cast an `nint` to an `object`?

badcel:
Unfortunately I can't tell you the best way how to solve it as I'm not a GTK dev. So I don't know the alternatives.

At least in regard to the API it is correct as it is what the API says: https://docs.gtk.org/glib/callback.CompareDataFunc.html

Perhaps there could be some optimisation later on but this would need to be thought through carefully.

There recently was a [discussion](https://github.com/gircore/gir.core/discussions/1141). But I don't know the outcome of it.

It would be great to see a C sample how to do properly sort a column. From the docs I would think it should be a [ColumnViewSorter](https://docs.gtk.org/gtk4/class.ColumnViewSorter.html). If I have sample I could take a look what is needed to make it work as this is the 2nd request for this. But I really need more information to keep things efficient for me. If I have to do all the GTK stuff myself I prefer to do other things with less overhead first.

badcel:
As `nint` is just a pointer you are automatically on a lower level than the regular binding as you would need to create dotnet objects from pointers first.

Perhaps there could be some convenience API here to make life easier but then I would need to know more details:

If you manage to get the `CustomSorter` to work and provide a minimal sample I can figure out how to make life easier for you in a later GirCore version.

But again without any support I'm currently not willing to invest my time into it. You can check out the milestones to see the roadmap. There are more than enough points for the coming months and probably years.

So if you care to fundamentally improve the column sorter experience for GirCore let's do this together. I can support with GirCore details and support finding a solution but your part would be the analysis of the GTK column view sorter possibilities so I get an overview on how to do sorting with different c samples. Then I can pretty verify what is needed to get it working and start working on it or even support you getting your hands dirty in GirCore if you care.

medovina:
Thanks for the quick replies.

I changed the title of this issue to be a bit more descriptive.

[...]

PR #1375, opened 2025-12-05
Gtk: Add a generic constructor to CustomSorter

badcel:
This allow to work directly which GObject.Object data in the CompareDataFunc.

See the added sample.

Fixes: #1180


medovina:
Great, thanks!

PR #1423, opened 2026-01-04
GdkWayland-4.0

### LINKED ISSUE #1408: No way to get Wayland Toplevel handle
Applications wanting to work with XDG Desktop Portals frequently need to provide a X11 or Wayland window ID. Currently there is no way to get the handle string of a toplevel on Wayland. Gdk provides the function `gdk_wayland_toplevel_export_handle` and `gdk_wayland_toplevel_unexport_handle` but similarly to getting Win32 native handles this would require adding support for the GdkWayland library.

pancakesmeow:

Closes #1408

### DIFF
--- a/src/Libs/GdkWayland-4.0/GdkWayland-4.0.csproj
+++ b/src/Libs/GdkWayland-4.0/GdkWayland-4.0.csproj
@@ -1,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <PackageId>GirCore.GdkWayland-4.0</PackageId>
+    <RootNamespace>GdkWayland</RootNamespace>
+    <Description>C# bindings for GdkWayland.</Description>
+  </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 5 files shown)


src/Libs/GdkWayland-4.0/GdkWayland-4.0.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
```
badcel:
```suggestion
    <PackageId>GirCore.GdkWayland-4.0</PackageId>
```

Please do not specify a target framework. This is done automatically through `GirCore.Libraries.props`. Instead please define the PackageId to define the name of the resulting nuget package.

badcel (changes requested):
This looks almost good to go 🚀 

- Please remove the target framwork and add the PackageId like suggested.
- Please amend your changes in git, so that no 2nd commit is created for this trivial change. I want to keep the commit history clean.

PR #1426, opened 2026-01-06
Added GirCore1005 analyser

hol430:
Due to the overlap (adjacency, really) with GirCore1004, I added unit tests to ensure that classes which would emit GirCore1004 don't emit GirCore1005, and vice versa. Not sure if this is strictly necessary - I left it in for now but am happy to remove it.


badcel:
It’s perfectly fine to check that there is no overlap 👍

The only thing I see that is missing is a new file file for the Homepage like this: https://github.com/gircore/gir.core/blob/main/docs/docs/integration/diagnostic/1004.md

you can just copy the Texts from the descriptor.

and a link to the new file here: https://github.com/gircore/gir.core/blob/main/docs/docs/integration/index.md

(The diagnostic link is referring to the homepage and IDEs display the link inside the popup for the diagnostic message)

Please amend your changes to avoid a second commit.

Everything else looks ready to be merged, thank you 🚀

PR #1435, opened 2026-01-07
GObject-2.0.Integration: Subclass - Ensure constructor initialization

badcel:


badcel:
Superseded by #1455.

PR #1461, opened 2026-02-15
Add async sample for updating UI in tasks and using synchronization context

ousnius:
Add sample project "Async.csproj" for processing UI updates from asynchronous code with await (uses syncronization context) or tasks using two methods:

1. Manually creating a task from anywhere that doesn't necessarily use async/await patterns.
  - This method is used by the "runTaskButton" and calls into "Work.DoHeavyWorkWithInvoke".
  - Any UI updates need to be done in a delegate invoked by the UI queue later on (see "Helper.Invoke").

2. Using the gir.core MainLoopSynchronizationContext that makes sure any awaited code is queued in the correct order and then run on the UI thread.
  - This method is used by the "runAsyncButton" and calls into "Work.DoHeavyWorkAsync".
  - Any UI updates are added to the UI queue automatically when awaited.

There are two sets of "run" and "cancel" buttons (one for each method).
Progress is displayed using a Gtk.ProgressBar and Gtk.Spinner.

### DIFF
--- a/src/Samples/Gtk-4.0/Async/Program.cs
+++ b/src/Samples/Gtk-4.0/Async/Program.cs
@@ -1,0 +1,186 @@
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Async;
+
+/*
+ * Sample for processing UI updates from asynchronous code or tasks using two methods:
+ * 
+ * 1. Using the GirCore MainLoopSynchronizationContext that makes sure any awaited code is queued in the correct order and then run on the UI thread.
+ *   - This method calls into "Work.WaitForStuffAsync".
+ *   - Any UI updates are added to the UI queue automatically when awaited.
+ * 
+ * 2. Manually creating a task from anywhere that doesn't necessarily use async/await patterns.
+ *   - This method calls into "Work.DoHeavyWorkWithInvoke" and would be doing some heavy computations.
+ *   - Any UI updates need to be done in a delegate invoked by the UI queue later on (see "Invoke").
+ */
+
+var application = Gtk.Application.New("org.gir.core", Gio.ApplicationFlags.FlagsNone);
+application.OnActivate += (sender, args) =>
+{
+    var window = Gtk.ApplicationWindow.New((Gtk.Application) sender);
+    window.Title = "Async Sample";
+    window.SetDefaultSize(800, 250);
+
+    var hbox = Gtk.Box.New(Gtk.Orientation.Horizontal, 5);
+
+    SampleBox asyncSample = SetupSampleBox("Async");
+    hbox.Append(asyncSample.Box);
+
+    SampleBox tasksSample = SetupSampleBox("Tasks");
+    hbox.Append(tasksSample.Box);
+
+    /*
+     * Sample Part 1
+     * 
+     * 'OnClicked' event handler using awaits that are queued
+     * and kept in sync automatically by GirCore using a synchronization context.
+     */
+    asyncSample.RunButton.OnClicked += async (button, args) =>
+    {
+        if (asyncSample.WorkRunning)
+            return;
+
+        asyncSample.WorkRunning = true;
+        asyncSample.CancelTokenSrc = new CancellationTokenSource();
+
+        asyncSample.RunButton.Visible = false;
+        asyncSample.CancelButton.Visible = true;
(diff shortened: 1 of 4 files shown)


src/Samples/Gtk-4.0/Async/Helper.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+
+namespace Async
```
badcel:
```suggestion
namespace Async;
```

Can you use file scoped namespaces?

src/Samples/Gtk-4.0/Async/Work.cs
```diff
@@ -0,0 +1,88 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+
+namespace Async
```
badcel: File scoped namespace please.

badcel (changes requested):
Hey thanks for the sample. The topic is a bit complex. Please see my lengthy  comment in `Program.cs`.

src/Samples/Gtk-4.0/Async/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+            return Work.DoHeavyWorkWithInvoke(progressBarTask, 8, taskCancelTokenSrc.Token);
+#pragma warning restore CS4014
+        });
+        task.ContinueWith((task) =>
+        {
+            /*
+             * "ContinueWith" is asynchronous as well.
+             * UI updates need to be invoked from the main thread.
+             */
+            Helper.Invoke(() =>
```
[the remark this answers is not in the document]
badcel: The pragma warning should get obsolete with this, too

src/Samples/Gtk-4.0/Async/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+        cancelAsyncButton.Visible = false;
+        runAsyncButton.Visible = true;
+        spinnerAsync.Stop();
+        asyncWorkRunning = false;
+    };
+
+    /*
+     * Regular 'OnClicked' event handler for manually invoking the UI thread
+     * in asynchronous code that may not be coming from the UI originally.
+     * gir.core creates a synchronization context that is not being used with this method.
```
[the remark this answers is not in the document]
ousnius: I don't think it does, at least it didn't hit my debugger breakpoint in the task example. Bug?

src/Samples/Gtk-4.0/Async/Helper.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public static class Helper
+    {
+        public static void Invoke(Action? action)
```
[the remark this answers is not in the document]
ousnius: IdleAdd requires a boolean return false which the helper function eliminates. It would add a line everywhere. Do you still want me to remove it?

src/Samples/Gtk-4.0/Async/Program.cs
```diff
@@ -0,0 +1,193 @@
...
+            for (var i = 0; i < tasks.Length; i++)
+            {
+                var taskIndex = i;
+                tasks[taskIndex] = Task.Run(() => Work.DoHeavyWorkWithInvoke(tasksSample.ProgressBar, taskIndex + 1, taskCancelTokenSrc.Token));
+            }
+
+            Task.WaitAll(tasks);
+        });
+
+        await task;
```
badcel: Can we `await Task.WhenAll()` here and remove the outer task (analog to the async sample).

If the inner tasks have Task.Run they should still run independently.

`Task.WaitAll` should be obsolete then.

badcel (commented):
I think I found a little improvement.

src/Samples/Gtk-4.0/Async/Program.cs
```diff
@@ -0,0 +1,188 @@
...
+        tasksSample.RunButton.Visible = false;
+        tasksSample.CancelButton.Visible = true;
+        tasksSample.ProgressBar.Visible = true;
+        tasksSample.Spinner.Start();
+
+        /*
+         * Do some heavy work in a task without using await.
+         * Run multiple tasks at the same time to force multiple threads queueing UI actions.
+         */
+        task = Task.Run(() =>
```
badcel: You can just await this line?

PR #1498, opened 2026-04-10
Gtk.Dropdown sample: Refactor code

badcel:


badcel:
@kashifsoofi Any thoughts on this?

kashifsoofi:
> @kashifsoofi Any thoughts on this?

Apologies @badcel  have been really busy and could not check my emails for some time :(, anything that makes it better and easy to use, thanks for updating this and the original project

badcel:
@kashifsoofi Ah no problem. I just did not know when you would check next time so i merged it to get the topic out of my head. Do you agree that it's an improvement over the previous solution?

kashifsoofi:
Yes definately

PR #1502, opened 2026-04-13
Added JellyTune to apps list

bacmanni:
- [x ] I agree that my contribution may be licensed either under MIT or any version of LGPL license.


badcel:
Hey thanks for using GirCore for your app 🚀.

Can you squash your both commits into one and force push your branch so only one commit is left.

This is for some clean git history.

I have no other remarks. Looks good to merge otherwise 👍

PR #1504, opened 2026-04-13
Added JellyTune to apps.md

bacmanni:

Sorry for the hassle. I have never played with forks before. Decided to reset my fork, and created changes from beginning. Now it should be better?


badcel (approved):
Yep it's perfect, thank you. 🚀

PR #1510, opened 2026-04-23
Implement Support for Composite Template via GResource

theycallhermax:

image

### DIFF
--- a/src/Libs/Gtk-4.0/Public/GResource.cs
+++ b/src/Libs/Gtk-4.0/Public/GResource.cs
@@ -1,0 +1,12 @@
+using Gio;
+
+namespace Gtk;
+
+public class GResource : TemplateLoader
+{
+    public static GLib.Bytes Load(string resourceName)
+    {
+        File file = Gio.Functions.FileNewForUri($"resource://{resourceName}");
+        return file.LoadBytes(null, out _);
+    }
+}
--- a/src/Samples/Gtk-4.0/CompositeTemplate/CompositeTemplate.csproj
+++ b/src/Samples/Gtk-4.0/CompositeTemplate/CompositeTemplate.csproj
@@ -13,9 +13,17 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
+  <Target Name="CompileGlibResourcesForBuild" BeforeTargets="CoreCompile">
+    <Message Importance="high" Text="Compiling GResources..." />
+    <Exec WorkingDirectory="$(MSBuildProjectDirectory)" Command="glib-compile-resources --sourcedir &quot;.&quot; &quot;./CompositeTemplate.gresource.xml&quot; --target=&quot;$(TargetDir)CompositeTemplate.app.gresource&quot;" />
+  </Target>
+  
   <ItemGroup>
     <None Remove="**\*.ui" />
     <EmbeddedResource Include="**\*.ui">
+      <LogicalName>%(Filename)%(Extension)</LogicalName>
+    </EmbeddedResource>
+    <EmbeddedResource Include="$(TargetDir)CompositeTemplate.app.gresource">
       <LogicalName>%(Filename)%(Extension)</LogicalName>
     </EmbeddedResource>
   </ItemGroup>
--- a/src/Samples/Gtk-4.0/CompositeTemplate/Program.cs
+++ b/src/Samples/Gtk-4.0/CompositeTemplate/Program.cs
@@ -1,4 +1,8 @@
+using System.Reflection;
 using CompositeTemplate;
+
+Gtk.Module.Initialize();
+GirCore.Integration.Initialize();
 
 var application = Gtk.Application.New("org.gir.core", Gio.ApplicationFlags.FlagsNone);
 application.OnActivate += (sender, args) =>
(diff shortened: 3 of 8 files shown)


src/Samples/Gtk-4.0/GResourceCompositeTemplate/GResourceCompositeTemplate.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <ItemGroup>
```
badcel: I wonder if the project should be combined with the existing CompositeTemplate sample project? The sample Window could contain two composite widgets: One resulting from an assembly ressource and one from a GRessource.

This would keep CompositeTemplate related samples better together.

src/Samples/Gtk-4.0/GResourceCompositeTemplate/GResourceCompositeTemplate.app.gresource
badcel, line 1: Please don't include binary files.

src/Libs/Gtk-4.0/Public/GResource.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using GLib.Internal;
+
+namespace Gtk;
+
+public class GResource : TemplateLoader
+{
+    public static GLib.Bytes Load(string resourceName)
+    {
+        var resourceUri = NonNullableUtf8StringUnownedHandle.Create($"resource://{resourceName}");
```
badcel: The code looks correct but uses the internal APIs. The APIs you used are all available via the public API which makes things a lot easier. It's just not easy to finde because GIO has global helper methods (like most GObject based libraries) which are availablie in GirCore via `Gio.Functions`:

```csharp
        var file = Gio.Functions.FileNewForUri($"resource://{resourceName}");
        return file.LoadBytes(null, out _);
```

src/Samples/Gtk-4.0/GResourceCompositeTemplate/GResourceCompositeTemplate.gresource.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+  To rebuild GResources:
+  
+  glib-compile-resources \
```
badcel: This command should probably be executed dynamically during build to generate the binary automatically.

badcel (changes requested):
Hey, thanks for your contribution and the pretty good idea :-)

There are some comments to fix / clarify before it can be merged.

src/Samples/Gtk-4.0/GResourceCompositeTemplate/GResourceCompositeTemplate.csproj
theycallhermax: I really like this idea, it really shows how extensible the interface could be

image

theycallhermax:
not really proud of how I just initialized and discarded the `CompositeLabelWidget` just to register it's object type, theres probably a better way...

```cs
    _ = CompositeLabelWidget.NewWithProperties([]);
```

src/Samples/Gtk-4.0/CompositeTemplate/Program.cs
```diff
@@ -1,12 +1,26 @@
...
+
+using var stream = Assembly.GetExecutingAssembly()
+    .GetManifestResourceStream("CompositeTemplate.app.gresource");
+
+var buffer = new byte[stream!.Length];
+stream.ReadExactly(buffer);
+
+using var bytes = GLib.Bytes.New(buffer);
+using var resource = Gio.Resource.NewFromData(bytes);
+resource.Register();
```
badcel, line 25: I'm no expert in GTK/Gio so I wonder why we need to register the resource?

In your first sample it worked without registering?

badcel:
> not really proud of how I just initialized and discarded the `CompositeLabelWidget` just to register it's object type, theres probably a better way...
> 
> 
> 
> ```cs
> 
>     _ = CompositeLabelWidget.NewWithProperties([]);
> 
> ```

Yes there is a better way. The problem is that the type is not known to GObject and can't be instantiated. By creating an instance the type is registered implicitly.

There is a generated method which registers all types: https://github.com/badcel/Maus/blob/main/src/Maus/Program.cs#L5

But for it to work you need to initialize the library your subclasses depend on. In your case this is Gtk not adwaita like in my link.

src/Samples/Gtk-4.0/CompositeTemplate/Program.cs
theycallhermax: it actually was registered in my first sample, registering the resource hooks its data into the process so that we can lookup it's files later in the code. if we didn't, then the gresource template loader would just fail since theres nothing provided to load files from

src/Samples/Gtk-4.0/CompositeTemplate/CompositeTemplate.csproj
```diff
@@ -13,11 +13,19 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
+  <Target Name="CompileGlibResourcesForBuild" BeforeTargets="CoreCompile">
+    <Message Importance="high" Text="Compiling GResources..." />
+    <Exec WorkingDirectory="$(MSBuildProjectDirectory)" Command="glib-compile-resources --sourcedir &quot;.&quot; &quot;./CompositeTemplate.gresource.xml&quot; --target=&quot;$(TargetDir)/CompositeTemplate.app.gresource&quot;" />
```
badcel:
```suggestion
    <Exec WorkingDirectory="$(MSBuildProjectDirectory)" Command="glib-compile-resources --sourcedir &quot;.&quot; &quot;./CompositeTemplate.gresource.xml&quot; --target=&quot;$(TargetDir)CompositeTemplate.app.gresource&quot;" />
```

`$(TargetDir)` already contains a `/` for me. The [docs](https://learn.microsoft.com/en-us/cpp/build/reference/common-macros-for-build-commands-and-properties?view=msvc-170#list-of-common-macros) hint into the same direction. (I found no concrete docs for C# though.)

badcel (commented):
This looks almost good to go :rocket:. Please see my little comment. If it is the same for you I think the slash can be removed.

Please amend your changes again to keep this as a single commit.

I opened #1511 to clarify where to put the template loaders. Your opinion over there is very welcome. But in case we change this, this could be an additional PR.

Thanks for your work already :-)

badcel (approved):

PR #1516, opened 2026-05-02
Use `global::Gtk` instead of `Gtk` in template codegen to avoid namespace conflicts

Daniel15:
My app's namespace is `WebCamControl.Gtk`, which conflicts with the `Gtk` namespace references in the generated templates:
```
0>SavePresetDialog.Template.g.cs(10,35): Error CS0234 : The type or namespace name 'Internal' does not exist in the namespace 'WebCamControl.Gtk' (are you missing an assembly reference?)
0>SavePresetDialog.Template.g.cs(11,9): Error CS0234 : The type or namespace name 'Internal' does not exist in the namespace 'WebCamControl.Gtk' (are you missing an assembly reference?)
0>SavePresetDialog.Template.g.cs(16,9): Error CS0234 : The type or namespace name 'Internal' does not exist in the namespace 'WebCamControl.Gtk' (are you missing an assembly reference?)
0>SavePresetDialog.Template.g.cs(22,1): Error CS0234 : The type or namespace name 'Internal' does not exist in the namespace 'WebCamControl.Gtk' (are you missing an assembly reference?)
0>SavePresetDialog.Template.g.cs(34,9): Error CS0234 : The type or namespace name 'Internal' does not exist in the namespace 'WebCamControl.Gtk' (are you missing an assembly reference?)
0>PanAndTiltButtons.Template.g.cs(10,35): Error CS0234 : The type or namespace name 'Internal' does not exist in the namespace 'WebCamControl.Gtk' (are you missing an assembly reference?)
0>PanAndTiltButtons.Template.g.cs(11,9): Error CS0234 : The type or namespace name 'Internal' does not exist in the namespace 'WebCamControl.Gtk' (are you missing an assembly reference?)
...
```

This diff updates the codegen to use `global::Gtk` instead of `Gtk` to avoid the conflict.

# Test plan

I couldn't find any unit tests for the codegen, however I did manually verify that the `CompositeTemplate` sample still compiles and runs properly if I change the namespace to `CompositeTemplate.Gtk`:

image

### DIFF
--- a/src/Integrations/GObject-2.0.Integration/SourceGenerator/Subclass/SubclassCode.cs
+++ b/src/Integrations/GObject-2.0.Integration/SourceGenerator/Subclass/SubclassCode.cs
@@ -58,16 +58,16 @@ internal static class SubclassCode
             : $"\"{subclassData.QualifiedName}\"";
 
         return $$"""
-                  {{subclassData.TypeData.Properties.Accessibility}} unsafe partial class {{subclassData.TypeData.Properties.NameGenericArguments}} : {{subclassData.Parent}}, GObject.GTypeProvider
+                  {{subclassData.TypeData.Properties.Accessibility}} unsafe partial class {{subclassData.TypeData.Properties.NameGenericArguments}} : {{subclassData.Parent}}, global::GObject.GTypeProvider
                   {
                        {{GeneratedCodeAttribute.Render()}}
-                       private static readonly GObject.Type GType = GObject.Internal.SubclassRegistrar.Register<{{subclassData.TypeData.Properties.NameGenericArguments}}, {{subclassData.Parent}}>(&ClassInit, &InstanceInit, {{qualifiedName}});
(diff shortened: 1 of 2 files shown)


badcel:
Would you mind doing the same (for the GObject namespace) for GObject integration either here or in a separate PR?

Daniel15:
Sure, I'll add it here.

badcel:
Sorry I have another thought: you did the change because you have a compilation error right? 

So we should probably make sure this does not happen again with some unit test data: Could you add two new files in the same folder like this file: https://github.com/gircore/gir.core/blob/main/src/Tests/Data/GObject-2.0.Integration.Tests.Data/SomeSubClass.cs#L3

Those files could be in a GTK / GObject sub namespace to replicate the error you faced. This should ensure that the problem does not come up again.

Daniel15:
Good idea. I can do that but might not be able to get to it until later in the week.

badcel:
There is only a data project for GObject available and not GTK.

I will make a follow up issue for this PR. And we can do it separately and merge this instead.

PR #1545, opened 2026-06-30
Generator hintName collision

### LINKED ISSUE #1540: Generator hintName collision
## Summary

In GirCore 0.8.0, the GObject subclass source generator fails when two or more `[GObject.Subclass<T>]` partial classes have the same class name but are declared in different namespaces.

The generator creates the generated source `hintName` from the short type name only. Because the namespace is not included, different valid C# types can produce the same generated source name.

## Version

GirCore 0.8.0

## Error message

```text
Generator 'Generator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'ArgumentException' with message 'The hintName 'View.Subclass.g.cs' of the added source file must be unique within a generator. (Parameter 'hintName')'.
```

## Minimal example

```csharp
namespace GirCoreSample.Axes.ColorsAndPosition;

[GObject.Subclass<Gtk.Box>]
public partial class View
{
}
```

```csharp
namespace GirCoreSample.Bars.DelayedAnimation;

[GObject.Subclass<Gtk.Box>]
public partial class View
{
}
```

Both classes are valid C# types, because they are declared in different namespaces.

The generator currently produces the same generated source name for both classes:

```text
View.Subclass.g.cs
```

This causes an `AddSource` collision, because the `hintName` value must be unique within a generator.

## Cause

The GObject subclass generator uses `TypeData.Filename` when creating the generated source name:

```csharp
hintName: $"{subclassData.TypeData.Filename}.Subclass.g.cs"
```

The `TypeData.Filename` value is produced by `TypeDataHelper.GetFileName(...)`.

The current filename generation is based on the short type name. This is not enough when two different namespaces contain the same class name.

## Fix

`TypeDataHelper.GetFileName(...)` should generate namespace-aware filenames.

The namespace should become part of the generated `Filename` value, while the existing nested type name handling and generic parameter count handling should remain part of the filename.

This makes generated source names unique for types such as:

```text
GirCoreSample.Axes.ColorsAndPosition.View
GirCoreSample.Bars.DelayedAnimation.View
```

Instead of both producing:

```text
View.Subclass.g.cs
```

they should produce different `hintName` values.

## Affected generator paths

This affects every source generator path that uses `TypeData.Filename` as part of its generated source `hintName`.

The two relevant paths are:

```text
src/Integrations/GObject-2.0.Integration/SourceGenerator/Subclass/SubclassCode.cs
src/Integrations/Gtk-4.0.Integration/SourceGenerator/Template/TemplateCode.cs
```

## Tests

The GObject integration tests should include two `Widget` classes with the same short name declared in different namespaces.

This verifies that the GObject subclass generator no longer fails with `CS8785` when identical short type names appear under different namespaces.

The GTK integration tests should include the same situation for the GTK template generator.

This verifies that the GTK template generator also receives unique generated source `hintName` values in this case.

## Expected result

The same class name in different namespaces should receive different generated source `hintName` values.

There should be no `AddSource` collision when multiple valid C# types share the same short class name.

czirok:

Include the namespace in generated source filenames to avoid hintName collisions when different namespaces contain types with the same short name.

The filename generation in TypeDataHelper is now namespace-aware while keeping nested type names and generic parameter counts as part of the generated name.

This prevents AddSource collisions in source generators that use TypeData.Filename, including the GObject subclass generator and the GTK template generator.

Tests were added for both GObject and GTK integration paths with identical Widget type names declared in different namespaces, verifying that source generation no longer fails with CS8785.

### DIFF
--- a/src/Tests/Libs/GObject-2.0.Integration.Tests/DiagnosticAnalyzerTest.cs
+++ b/src/Tests/Libs/GObject-2.0.Integration.Tests/DiagnosticAnalyzerTest.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -65,6 +66,24 @@ public class DiagnosticAnalyzerTest : Test
                                                public class NotRaiseGirCore1004<T> : NotRaiseGirCore1004;
                                                """;
 
+    private const string DuplicateSubclassNameOne = """
+                                                    using GObject;
+
+                                                    namespace DuplicateSubclassName.One;
+
+                                                    [Subclass<GObject.Object>]
+                                                    public partial class Widget;
+                                                    """;
+
+    private const string DuplicateSubclassNameTwo = """
+                                                    using GObject;
+
+                                                    namespace DuplicateSubclassName.Two;
+
+                                                    [Subclass<GObject.Object>]
+                                                    public partial class Widget;
+                                                    """;
+
     [TestMethod]
     [DataRow(RaiseGirCore1002, "GirCore1002", true)]
     [DataRow(RaiseGirCore1004, "GirCore1004", true)]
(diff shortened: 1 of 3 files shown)


src/Tests/Libs/GObject-2.0.Integration.Tests/DiagnosticAnalyzerTest.cs
```diff
@@ -100,4 +118,28 @@ public async Task ShouldRaiseExpectedDiagnosticIds(string code, string diagnosti
         else
             diagnostics.ContainsNoDiagnostic(diagnosticId);
     }
+
+    [TestMethod]
+    public void ShouldNotFailWhenSubclassNamesRepeatInDifferentNamespaces()
```
badcel: If I disable your fix in the `TypeDataHelper" this test is not failing. Is this a problem on my end or is something missing here?

I don't know if this is easily testable. It would be easy to update `GObject-2.0.Integration.Tests.Data.NamespaceTest.cs` like this:

```cs
namespace MyNamespace.GObject
{
    [global::GObject.Subclass<global::GObject.Object>]
    public partial class NamespaceSubClass;
}

namespace MyNamespace.Intern
{
    [global::GObject.Subclass<global::GObject.Object>]
    public partial class NamespaceSubClass;
}
```

The drawback is that there is no clean error message that pinpoints the problem. But at least it works and needs no contrapositive testing.

In case you manage to make this testable via a unit test I think we need to have contrapositive testing. So you confirm both cases: 
1. Confirm that the error is not raised
2. Confirm that the error can be raised. This would require some configuration inside the source generator that allows to disable the namespace prefixing. As the generator is instanciated in the  unit test itself this should be doable.

Without contrapositive testing we are only seeing that something is not happening. But this could have endless reasons, like something is broken in code or perhaps like in this case: The test is simply not working as intended and nobody recognizes that there is a problem with the test.

src/Tests/Libs/Gtk-4.0.Integration.Tests/DiagnosticAnalyzerTest.cs
```diff
@@ -84,4 +100,29 @@ public async Task ShouldRaiseExpectedDiagnosticIds(string code, string diagnosti
         else
             diagnostics.ContainsNoDiagnostic(diagnosticId);
     }
+
+    [TestMethod]
+    public void ShouldNotFailWhenTemplateNamesRepeatInDifferentNamespaces()
```
badcel: This test probably has the same issue like the one in `GObject-2.0.Integration.Test`.

badcel (commented):
The code itself looks perfectly fine for me.

Unfortunately the tests seem not to work like intended. I'm fine with having a "stupid" test like mentioned in the comments to detect if "something" is broken. I this happens to often we can still think about moving tests for namespace verification in actual unit tests.

czirok:
I reworked the test so it verifies the actual generator output.

It now creates two subclasses with the same class name in different namespaces, runs the generator, checks that no `CS8785` diagnostic is produced, verifies that both namespace-qualified hint names are generated, and ensures that all generated hint names are unique.

badcel (approved):
Great work! Thank you. Let's merge :-)

Issue #1506, opened 2026-04-16
Consider aligning cairo API with the rest of GirCore


badcel:
Consider aligning cairo API with the rest of GirCore.

Consider especially the implications that constructors are available which means inheritance of instances is possible.

PR #1553, opened 2026-07-04
Expose routines for reading and creating .png files

### LINKED ISSUE #1229: support cairo_image_surface_create_from_png()
The function cairo_image_surface_create_from_png() does not seem to be exposed in Gir.Core:

   https://www.cairographics.org/manual/cairo-PNG-Support.html

This is a useful function for loading a PNG image directly into a Cairo image surface, so it would be nice to have it available.

zur003:

This pull request exposes the `cairo_image_surface_create_from_png` and `cairo_surface_write_to_png` routines and thereby resolves #1229.


src/Libs/cairo-1.0/Internal/ImageSurface.cs
```diff
@@ -25,4 +25,7 @@ public class ImageSurface
 
     [DllImport(CairoImportResolver.Library, EntryPoint = "cairo_image_surface_get_width")]
     public static extern int GetWidth(SurfaceHandle handle);
+
+    [DllImport(CairoImportResolver.Library, EntryPoint = "cairo_image_surface_create_from_png")]
+    public static extern SurfaceOwnedHandle CreateFromPng(GLib.Internal.NonNullableUtf8StringOwnedHandle filename);
```
badcel: The methods are alphabetically ordered. Would you mind moving the new code after `CreateForData`?

src/Libs/cairo-1.0/Internal/Surface.cs
```diff
@@ -64,4 +64,9 @@ public partial class Surface
 
     [DllImport(CairoImportResolver.Library, EntryPoint = "cairo_surface_status")]
     public static extern Status Status(SurfaceHandle handle);
+
+    [DllImport(CairoImportResolver.Library, EntryPoint = "cairo_surface_write_to_png")]
+    public static extern Status WriteToPng(SurfaceHandle handle, GLib.Internal.NonNullableUtf8StringOwnedHandle filename);
+
```
badcel: Please remove empty lines at the end.

src/Libs/cairo-1.0/Public/ImageSurface.cs
```diff
@@ -45,5 +45,11 @@ public Span<byte> GetData()
             return new Span<byte>(data.ToPointer(), GetSizeInBytes());
         }
     }
+
+    public static Surface CreateFromPng(string filename)
```
badcel: Wouldn't it make sense to return an `ImageSurface` here?

See the docs: https://www.cairographics.org/manual/cairo-PNG-Support.html#cairo-image-surface-create-from-png

src/Libs/cairo-1.0/Internal/Surface.cs
```diff
@@ -64,4 +64,9 @@ public partial class Surface
 
     [DllImport(CairoImportResolver.Library, EntryPoint = "cairo_surface_status")]
     public static extern Status Status(SurfaceHandle handle);
+
+    [DllImport(CairoImportResolver.Library, EntryPoint = "cairo_surface_write_to_png")]
```
badcel, line 68: I think the code belongs into `Internal.ImageSurface`?

The docs say that the surface must contain pixel data: https://www.cairographics.org/manual/cairo-PNG-Support.html#cairo-surface-write-to-png

I would assume that means an `ImageSurface`?

src/Libs/cairo-1.0/Public/Surface.cs
```diff
@@ -86,4 +86,10 @@ public void MarkDirty(int x, int y, int width, int height)
             ? null
             : new Device(deviceHandle.OwnedCopy());
     }
+
+    public Status WriteToPng(string filename)
```
badcel, line 90: Move accordingly to the internal code.

badcel (commented):
Thanks for the PR 🚀 

Please see my comments.

If you do changes please squash all your commits into one afterwards, so that we can merge your code with one commit only.

src/Libs/cairo-1.0/Internal/ImageSurface.cs
zur003: Easily done.

src/Libs/cairo-1.0/Internal/Surface.cs
zur003: Also easily done

src/Libs/cairo-1.0/Public/ImageSurface.cs
zur003: I think you're right. To generate an ImageSurface rather than a Surface, I've replaced the static method with a constructor, paralleling the logic used for `cairo_image_surface_create`.

src/Libs/cairo-1.0/Internal/Surface.cs
zur003: I'm not sure I fully agree. There are a few other Surface types in Cairo, such as QuartzSurface, with pixel data, and potentially a user might want to generate a PNG file from one of these. However, in the gir-core context, probably only ImageSurface is likely to ever be used. 
My initial placement was based on the cairo naming convention (that is, it's `cairo_surface_write_to_png`, not `cairo_image_surface_write_to_png`, without giving it much thought. If you would definitely prefer to have it in ImageSurface rather than Surface I'd be happy to make that change.

src/Libs/cairo-1.0/Public/Surface.cs
zur003: See comment above.

zur003:
I've partially made the changes you suggested, but haven't moved WriteToPng. Provide me with your feedback, and I'll squash my changes into a new PR.

src/Libs/cairo-1.0/Public/ImageSurface.cs
```diff
@@ -19,6 +19,12 @@ public ImageSurface(GLib.Bytes data, Format format, int width, int height, int s
         SetUserData(data);
     }
 
+    public ImageSurface(string filename) :
```
badcel: I'm not sure I like the constructor here as it does not make clear that it is valid for png only.

We should at least rename the parameter to `filenamePng` or similar. But then there would be no way to create other image surfaces from a file name.

On the other hand I like that it's aligned with the rest of the API.

There is still issue: https://github.com/gircore/gir.core/issues/1506

So to come to a conclusion: Let's start with a regular constructor in this PR and other considerations come later.

What do you think?

src/Libs/cairo-1.0/Internal/Surface.cs
badcel: Ah I did not know that. Then I agree with you. Let's keep it like it is.

badcel (commented):
Thanks for the quick updates.

Only one comment remains. I will need to test your code manually. Which will happen probably tomorrow.

Afterwards I could merge it.

If you squash your commits you don't need to create a new PR. You just make 1 commit from multiple and force push your branch. The PR will update itself. If you prefer it the other way, it's fine for me, too.

Thanks for your contribution already 👍

src/Libs/cairo-1.0/Public/ImageSurface.cs
zur003: I went for `pngFilename` as the parameter name. I agree that not being able to create image surface from other types of files is a problem. I suppose it would be possible to make the constructor private and use a static `NewFromPng` method to create the ImageSurface, but this might be done as part of dealing with #1506.

zur003:
I've squashed my changes. Let me know if anything further is needed.

badcel:
You fixed the filename but did not update its usages thus the build is failing.

I don't know if it is needed but if only the Linux build should fail next time you need to run `dotnet format GirCore.slnx` to automatically fix all formatting errors. Please squash afterwards again.

You can run the command just for security then we probably save an additional round of fixes.

zur003:
That was an embarrassing mistake on my part. Should be fixed now.

badcel:
I don't know what happened but something got mixed up.

Can you evaluate?

zur003:
I think I made some mistakes with git rebase and squash. It may be tidier if I close this pull request and start afresh with a new one.

PR #1328, opened 2025-09-24
Generate methods with untyped records as container-transfer return values

ousnius:
`transfer-container` is usually used with `GLib.List` and `GLib.SList` as return values.
The ownership of the returned container is transferred, but the ownership of the elements in the container is not.

API examples:
* `gdk_pixbuf_get_formats`
* `gtk_textiter_get_tags`

Added a test that returns a container with two static elements that are owned by native code.

- [x] I agree that my contribution may be licensed either under MIT or any version of LGPL license.

PR #1572, opened 2026-07-26
Generator: Use UntypedRecord model in untyped record renderers

ousnius:
The untyped record renderers and converters used the TypedRecord model to look up handle, data and class names. The generated output is identical as both models currently produce the same names, but the naming could diverge one day.

All generated files were checked before and after the change - they remain identical.

Merge #1328 afterwards.


badcel:
Thank you 👍

PR #1573, opened 2026-07-26
Add GLib List/SList return value copy/free support

ousnius:
Implement end-to-end handling for GLib.List and GLib.SList return values by loading nested G-IR element types, resolving them into GirModel metadata, and generating managed array conversions for supported element types (strings, classes/interfaces, and typed records) with transfer-aware ownership behavior.

Add internal/container list handle types and UTF-8 string helpers for safe conversion/freeing, wire new converters into return-type rendering, and keep unsupported list element cases on container fallback paths. Also add native GirTest list fixtures, binding/unit tests for element-type resolution and transfer semantics, and update the Secret sample to iterate returned collections directly.

⚠️ Breaking API change
Methods returning a GList/GSList now hand back a typed managed array whose elements are adopted or copied according to the transfer mode, instead of an opaque GLib.List wrapper that leaked its contents. 162 methods across the bound libraries changed shape — for example Gio.AppInfoHelper.GetAll() went from returning GLib.List to Gio.AppInfo[].

Once the elements are copied out into managed objects, the only thing left to free is the container itself, so transfer full and transfer container share one internal handle type (ListContainerHandle), releasing with `g_list_free`/`g_slist_free`); transfer none keeps the existing unowned handle that frees nothing. What differs per transfer mode is the element, decided in ListElement.cs: with transfer full we own each element, so it is adopted by an owned handle or an owning wrapper; with container or none we don't, so each element is copied or referenced.

Supported element types are GObject classes and interfaces, UTF-8 strings, and typed and opaque-typed records; anything else (a gpointer list, for instance) still falls back to the old GLib.List wrapper rather than losing the method.


src/Generation/GirLoader/Output/TypeReference.cs
```diff
@@ -34,4 +36,13 @@ public override string ToString()
 
     internal bool GetIsResolved()
         => Type is { };
+
+    internal IReadOnlyList<GirModel.AnyType> GetResolvedElementTypes() => this switch
```
badcel, line 40: I think this method does not belong here. It adds logic for a class that is more concrete than this one. The data it outputs is only available in specific subclass. This makes it hard to add new subclasses to "TypeReference" as one would always need to check if this methods need to be changed in the "superclass".

Thus it probably violates the "Liskov substitution principle".

I would probably move this method one hierarchy level up onto `ResolveableTypeReference`.

src/Generation/GirModel/AnyType.cs
```diff
@@ -34,6 +34,15 @@ public static bool IsGLibPtrArray(this GirModel.AnyType anyType)
         => anyType.TryPickT1(out var arrayType, out _)
            && arrayType is GirModel.PointerArrayType;
 
+    /// <summary>
+    /// Checks whether the given type is a GLib.List or a GLib.SList. Both are
+    /// container types which store their element type as a nested type.
+    /// </summary>
+    public static bool IsGLibList(this GirModel.AnyType anyType)
```
badcel, line 41: I think we should try to avoid hard coding concrete libraries on the level of the generator. The generator is a representation of the [rnc](https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/docs/gir-1.2.rnc?ref_type=heads) file which does not know of concrete libraries either.

So let's remove this method for now. We can see how to solve it when we reach a point where it is needed.

src/Generation/GirLoader/Output/ReturnValue.GirModel.cs
```diff
@@ -8,6 +10,8 @@ public partial class ReturnValue : GirModel.ReturnType
         _ => GirModel.AnyType.From(TypeReference.GetResolvedType())
     };
 
+    IReadOnlyList<GirModel.AnyType> GirModel.ReturnType.ElementTypes => TypeReference.GetResolvedElementTypes();
```
badcel, line 13: This should not be here. Please see notes in the GirModel.

src/Generation/GirLoader/Output/SingleParameter.GirModel.cs
```diff
@@ -26,6 +29,11 @@ public partial class SingleParameter : GirModel.Parameter
         },
         varargs => varargs
     );
+
+    IReadOnlyList<GirModel.AnyType> GirModel.Parameter.ElementTypes => TypeReferenceOrVarArgs.Match(
```
badcel, line 33: This should not be here. Please see notes in the GirModel.

src/Generation/GirLoader/Output/TypeReference.cs
```diff
@@ -34,4 +36,13 @@ public override string ToString()
...
+    {
+        ResolveableTypeReference resolveableTypeReference => resolveableTypeReference.ElementTypeReferences
+            .Where(x => x.GetIsResolved())
```
badcel, line 43: Why do you filter on resolved types only?

If a `TypeReference` has element types which can't be resolved there is a bug in the generator or the GIR file.

With the filter the element types could be incomplete which would generate a broken API. If an exception is thrown in this case the API is not generated which I prefer over having a broken API.

For me it looks like the method `GetIsResolved` can be removed as there are no other usages except in your new code?

src/Generation/GirModel/ReturnType.cs
```diff
@@ -1,8 +1,17 @@
...
 public interface ReturnType : Nullable
 {
     AnyType AnyType { get; }
+
+    /// <summary>
+    /// The resolved element types if <see cref="AnyType"/> is a container type
+    /// like GLib.List. GLib.HashTable has two element types: the key and the
+    /// value type. Empty if there are no (resolved) element types.
+    /// </summary>
+    IReadOnlyList<AnyType> ElementTypes { get; }
```
badcel, line 14: I think this does not belong here. Not return types have ElementTypes, but Types can have element types. This would match the [RNC](https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/docs/gir-1.2.rnc?ref_type=heads) file again.

src/Generation/GirModel/Parameter.cs
```diff
@@ -13,5 +15,12 @@ public interface Parameter : Nullable
     bool IsConst { get; }
     bool IsVolatile { get; }
     OneOf.OneOf<AnyType, VarArgs> AnyTypeOrVarArgs { get; }
+
+    /// <summary>
+    /// The resolved element types if this parameter is a container type
+    /// like GLib.List. GLib.HashTable has two element types: the key and the
+    /// value type. Empty if there are no (resolved) element types.
+    /// </summary>
+    IReadOnlyList<AnyType> ElementTypes { get; }
```
badcel, line 24: I think this does not belong here. Not return types have ElementTypes, but Types can have element types. This would match the [RNC](https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/docs/gir-1.2.rnc?ref_type=heads) file again.

In general the `GirModel` project defines an API which should be independent of the XML format of the GIR-Files and should be easy to use.

I think we have several possibilities where to put the the list of `ElementTypes` inside `GirModel`:
* `GirModel.Type`: This would match the RNC but might be too low level. As theoretically every renderer of any type would suddenly be able to access the element types but they don't make sense everywhere. So I would say this adds to much obsolete brain load.
* `GirModel.Record`: As every code renderer needs to explicitly support element types I think it makes sense to add them in the GirModel where we can expect them. This would be on records for the time beeing. If we encounter them on other type kinds we can add them where needed. This matches the renderer logic, as those are split by types (like class / record / etc), too.

This would ensure everywhere where a record comes up, we have the element types available. E.g. if we want to render generic C# code of a record the information is just there because `ElementTypes.Length > 0`.

badcel (changes requested):
Thank you for your PR!

You already did a lot of stuff. I started reviewing only the foundations as I think those require modifications.

To minimise the effort on your side I would suggest that we stop this PR and create several new PRs one after the other with only one (or at max. 2) are already coded. In this way there is not much code relying on code of lower levels which might need to change due to PR review.

I would suggest that we work our way up all the layers until we come to some code rendering. Every layer is one PR. If we need to refactor lower layers this would be fine for me.

The great thing is that you already know where the pitfalls are so I think this should be doable.

 We can keep this PR as a reference which does not change and you can copy ideas from it to the final PRs.

The review I did only covers the folders `GirModel` and `GirLoader`. which is the lowest level for me. So the next PR should only contain those changes. In this way we prepare the low level data to be consumed by higher levels.

If this is  done we can discuss what to do next.

P.S.: If you instead would like to create a new PR for the first layer and rebase this PR on the changes and get a new review for this PR, which perhaps results in a separate PR again with another rebase as follow up, this would be fine for me, too. You can decide how it is working best for you.

src/Generation/GirModel/Parameter.cs
badcel: Perhaps I'm wrong here. Type references can have element types. And there are no type references in the GirModel anymore.

Is this the reason you put the element types on the return value and parameter?

badcel: I think I got confused because I saw that the ElementType can be available on `Fields`, too.

With the current approach we would need to add it fields, too.

We could either create an Interface for it with only the `ElementTypes` on it and let the other interfaces inherit it or we somehow wrap AnyType and create something that can represent a type with Elements and use it on fields, parameters and Return types.

PR #1596, opened 2026-08-26
Improve Exception integration

### LINKED ISSUE #873: Improve Exception integration
Currently GLib has `GException` which is thrown if an error occurs in C: 

https://github.com/gircore/gir.core/blob/712ab1740ca138a9a6dcd36975e2302b0a58f8d0/src/Libs/GLib-2.0/Public/Exception.cs#L6

GLib errors contain a domain and error code in addition to its message.

Derive from `GException` and create `GException<T>` with a property `Code` of type `T`.

The libraries define their error domains as enums which map to error domains. If an error is received we can get the domain code, map it to the corresponding enum and convert the error code into an enum value.

A user catching `GException<ConvertError>` can get the specific error from the `Code` property. 

Compare the `ConvertError` domain: https://docs.gtk.org/glib/error.ConvertError.html

See generated `ConvertError` enum: https://gircore.github.io/api/GLib.ConvertError.html 

Using generics could potentially be hard if the domain is unknown during compile time.

badcel:

Fixes: #873 

Add tostring for exception type


src/Generation/Generator/Renderer/Public/EnumerationErrorDomain.cs
```diff
@@ -0,0 +1,44 @@
...
+using System;
+using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
+
+#nullable enable
+
+namespace {Namespace.GetPublicName(enumeration.Namespace)};
+
+// AUTOGENERATED FILE - DO NOT MODIFY
+
```
badcel: Summary with reference to the error domain name.

src/Libs/GLib-2.0/Public/ErrorType.cs
```diff
@@ -0,0 +1,9 @@
+namespace GLib;
+
+public readonly ref struct ErrorKind
```
badcel: Add Summary and debugger display which contains the name of the error type which is calculated from the error domain and some string magic taking into account symbols prefixes like "g_":

https://docs.gtk.org/glib/error-reporting.html#error-domains

src/Generation/Generator/Renderer/Public/EnumerationErrorDomain.cs
```diff
@@ -0,0 +1,44 @@
...
+{PlatformSupportAttribute.Render(enumeration as GirModel.PlatformDependent)}
+public readonly ref struct {enumeration.Name}
+{{
+    private static readonly GLib.Quark Quark = GLib.Functions.QuarkFromString(""{enumeration.ErrorDomain}"");
+
+    {enumeration
```
badcel: Add an "Any" Member and allow to match it with only the domain in case the error code is not relevant.

src/Libs/GLib-2.0/Public/GException.cs
```diff
@@ -1,20 +1,47 @@
...
+/// </summary>
+[DebuggerDisplay("GException: ErrorDomain={Type.GetErrorDomain()}, ErrorCode={Type.Code}")]
+public sealed class GException(Internal.ErrorHandle errorHandle) 
+    : Exception(Marshal.PtrToStringUTF8(errorHandle.GetMessage())), IDisposable
 {
-    private readonly Internal.ErrorHandle _errorHandle;
+    /// <summary>
+    /// The <see cref="ErrorCode"/> of this exception.
+    /// </summary>
+    private ErrorCode Type => new()
```
badcel: Public? Wird in Doku genannt.

src/Libs/GLib-2.0/Public/GException.cs
```diff
@@ -1,20 +1,47 @@
 using System;
+using System.Diagnostics;
 using System.Runtime.InteropServices;
 
 namespace GLib;
 
-public sealed class GException : Exception, IDisposable
+/// <summary>
+/// A GException wraps a GLib.Error and allows to catch it based on a given exception type.
```
badcel: ExceptionType is invalid: ErrorCode

PR #1603, opened 2026-09-15
Fix Cdecl calling conventions for native callbacks on Windows x86

Evgeniy347:
GLib and GObject invoke callbacks using the C ABI. Leaving the unmanaged calling convention implicit selects the wrong convention on Windows x86, which can corrupt the stack during native-to-managed callbacks, including the toggle notifications used when constructing objects.

Specify Cdecl on generated internal callback delegates and unmanaged subclass entry points, and on the matching toggle notification, subclass registration, and GObject vtable function pointer types.

Add 12 regression cases to the existing GLib, GObject, and GObject integration test projects. They inspect the compiled ABI metadata for both delegate renderers, the vtable, registration callbacks, and concrete/abstract generated subclasses. Native tests also exercise repeated main-context callbacks and object creation/disposal. The metadata checks detect the regression on x64 as well as x86.

Validation:

- Before the fix, all 10 ABI metadata cases failed on current main (`ccfc283f`); after the fix, they pass.
- Windows x64: all tests in the three affected projects pass on .NET 8, 9, and 10 (291 passed, 12 existing skips across the three frameworks).
- Windows x86, .NET 10: 21 selected tests pass, including all new cases and existing concrete/abstract subclass integration tests, using the x86 GLib/GObject runtime.
- Regenerated all library bindings and built Gtk-4.0 and its dependencies in Release for .NET 8, 9, and 10 (zero errors; warnings remain in generated code).
- Formatting checks for the changed files and `git diff --check` pass.


badcel:
Hi thanks for the PR!

Do you know how I can confirm that the libraries use `CDECEL` when build for windows x86?

Do you use GirCore on x86 windows?

Evgeniy347:
Hi, thanks for looking at this!

Yes — we're migrating a desktop app that has to run as a 32-bit process, so this came out of real x86 usage (an x86 process on x64 Windows; a 32-bit Windows install is untested).

On confirming the convention: instead of inferring it, you can measure the shipped DLLs. On x86 a cdecl function leaves its arguments for the caller and ends with a plain `ret`, while a stdcall function ends with `ret <imm16>`. Counting both forms across `objdump -d` (MSYS2 i686 GTK 4.18.2-1, GLib 2.84.0-1):

| DLL | `ret` | `ret $imm` | stdcall |
| --- | --- | --- | --- |
| libgobject-2.0-0.dll | 1295 | 10 | 0.8% |
| libglib-2.0-0.dll | 3422 | 15 | 0.4% |
| libgtk-4-1.dll | 22623 | 117 | 0.5% |
| kernel32.dll (control) | 871 | 1210 | 58.1% |
| user32.dll (control) | 3032 | 3105 | 50.6% |

The two Windows system DLLs are the stdcall control. All ten `ret $0xc` sites in libgobject are WINAPI-shaped callbacks in the CRT/entry region, not the GLib API surface. This measures the delivered binaries, so it doesn't depend on recovering the original compiler flags.

Since .NET defaults unmanaged callbacks to stdcall on Windows x86, while x64/ARM64 have a single calling convention, the fix changes behaviour on x86 only. I re-ran everything today on Windows 11 / .NET 10: unpatched 0.8.1 aborts with `stack smashing detected` (`0xC0000409`) while creating a `Gtk.Box`; with the patch both scenarios pass. The ABI tests added in this PR also fail on current main and pass with the fix on x64, so your normal CI covers the regression.

Reproduction, logs and the measurement script: https://github.com/Evgeniy347/gircore-x86-cdecl-repro

For transparency: the investigation and the patch were produced by an AI coding agent working under my supervision; I reviewed and ran the results myself.

badcel:
Thanks for pointing out the usage of AI.

There are no AI rules for GirCore yet but it will be added shortly and not allow autonomous Code generation.

So at least the unit test won't be acceptable.

I will do a code review and think the unit test would need to change anyway.

I'm happy you got here to improve the project and work with me to get this PR merged.

src/Libs/GObject-2.0/Internal/InstanceCache.cs
```diff
@@ -63,7 +63,7 @@ public static unsafe void Remove(IntPtr handle)
         Debug.WriteLine($"Handle {handle}: Removed object from {nameof(InstanceCache)}.");
     }
 
-    [UnmanagedCallersOnly]
+    [UnmanagedCallersOnly(CallConvs = new[] { typeof(System.Runtime.CompilerServices.CallConvCdecl) })]
```
badcel, line 66: Can you please use the collection expression.
```suggestion
    [UnmanagedCallersOnly(CallConvs = [typeof(System.Runtime.CompilerServices.CallConvCdecl)])]
```

src/Integrations/GObject-2.0.Integration/SourceGenerator/Subclass/SubclassCode.cs
```diff
@@ -120,7 +120,7 @@ private static string RenderSubclassContent(SubclassData subclassData)
                        }
                        
                        {{GeneratedCodeAttribute.Render()}}
-                       [System.Runtime.InteropServices.UnmanagedCallersOnly]
+                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = new[] { typeof(System.Runtime.CompilerServices.CallConvCdecl) })]
```
badcel, line 123:
```suggestion
                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = [typeof(System.Runtime.CompilerServices.CallConvCdecl)])]
```

src/Integrations/GObject-2.0.Integration/SourceGenerator/Subclass/SubclassCode.cs
```diff
@@ -131,22 +131,22 @@ private static void ClassInit(System.IntPtr cls, System.IntPtr clsData)
                        }
                       
                        {{GeneratedCodeAttribute.Render()}}
-                       [System.Runtime.InteropServices.UnmanagedCallersOnly]
+                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = new[] { typeof(System.Runtime.CompilerServices.CallConvCdecl) })]
```
badcel, line 134:
```suggestion
                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = [typeof(System.Runtime.CompilerServices.CallConvCdecl)])]
```

src/Integrations/GObject-2.0.Integration/SourceGenerator/Subclass/SubclassCode.cs
```diff
@@ -131,22 +131,22 @@ private static void ClassInit(System.IntPtr cls, System.IntPtr clsData)
...
                        private static void InstanceInit(System.IntPtr instance, System.IntPtr cls)
                        {
                            CompositeTemplateInstanceInit(instance, cls);
                        }
                       
                        {{GeneratedCodeAttribute.Render()}}
-                       [System.Runtime.InteropServices.UnmanagedCallersOnly]
+                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = new[] { typeof(System.Runtime.CompilerServices.CallConvCdecl) })]
```
badcel, line 141:
```suggestion
                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = [typeof(System.Runtime.CompilerServices.CallConvCdecl)])]
```

src/Integrations/GObject-2.0.Integration/SourceGenerator/Subclass/SubclassCode.cs
```diff
@@ -131,22 +131,22 @@ private static void ClassInit(System.IntPtr cls, System.IntPtr clsData)
...
                        private static void Dispose(System.IntPtr instance)
                        {
                            CompositeTemplateDispose(instance);
                            GetParentTypeClass()->Dispose(instance);
                        }
                        
                        {{GeneratedCodeAttribute.Render()}}
-                       [System.Runtime.InteropServices.UnmanagedCallersOnly]
+                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = new[] { typeof(System.Runtime.CompilerServices.CallConvCdecl) })]
```
badcel, line 149:
```suggestion
                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = [typeof(System.Runtime.CompilerServices.CallConvCdecl)])]
```

src/Integrations/GObject-2.0.Integration/SourceGenerator/Subclass/SubclassCode.cs
```diff
@@ -251,7 +251,7 @@ private static string RenderAbstractSubclassContent(SubclassData subclassData)
...
```
badcel, line 254:
```suggestion
                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = [typeof(System.Runtime.CompilerServices.CallConvCdecl)])]
```

src/Integrations/GObject-2.0.Integration/SourceGenerator/Subclass/SubclassCode.cs
```diff
@@ -261,14 +261,14 @@ private static void ClassInit(System.IntPtr cls, System.IntPtr clsData)
                        }
                       
                        {{GeneratedCodeAttribute.Render()}}
-                       [System.Runtime.InteropServices.UnmanagedCallersOnly]
+                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = new[] { typeof(System.Runtime.CompilerServices.CallConvCdecl) })]
```
badcel, line 264:
```suggestion
                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = [typeof(System.Runtime.CompilerServices.CallConvCdecl)])]
```

src/Integrations/GObject-2.0.Integration/SourceGenerator/Subclass/SubclassCode.cs
```diff
@@ -261,14 +261,14 @@ private static void ClassInit(System.IntPtr cls, System.IntPtr clsData)
...
                        private static void InstanceInit(System.IntPtr instance, System.IntPtr cls)
                        {
                            CompositeTemplateInstanceInit(instance, cls);
                        }
                       
                        {{GeneratedCodeAttribute.Render()}}
-                       [System.Runtime.InteropServices.UnmanagedCallersOnly]
+                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = new[] { typeof(System.Runtime.CompilerServices.CallConvCdecl) })]
```
badcel, line 271:
```suggestion
                       [System.Runtime.InteropServices.UnmanagedCallersOnly(CallConvs = [typeof(System.Runtime.CompilerServices.CallConvCdecl)])]
```

src/Tests/Libs/GLib-2.0.Tests/CallbackCallingConventionTest.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using AwesomeAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace GLib.Tests;
+
+[TestClass, TestCategory("UnitTest")]
+public class CallbackCallingConventionTest
```
badcel, line 10: I think all tests can be removed. This kind of tests does not align with the current style of tests I want to have. I want to be more on the side of integration tests.

An idea would be to compile the tests  (only the tests) not only for windows on "AnyCpu", but additionally explicitly for "x86" and run both suites.

This should ensure that all tested code paths works on "x86", without the need to pin down the current implementation via reflection to check if a certain attribute is used.

So if the code continues to evolve (e.g. if microsoft creates new APIs) the generator can be updated and the unit test must not be updated, they continue to verify that the "x86" code works as expected.

Please keep in mind that if you want to contribute this change that no significant code blocks may be autonomously be generated by an AI.

Adding some attributes to the generator is not problematic from my point of view, because this is the proper fix to get things working for x86 compiled binaries and is in not some creativ process at all.

Please keep in mind that it might be worth contributing this change as it would ensure that GirCore works for your usecase. I myself do not plan to add those tests immediately. Adding them myself would at least mean a significant delay from month to years until it gets added as I would add them to the already available backlog (see [milestones](https://github.com/gircore/gir.core/milestones)).

badcel (changes requested):
Thanks for the PR.

Please see my comments.

Please squash your commits into one to keep the commit history clean. I would prefer to see your name in the commit which links to your github account, instead of the current one.

A general question: Do you know if GTK4 officially supports x86? There probably is some reason they removed the i686 binaries from MSYS2?

